PR #3, opened 2017-04-04
Bug #2287 If you enter a query with EXPLAIN statement <f7> will lead to an error

rjskr01:


SrMouraSilva:
Mirror of the pgAdmin 4 GIT repo. DO NOT submit pull requests here! Use the pgadmin-hackers@postgresql.org mailing list.

PR #5, opened 2017-05-09
security bug -- need more HTML-escaping for working with tree-nodes

antonov-impulsm:
i mailed this patch to  pgadmin-hackers@postgresql.org 

--------------------------------------------------

good day!

i fixed tiny errors (html-escaping) , but it has security effects.


antonov-impulsm:
thank you vary match! pull request "merged" as cbf2de6c27760669a7c7e48f6bfc47c6053309f8 ..

I closing it

PR #6, opened 2018-01-02
remove httpd.pid before starting httpd

mdeneen:
httpd will not start properly if this file exists.  The docker container will exit if httpd exits, making this file useless.

Without this change it is impossible to start the docker container after stopping it.


mdeneen:
This was merged via the mailing list.

PR #20, opened 2019-01-28
Added the possiblity to bind to ipv4-address on startup

Gontrum:
hello everyone,

unfortunately we had difficulties to bind the container to a host for which no IPV6 is configured. 
For this reason we have made the changes so that gunicorn can also bind to an IPV4 address. 

Many thanks for your work.
Best regards

David

### DIFF
--- a/pkg/docker/README
+++ b/pkg/docker/README
@@ -47,6 +47,14 @@ The expected paths are `/certs/server.crt` and `/certs/server.key`.
 
 You need to explicitly map these ports with `-p` option to some port at your machine.
 
+PGADMIN_USE_IPV4
+----------------
+
+Default: unset
+
+If not set, the container will try to bind the daemon to a IPV6 address,
+which would result in an error if no IPV6 network is configured on the host system.
+
 Examples
 ========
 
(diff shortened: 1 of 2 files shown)


deric:
Any progress with this issue? I'm having the same problem:
```
[2019-02-15 12:10:24 +0000] [1] [INFO] Starting gunicorn 19.8.1
[2019-02-15 12:10:24 +0000] [1] [ERROR] Retrying in 1 second.
[2019-02-15 12:10:25 +0000] [1] [ERROR] Retrying in 1 second.
[2019-02-15 12:10:26 +0000] [1] [ERROR] Retrying in 1 second.
[2019-02-15 12:10:27 +0000] [1] [ERROR] Retrying in 1 second.
[2019-02-15 12:10:28 +0000] [1] [ERROR] Retrying in 1 second.
[2019-02-15 12:10:29 +0000] [1] [ERROR] Can't connect to ('::', 80)
```

deric:
Ahh, it looks like it has been fixed in 36ad0632c69eafa8fa674e4e8b14013a2940eeae. It Might be better idea to detect IPv6 automatically using e.g.:
```
if [ -f /proc/net/if_inet6 ]; then
    BIND_ADDRESS=0.0.0.0
else
    BIND_ADDRESS=[::]
fi
```

xbb:
You can use the env var PGADMIN_LISTEN_ADDRESS, set it to 0.0.0.0 or some other ipv4 address

overlinden:
I had the same issue during the deployment on cloud foundry in the SAP cloud platform. There seems to be no IPv6 support, so gunicorn startup failed with the above mentiones error message.
The workaround to set PGADMIN_LISTEN_ADDRESS to 0.0.0.0 solved the issue. An environment variable to switch between IPv4 and IPv6 would be great.

Issue #5402, opened 2022-10-05
Tree view stuck on fixed height


rayodonnell:
Since the most recent update to pgAdmin4 6.14, the tree view on the left seems to be stuck at a specific height.

When the pgAdmin window is taller, the tree view doesn't expand vertically to fill the available space; and when the window is smaller, the tree view extends out of sight below the bottom of the window.

I'm on Kubuntu 22.04, and here are some details from the "About" dialog:

- Version: 6.14
- Application Mode: Desktop
- Current User: [pgadmin4@pgadmin.org](mailto:pgadmin4@pgadmin.org)
- NW.js Version: 0.62.2
- Browser: Chromium 99.0.4844.84
- Operating System: Linux-5.15.0-48-generic-x86_64-with-glibc2.35
- pgAdmin Database File: /home/rod/.pgadmin/pgadmin4.db
- Log File: /home/rod/.pgadmin/pgadmin4.log

pravesh-sharma:
Fixed in latest snapshot build

PR #5404, opened 2022-10-06
Ensure that scroll bar on browser tree should be visible on windows resize #5402

nikhil-mohite:


web/pgadmin/browser/static/js/browser.js
```diff
@@ -196,7 +196,7 @@ define('pgadmin.browser', [
         isPrivate: true,
         icon: '',
         limit: 1,
-        content: '<div id="tree" class="aciTree"></div>',
+        content: '<div id="tree" class="aciTree browser-tree"></div>',
```
adityatoshniwal, line 199: Can't we remove `aciTree` ?

adityatoshniwal (changes requested):

web/pgadmin/browser/static/js/browser.js
nikhil-mohite: Removed aciTree class with updated PR.

Issue #5058, opened 2022-05-12
Save Button should be disabled by default on the Sort/Filter options (RM #7371)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/7371***
*Originally created by **Fahar Abbas** at 2022-05-12 07:08:59 UTC.*

Save Button should be disabled by default on the Sort/Filter options

Steps
1. Launch pgadmin4
2. open Sort/Filter options
3. Save button is enabled

!Screen%20Shot%202022-05-12%20at%2012.05.56%20PM.png!

FaharAbbasRizvi:
This is tested in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2022-10-24/

PR #5444, opened 2022-10-18
Ensure that the save button should be disabled by default on the Sort/Filter dialog in the query tool #5058

nikhil-mohite:


web/pgadmin/static/js/SchemaView/index.jsx
```diff
@@ -773,7 +773,7 @@ function SchemaDialogView({
               <DefaultButton data-test="Reset" onClick={onResetClick} startIcon={<SettingsBackupRestoreIcon />} disabled={!dirty || saving} className={classes.buttonMargin}>
                 {gettext('Reset')}
               </DefaultButton>
-              <PrimaryButton data-test="Save" onClick={onSaveClick} startIcon={ButtonIcon} disabled={ !(viewHelperProps.mode === 'edit' ? dirty : true) || saving || Boolean(formErr.name) || !formReady}>
+              <PrimaryButton data-test="Save" onClick={onSaveClick} startIcon={ButtonIcon} disabled={ !(viewHelperProps.mode === 'edit' || viewHelperProps.mode == 'create' ? dirty : true) || saving || Boolean(formErr.name) || !formReady}>
```
adityatoshniwal, line 776: No this is not the correct fix. If a data is valid then `Save` button should be enabled irrespective of dirty or not.

adityatoshniwal (changes requested):

web/pgadmin/static/js/SchemaView/index.jsx
adityatoshniwal: The correct fix for this would be to use `edit` mode in filter dialog. And handle things in backend accordingly.

Issue #1832, opened 2018-05-07
[WEB] External database user settings storage  (RM #3336)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/3336***
*Originally created by **Peter Palka** at 2018-05-07 02:32:11 UTC.*

h1. Current user settings storage
Right now all user settings are stored in SQLite database file in /var/lib/pgadmin/pgadmin4.db.

h1. Proposal
Ability to store user settings on external database like PostgreSQL

h1. Motivation
During large pgAdmin4 deployment (for example on AWS using Auto Scaling group with load balancer) we can't rely on single SQLite file database as it:
# Is not prepared to be failure-proof (no HA support),
# May be a bottleneck (not designed to handle many connections concurrently reading/writing data to it).
This implies it can be single point of failure.

To prevent it we can separate user data storage from pgAdmin4 web service by using external database for user storage on another machine. To achieve it the external database connector will be needed.

anovis:
any thoughts on this. this would be very helpful, when deploying in a serverless environment (ie gcp cloudrun)

FaharAbbasRizvi:
This feature is thoroughly tested

Here are the permutations:

External database tested on Windows 64:
------------------
1. pgadmin4 6.14>pgadmin4 6.15 with latest snapshot build
2. pgadmin4 6.9>pgadmin4 6.15 with latest snapshot build
3. pgadmin4 4.30 >pgadmin4 5.7 > pgadmin4 6.9> pgadmin4 6.15 with latest snapshot build
4. Fresh installation mode on Windows 64

Linux 64
------------
1. The fresh mode of pgadmin4 6.15(Latest snapshot build)rpm on the Desktop and the server mode
2. Upgrade scenario on the Desktop and the server mode

MAC OS X
---------------
Fresh and Update scenario

pgadmin4 with CONFIG_DATABASE_URI:
-------------------------------------------
Performed testing with CONFIG_DATABASE_URI  on Linux and Windows with fresh and upgrade scenarios.

adityatoshniwal:
@FaharAbbasRizvi Can you please test this scenario as well:
1. Install v5.3 or or below in server mode.
2. Add some users in user management.
3. Upgrade to the latest.

Changes were done to add a new unique column in the users table, need to test if working fine.

FaharAbbasRizvi:
[...]

PR #5445, opened 2022-10-18
Added support for storing configurations of pgAdmin in an external database. #1832

akshay-joshi:


web/migrations/versions/ca00ec32581b_.py
```diff
@@ -25,7 +24,7 @@
 
 
 def upgrade():
-    db.engine.execute(
+    op.execute(
```
adityatoshniwal, line 27: instead of using plain SQL may be you can try similar to `web/migrations/versions/81c7ffeffeee_.py`.

adityatoshniwal (changes requested):
I have done the code review. I will parallely test this meanwhile you make the changes.

adityatoshniwal:
The code looks good to me. Tested on both PostgreSQL and SQLite3

Issue #4841, opened 2022-01-07
Schema diff - Timeout error is not handled when pgAdmin4 is hosted on Apache/Httpd (RM #7104)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/7104***
*Originally created by **Murtuza Zabuawala** at 2022-01-07 06:32:35 UTC.*

Newer version of httpd on CentOS/Redhat has very low TimeOut value by default - 60 sec
So if schema diff takes longer than that it throws blank error popup.

Refer attached screenshot.

Expected - Display proper error instead of blank popup.

adityatoshniwal:
This can be handled with SocketIO lib which pgAdmin is already using in psql and in ERD to fetch database tables data.

FaharAbbasRizvi:
Tested on candidate build for v6.16.

PR #5456, opened 2022-10-20
use SocketIO instead of REST in schema diff tool #4841

pravesh-sharma:


web/pgadmin/tools/schema_diff/static/js/components/SchemaDiffCompare.jsx
```diff
@@ -268,15 +269,18 @@ export function SchemaDiffCompare({ params }) {
     }
   };
 
-  const triggerCompareDiff = ({ sourceData, targetData, compareParams, filterParams }) => {
+  const handleAxiosCatch = (err) => {
```
adityatoshniwal, line 272: No need to add this function if its only used once.

web/pgadmin/tools/schema_diff/__init__.py
```diff
@@ -436,39 +436,35 @@ def schemas(sid, did):
...
+        compare_pre_validation(params['trans_id'], params['source_sid'],
+                               params['target_sid'])
     if not status:
+        socketio.emit('compare_database_failed', error_msg,
+                      namespace=SOCKETIO_NAMESPACE, to=request.sid)
         return error_msg
 
     comparison_result = []
 
-    diff_model_obj.set_comparison_info(COMPARE_MSG, 0)
```
adityatoshniwal, line 460: `socketio.emit('compare_status'` is missing here.

adityatoshniwal (changes requested):

Issue #5455, opened 2022-10-20
{EPAS-15/PG-15] Error Message displayed if Click on dependencies Tab for user


FaharAbbasRizvi:
Steps

1. Install PG-15 and EPAS 15 
2. Install the latest pgadmin4 v6.15
3. Click on user postgres or enterprisedb
4. Click on the dependencies Tab
5. Error message displayed "Request failed with status code 500"

Expected Result:
No error is displayed if the user clicks on the dependencies tab

![Uploading Screen Shot 2022-10-20 at 12.19.53 PM.png…]()

FaharAbbasRizvi:
<img width="1786" alt="Screen Shot 2022-10-20 at 12 19 53 PM" src="https://user-images.githubusercontent.com/43006122/196884227-0ed79ba7-f415-4628-a255-20a7f3738caf.png">

FaharAbbasRizvi:
This is tested in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2022-11-01

PR #5461, opened 2022-10-21
fixed issue where dependents tab was not working for PG-15 #5455

pravesh-sharma:


web/pgadmin/browser/server_groups/servers/roles/templates/roles/sql/default/dependents.sql
```diff
@@ -10,7 +10,7 @@ ORDER BY 1, 2
 {% endif %}
 
 {% if fetch_dependents %}
-SELECT cl.relkind, COALESCE(cin.nspname, cln.nspname) as nspname,
+SELECT CAST(cl.relkind as text), COALESCE(cin.nspname, cln.nspname) as nspname,
```
adityatoshniwal, line 13: or try `cl.relkind::text` which is used mostly.

adityatoshniwal (changes requested):

web/pgadmin/browser/server_groups/servers/roles/templates/roles/sql/default/dependents.sql
adityatoshniwal: Also test on other PG server verions.

pravesh-sharma: Tested on PG 12 and above.

adityatoshniwal (approved):

Issue #5468, opened 2022-10-26
Oauth2: Allow to use other claims than mail/email as username


lmaraite:
It would be really nice to customize the Oauth2 claim, which is used for the username. Until now the email or mail claim are used by default and are not configurable. There are cases, e.g. if you use AzureAD where you want to use another id like the oid as unique username.

yogeshmahajan-1903:
This is working fine. User can configure claim that should be used as user in Oauth2 configuration.

PR #5469, opened 2022-10-26
feat(oauth2): Add possibility to configure a custom Oauth2 username claim

lmaraite:
There are cases where you don't want to use the email as username when using Oauth2 authentication.
This feature provides the possibility to configure the Oauth2 claim which should be used as username. The key in the config.py is called 'OAUTH2_USERNAME_CLAIM'. If you don't provide a custom key, the email is used as username, like before. So it is completely backwards compatible.

#5468

### DIFF
--- a/web/config.py
+++ b/web/config.py
@@ -11,6 +11,7 @@
 #
 ##########################################################################
 
+from pgadmin.utils import env, IS_WIN, fs_short_path
 import builtins
 import logging
 import os
@@ -22,7 +23,6 @@ root = os.path.dirname(os.path.realpath(__file__))
 if sys.path[0] != root:
     sys.path.insert(0, root)
 
-from pgadmin.utils import env, IS_WIN, fs_short_path
 
 ##########################################################################
 # Application settings
@@ -754,6 +754,10 @@ OAUTH2_CONFIG = [
         # Oauth scope, ex: 'openid email profile'
         # Note that an 'email' claim is required in the resulting profile
         'OAUTH2_SCOPE': None,
+        # The claim which is used for the username. If the value is empty the
+        # email is used as username, but if a value is provided,
+        # the claim has to exist.
+        'OAUTH2_USERNAME_CLAIM': None,
         # Font-awesome icon, ex: fa-github
         'OAUTH2_ICON': None,
         # UI button colour, ex: #0000ff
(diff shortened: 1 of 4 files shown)


web/config.py
```diff
@@ -761,6 +761,10 @@
     }
 ]
 
+# The claim which is used for the username. If the value is empty the
+# email is used as username, but if a value is provided, the claim has to exist.
+OAUTH2_USERNAME_CLAIM: None
```
adityatoshniwal, line 766: This should be `OAUTH2_USERNAME_CLAIM = None`

adityatoshniwal (changes requested):

web/config.py
lmaraite: Yes, you're right. I've changed it. Thank you!

web/config.py
```diff
@@ -761,6 +761,10 @@
     }
 ]
 
+# The claim which is used for the username. If the value is empty the
```
yogeshmahajan-1903, line 764: This setting should be a part of individual oauth2 config dictionary, which will allow segregation of oauth2 claim setting in case of multiple oAuth2 provider.
Also please fix pep8 issue.

yogeshmahajan-1903 (changes requested):

web/config.py
lmaraite: Yes, I didn't thought about that. I've changed it and its now working for multiple OAuth2 provider.
Also the workflows should be green now.

web/pgadmin/authenticate/oauth2.py
```diff
@@ -123,6 +123,22 @@ def login(self, form):
             [value for value in self.email_keys if value in profile.keys()]
         email = profile[email_key[0]] if (len(email_key) > 0) else None
 
+        username = email
+        username_claim = None
+        if 'OAUTH2_USERNAME_CLAIM' in self.oauth2_config:
```
yogeshmahajan-1903, line 128: Condition should be 'OAUTH2_USERNAME_CLAIM' in self.oauth2_config[self.oauth2_current_client]. 
self.oauth2_config is value for key which maps to OAUTH2_NAME.

yogeshmahajan-1903 (changes requested):

akshay-joshi:
@lmaraite Can you please fix the review comment and send the PR again? We will have a release next week.

web/pgadmin/authenticate/oauth2.py
lmaraite: Resolved this. Sorry for the late response.

yogeshmahajan-1903 (approved):

PR #5476, opened 2022-10-28
png optimization(loss less)

pea-sys:
Lighter png files. Image quality is not degraded.
I used WinMerge when checking the image loss less


adityatoshniwal:
@pea-sys Great work. Can you also let us know the tool which you used to achieve this. pgAdmin UI keeps changing and we always keep the docs updated with latest screenshots.

pea-sys:
@adityatoshniwal 
I used zopflipng
https://github.com/google/zopfli/blob/master/README.zopflipng

Issue #5475, opened 2022-10-28
Switch "Save unsaved query changes" in "Preferences" does not affect when the query tool is opened in a separate window


IIIIIME2IIIII:
When the option Preferences->Browser->Tab settings->Open in new browser tab 
containig "Query Tool" (e.g. when the "Query Tool" opens in separate window),
value of option Preferences->Query Tool->Options->Prompt to save unsaved query changes?
is always ignored, and prompt is raised every time on closing window.

IIIIIME2IIIII:
Thanks a lot!

nikhil-mohite:
@IIIIIME2IIIII ,

You need to disable the setting 'Confirm on close or refresh?' from preferences -> Browser -> Display, for this as the Query tool will get open in a new browser tab. (Fix will be available in the next release or you can try the daily snapshot build from   https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/, snapshot build are not tested you may face some issues)

akshay-joshi:
Reopen it for Testing.

FaharAbbasRizvi:
This is tested in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/

PR #5481, opened 2022-11-01
Fixed an issue where the 'Confirm on close or refresh?' setting is ignored while closing the query tool & ERD tool opened in the new tab #5475

nikhil-mohite:


web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -487,6 +486,17 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
     });
   };
 
+  useEffect(()=> {
+    // Add beforeunload event if "Confirm on close or refresh" option is enabled in the preferences.
+    if(qtState.preferences.browser.confirm_on_refresh_close){
```
adityatoshniwal, line 491: I think we'll need the else case as well. A user can change preference after query tool opens. So the already bound event will not be removed in that case.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
nikhil-mohite: Done, also added the check-in ERD tool.

adityatoshniwal (approved):

PR #5482, opened 2022-11-02
Update setup-web.sh

sabatex:
WSL 2 (Ubuntu 20.04) fixed (line 131 change unsupported in wsl command systemctl)


dpage (changes requested):
The `service` command isn't going to be present on all Ubuntu/Debian installations. The code should figure out whether to use `systemctl` or `service` automatically based on what is present on the system at runtime.

Thanks.

akshay-joshi:
Handled Dave's review comment in separate PR #5582, hence closing it.

Issue #4756, opened 2021-11-11
Generate ERD from a single Table (RM #6994)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/6994***
*Originally created by **Yu Jin** at 2021-11-11 12:19:39 UTC.*

pgadmin can create an ERD from a database, which is awesome but when working with large schemas especially the ones you didn't develop yourself, it can get a bit overwhelming. Sometimes I just want to know relations of a specific table to keep it simple and understand just that specific part of the schema. In this context it would be useful if the user could right click on a table and choose "create ERD" just as with the database, but then the ERD tool would create the diagram containing just the table and all the tables which have direct relations to the chosen table. For example, the oracle sqldeveloper can do it (clicking on a table and then choosing "model" tab) and I use it a lot, equivalent to that in pgadmin would make it even easier to prefer postgreSQL over oracle.

PR #5485, opened 2022-11-02
Allow to generate ERD for a table. #4756

adityatoshniwal:


web/pgadmin/browser/server_groups/servers/databases/schemas/tables/static/js/table.js
```diff
@@ -121,7 +121,11 @@ define('pgadmin.node.table', [
           applies: ['object', 'context'], callback: 'count_table_rows',
           category: 'Count', priority: 2, label: gettext('Count Rows'),
           enable: true,
-        },
+        },{
```
akshay-joshi, line 124: I don't think so we need a separate menu "Generate Table ERD" it should be "Generate ERD" only. It will generate the diagram based on the object.

web/pgadmin/tools/erd/__init__.py
```diff
@@ -386,6 +386,20 @@ def register_preferences(self):
             )
         )
 
+        self.preference.register(
+            'options',
+            'table_relations_depth',
+            gettext('Table Relations Depth'),
```
akshay-joshi, line 392: Depth of table relations or Depth of related tables?

web/pgadmin/tools/erd/__init__.py
```diff
@@ -386,6 +386,20 @@ def register_preferences(self):
...
+            'integer',
+            -1,
+            category_label=PREF_LABEL_OPTIONS,
+            help_str=gettext(
+                'The maximum depth pgAdmin should traverse to find '
```
akshay-joshi, line 397: The message should be "A maximum depth at which pgAdmin should search for related tables when generating an ERD. A value of -1 indicates that there is no limit."

akshay-joshi (changes requested):
-- This script was generated by a beta version of the ERD tool in pgAdmin 4. -- Please log an issue at https://redmine.postgresql.org/projects/pgadmin4/issues/new if you find any bugs, including reproduction steps.
1) Correct the above comment in SQL scripts. Remove "beta" and point to the correct GitHub link.

2) The title of the panel should be the name of the Table for which ERD is generated. 

3) As per the requirement, the "ERD tool would create the diagram containing just the table and all the tables which have direct relations to the chosen table" if we are going to implement the above then I think we should not need the preferences setting "Depth".

For example, I am having a table with only primary constraints and some other tables have a foreign key constraint on this table. If I select and click "Generate Table ERD" it will consider all the tables doesn't matter it is having direct relation or not.

4) No relationship is visible for one specific scenario (Macro and User Macro table with Depth) We can connect for this issue.

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/static/js/table.js
adityatoshniwal: OK

web/pgadmin/tools/erd/__init__.py
adityatoshniwal: It is depth of `relations`. Depth of table has no meaning.

web/pgadmin/tools/erd/__init__.py
adityatoshniwal: Using `A` is incorrect. There is only one input for depth, so it should be `The`

adityatoshniwal:
> -- This script was generated by a beta version of the ERD tool in pgAdmin 4. -- Please log an issue at https://redmine.postgresql.org/projects/pgadmin4/issues/new if you find any bugs, including reproduction steps.
> 
> 1. Correct the above comment in SQL scripts. Remove "beta" and point to the correct GitHub link.
OK
> 2. The title of the panel should be the name of the Table for which ERD is generated.
It is not like query tool. In query tool, the table is freezed and query cannot be changed. In case of ERD, table ERD is just a starting point. User can edit it completely. So adding table will only confuse.
> 3. As per the requirement, the "ERD tool would create the diagram containing just the table and all the tables which have direct relations to the chosen table" if we are going to implement the above then I think we should not need the preferences setting "Depth".
> 
> For example, I am having a table with only primary constraints and some other tables have a foreign key constraint on this table. If I select and click "Generate Table ERD" it will consider all the tables doesn't matter it is having direct relation or not.
And so the default value is `-1 no limit`. Depth is imp because in a hyper connected database the table can have indirectly connect to each and every table.
Probably this will give more context - https://www.postgresql.org/message-id/CAF5nyP8nUp8wAuB1RUG_S74HPeCGxhzxWJe+w0Q=M3A2TXXrOQ@mail.gmail.com
> 
> 4. No relationship is visible for one specific scenario (Macro and User Macro table with Depth) We can connect for this issue.
We can discuss.

adityatoshniwal:
> -- This script was generated by a beta version of the ERD tool in pgAdmin 4. -- Please log an issue at https://redmine.postgresql.org/projects/pgadmin4/issues/new if you find any bugs, including reproduction steps.
> 
> 1. Correct the above comment in SQL scripts. Remove "beta" and point to the correct GitHub link.

OK
> 2. The title of the panel should be the name of the Table for which ERD is generated.

It is not like query tool. In query tool, the table is freezed and query cannot be changed. In case of ERD, table ERD is just a starting point. User can edit it completely. So adding table will only confuse.
> 3. As per the requirement, the "ERD tool would create the diagram containing just the table and all the tables which have direct relations to the chosen table" if we are going to implement the above then I think we should not need the preferences setting "Depth".
> 
> For example, I am having a table with only primary constraints and some other tables have a foreign key constraint on this table. If I select and click "Generate Table ERD" it will consider all the tables doesn't matter it is having direct relation or not.
> 
> 4. No relationship is visible for one specific scenario (Macro and User Macro table with Depth) We can connect for this issue.

web/pgadmin/tools/erd/__init__.py
adityatoshniwal: @dpage Can you please suggest the strings here ?
Label: Table Relations Depth
Help: The maximum depth pgAdmin should traverse to find related tables when generating ERD for a table. Use -1 for no limit.

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/static/js/table.js
adityatoshniwal: Changed to `ERD For Database` and `ERD For Table`

akshay-joshi (approved):

Issue #5513, opened 2022-11-10
pgAdmin update required AZURE_CREDENTIAL_CACHE_DIR in config_local.py


dgilman:
Hello, just did an update from pgadmin4 6.8 to 6.15. I had to add AZURE_CREDENTIAL_CACHE_DIR to my config_local.py. Up until now, I was OK with following the default recommendations from [the docs](https://www.pgadmin.org/docs/pgadmin4/latest/server_deployment.html) which just had you define LOG_FILE, SQLITE_PATH, SESSION_DB_PATH and STORAGE_DIR.

I think the bug here is that the docs should be updated to mention that AZURE_CREDENTIAL_CACHE_DIR should be defined as well.

pravesh-sharma:
We should explore how we can set STORAGE_DIR dependant config variables automatically without user needing to set it manually.

khushboovashi:
Working fine. Tested with the wheel package. Ref:  https://developer.pgadmin.org/builds/2022-11-30-1/

PR #5574, opened 2022-11-25
#5513 - Create DATA_DIR dependent folder/files are automatically inside specified DATA_DIR in configuration file.

yogeshmahajan-1903:


web/config.py
```diff
@@ -860,10 +878,29 @@
...
+# of DATA_DIR is user defined
+data_dir_dependent_settings = ['LOG_FILE', 'SQLITE_PATH', 'SESSION_DB_PATH',
+                               'AZURE_CREDENTIAL_CACHE_DIR',
+                               'KERBEROS_CCACHE_DIR', 'STORAGE_DIR']
+
+if 'DATA_DIR' in user_config_settings:
+    for setting in data_dir_dependent_settings:
+        if setting not in user_config_settings:
+            data_dir = user_config_settings['DATA_DIR']
+            file_dir_name = config.__dict__[setting].split('/')[-1]
```
adityatoshniwal, line 898: Can you `locals()` to access local variables? Refer - https://docs.python.org/3/library/functions.html#locals.
I am worried `import config` inside config can create cyclic dependency.

adityatoshniwal (changes requested):
Otherwise, the code looks good to me.

web/config.py
adityatoshniwal: @yogeshmahajan-1903 I also noticed you've used `split('/')`. This will not work in case of Windows.

adityatoshniwal: Its better to use https://docs.python.org/3/library/os.path.html python OS lib for path manipulation. In your case - `os.path.basename` instead of split.

web/config.py
```diff
@@ -860,10 +877,30 @@
...
+data_dir_dependent_settings = ['LOG_FILE', 'SQLITE_PATH', 'SESSION_DB_PATH',
+                               'AZURE_CREDENTIAL_CACHE_DIR',
+                               'KERBEROS_CCACHE_DIR', 'STORAGE_DIR']
+
+if 'DATA_DIR' in user_config_settings:
+    for setting in data_dir_dependent_settings:
+        if setting not in user_config_settings:
+            data_dir = user_config_settings['DATA_DIR']
+            file_dir_name = \
+                locals().get(setting).os.path.basename(locals().get(setting))
```
adityatoshniwal, line 898: Just one minor thing, simply `os.path.basename(locals().get(setting))` instead of `locals().get(setting).os.path.basename(locals().get(setting))` won't work?

adityatoshniwal (changes requested):

web/config.py
yogeshmahajan-1903: Accommodated review comments

web/config.py
yogeshmahajan-1903: Accommodated review comments

adityatoshniwal (approved):
Cool! Looks good to me.

Issue #5569, opened 2022-11-24
Support AWS for BigAnimal cloud deployment


khushboovashi:

yogeshmahajan-1903:
This is working fine. Verified on candidate build

PR #5644, opened 2022-12-14
Support AWS provider for BigAnimal cloud deployment. #5569

khushboovashi:


web/pgadmin/misc/cloud/biganimal/__init__.py
```diff
@@ -261,11 +263,11 @@ def _check_admin_permission(self):
                 return True
         return False
 
-    def get_regions(self):
+    def get_regions(self, provider_id):
         """Get regions"""
-        _url = "{0}/{1}".format(
+        _url = '{0}/cloud-providers/{1}/regions'.format(
```
yogeshmahajan-1903, line 268: Fix Sonarlint issue - Change this code to not construct the URL from user-controlled data.

yogeshmahajan-1903 (commented):
1.Getting error - {"message":"Please provide connected cloud provider"} for AWS provider, as  user does not have AWS provider connected in Biganimal account. Enable/Disabled providers based on user scope to avoid this error.
2.Instance Sizes should be sorted.
3.Cluster Type step is visible and blank in case of Azure.
4.In Case of AWS,  Cluster Type tab is skipped and marked checked. User does not have any idea when did he navigate though that tab.
5.If user clicks back button, he gets blank screen for Cluster type.
6.Data is not persisted when navigated to previous tab again.

Issue #5613, opened 2022-12-06
pgAdmin installed by one Mac user can't open due to wrong permissions for other users


DrewMerkle:
Whether installed from direct download or using brew, pgAdmin can be used by the Mac user who installs it, but the permissions within the application bundle seem to be incorrect because other users on the same Mac cannot open or use the application. 

This can be corrected by adding the correct permissions in the right places within the app bundle to allow all users on the same Mac to open and run pgAdmin. 

Hopefully the Mac installer and the brew cask can be updated to set permissions to enable use by all users. 

For now, this command seems to work for me when run from the admin account I use for installing and updating software:
`cd /Applications/pgAdmin\ 4.app/; sudo chmod -R g+rx .`

I'm using macOS 12.6.1 on an M1 Mac. I've experienced this issue with pgAdmin 6.17 and several previous 6.x versions. 

Version 6.17
Application Mode Desktop
Current User pgadmin4@pgadmin.org
NW.js Version 0.70.1
Browser Chromium 107.0.5304.88
Operating System macOS-12.6.1-x86_64-i386-64bit

kelbstf:
Hello @DrewMerkle,

i just stumbled upon your post, which seems to be related to my experiences (https://github.com/pgadmin-org/pgadmin4/issues/5579). 

Or maybe not: 
a) i also had to change the permissions (i used bruteforce 777, just to make sure), and that allowed me to launch the application.
b) but: i still can't connect to any server, unless i launch pgAdmin as an administrative user...

Does this work for you?
If so: are you launching pgAdmin as an administrative user?

Which accounts own your application tree?
I tried both with "root:wheel" and "root:admin", but that did not help...

My guts feeling is that the network connection to the PG server may possibly involve another part of the filesystem, where administrative access is required?
Just guessing...

@akshay-joshi: thank you a ton for having assigned my post!

Best

DrewMerkle:
Hello @kelbstf, thank you for your comment. In answer to your question, I did not personally have difficulty making database connections from the Mac user account used to install pgAdmin (whether using Mac install or using Brew). 

The only issue I had was that I use an account with local-machine admin privileges to do installs and maintenance, but I use a non-admin account for daily use as an extra layer of security. The daily-use account could not open pgAdmin without correcting the file permissions within the application bundle. However, once the file permissions are corrected, everything works well for me. 

Therefore, I cannot determine if your database connection issue is related to the local machine file permissions issue that I experienced.

PR #5649, opened 2022-12-15
#5613 - Ensure appbundle has correct permissions so that pgAdmin can be accessed by users other than owner.

yogeshmahajan-1903:


dpage:
That doesn't seem right.

1) Shouldn't it be done at the end of the build (right before signing) so all the contents are correct?
2) Blindly setting 755 is going to make files executable, whether or not they should be.

Issue #5542, opened 2022-11-16
Query tool is on the wait state if user rename schema and click on generate script with schema diff


FaharAbbasRizvi:
Candidate build v6.16

Steps
1. Create two databases with sample and sample2 
2. execute pg-sample.sql on both databases and rename some objects in sample2 database
3. Open schema diff
4. select source and target databases without selecting any schema
5. Click on compare
6.  schema diff open with objects successfully
7. Now go into sample2 database and rename schema with public2
8. Now Click on compare again
9. schema diff open with objects successfully
10. Select all objects and click on generate script
11. query tool is on the wait state

Expected Result:
Generate script should be working if the user renames the schema and click on compare

<img width="1786" alt="Screen Shot 2022-11-16 at 2 46 22 PM" src="https://user-images.githubusercontent.com/43006122/202147561-b9a40f1a-fda6-40ee-98f6-eb026c231ec2.png">
<img width="1776" alt="Screen Shot 2022-11-16 at 2 47 24 PM" src="https://user-images.githubusercontent.com/43006122/202147613-821e1984-2346-4a7c-8409-d06b108dfe88.png">

adityatoshniwal:
Simple steps to reproduce:
1. Rename a schema.
2. Click on query tool button.

adityatoshniwal:
Related to #5559.

FaharAbbasRizvi:
This is tested in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-01-06/

PR #5651, opened 2022-12-15
Fixed an issue update schema node de-select the node in tree if only one schema present in the collection node. #5542

nikhil-mohite:


web/pgadmin/browser/static/js/browser.js
```diff
@@ -1699,6 +1699,8 @@ define('pgadmin.browser', [
                 // If the current node's inode is false
                 if (__ctx.i && !__ctx.t.isInode(__ctx.i)) {
                   __ctx.t.setInode(__ctx.i, {success: _appendNode});
+                  // Open the collection node.
```
adityatoshniwal, line 1702: Shouldn't the node be re-selected if it is renamed? If anything is removed then collection should be selected.

nikhil-mohite: If the collection node has a single child node that child node will get auto-selected on update/rename and if the user removes the child node it will select the collection node if no other child node is present.

adityatoshniwal: OK.

Issue #5487, opened 2022-11-02
Failed to decrypt the saved password


dev3fyfy:
i face this error when saved password to server db and  relogin agin to account , it don't save password and display this message 
Failed to decrypt the saved password. Error: 'utf-8' codec can't decode byte 0xcd in position 2: invalid continuation byte.
![iScreen Shoter - 20221101183154207](https://user-images.githubusercontent.com/92525726/199531533-64791ca2-fe3e-4400-bee0-9b147cfe7a87.jpg)

yogeshmahajan-1903:
@dev3fyfy Can you please share special characters used as part of password?

dev3fyfy:
> @dev3fyfy Can you please share special characters used as part of password?

I just use normal letters and numbers in this password

yogeshmahajan-1903:
@dev3fyfy Have you switched from Desktop to server mode or vice-versa?

dev3fyfy:
> @dev3fyfy Have you switched from Desktop to server mode or vice-versa?

no, this happen with me in server mode

pravesh-sharma:
Hi @dev3fyfy 
Are you using shared server? if yes are you getting this error for non privileged user or admin user.
Since when this error started occurring?
Can you please share the pgAdmin logs?

ramank775:
I am facing same issue with 6.16, I am using docker image to host the pgadmin server behind nginx proxy

> Hi @dev3fyfy Are you using shared server? if yes are you getting this error for non privileged user or admin user. Since when this error started occurring? Can you please share the pgAdmin logs?

I am using shared server, and facing problem with both admin as well as non privileged users.
Only the user who created the server is able to use the saved password.

FYI: All the users are using the same database user and password to connect. is that's  issue?

ramank775:
Is there any update on this issue?

If no one is working on this, i am willing to work on this but might need some guidance to start.

adityatoshniwal:
@ramank775 Sure. Please checkout the README and https://www.pgadmin.org/development/ on how to setup pgAdmin dev environment and contribute.

ramank775:
Thanks @adityatoshniwal.
I have setup the code base and started debugging the issue.

My initial finding
- If both the users shared the same password, this issues doesn't occur. (The user login password).
  My initial guess over here is the password is encrypted with password of the admin who has configured the shared server.

ramank775:
Hi @adityatoshniwal I have raised the PR with the fix. can you please review

[...]

PR #5665, opened 2022-12-21
fix: :bug:  incorrect password used with shared server

ramank775:
ISSUE: When connecting with shared server using saved password fails, due trying to decrypt saved password of User who created the server with the logged In user keys. 

FIX: for the shared server use the logged In user password as db password to connect.

Closes Failed to decrypt the saved password #5487


adityatoshniwal:
@ramank775 The changes looks good to me.

adityatoshniwal (approved):

Issue #5626, opened 2022-12-09
Actions performed on tree node does not update context/ menu bar options


yogeshmahajan-1903:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Actions performed on tree node does not update context/ menu bar options

**To Reproduce**

1. Create a table with trigger.
2. Consider all triggers are disabled.
3. Enable all triggers from context menu.
4. Check context menu updated or not.
5. Enable All is still enabled.

**Expected behavior**

Expected Enable all should be disabled and Disable all should be enabled.

**Error message**

If possible, include the full, unedited text of any error messages displayed.

**Screenshots**

<img width="746" alt="Screenshot 2022-12-09 at 2 08 08 PM" src="https://user-images.githubusercontent.com/16016019/206660250-f3bcd1e4-234c-4ea9-8444-548ec37758a5.png">
<img width="744" alt="Screenshot 2022-12-09 at 2 07 23 PM" src="https://user-images.githubusercontent.com/16016019/206660272-7eb652bd-60c1-4619-804a-8a0bedf0bea1.png">

FaharAbbasRizvi:
This is fixed in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-01-05/

Issue #5627, opened 2022-12-09
Trigger submenu 'Enable All' is enabled though all triggers are enabled.


yogeshmahajan-1903:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

**To Reproduce**

1. Create a table with trigger.
2. Enable the all trigger.
3. Refresh the page.
4.Trigger submenu 'Enable All' is enabled though all triggers are enabled after refresh also.
**Expected behavior**

Trigger submenu 'Enable All' should be disabled as all triggers are enabled.

**Error message**

If possible, include the full, unedited text of any error messages displayed.

**Screenshots**
<img width="740" alt="Screenshot 2022-12-09 at 2 02 25 PM" src="https://user-images.githubusercontent.com/16016019/206659267-eb49332c-c05f-478b-853a-53810f7a5b0d.png">

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10, OSX monterey] - Does not matter.
 - Version: [e.g. 6.15] - 6.16
 - Mode: [e.g Desktop or Server] - Any mode
 - Browser (if running in server mode): [e.g. chrome, safari] - Does not matter
 - Package type: [e.g. RPM, DEB, Python, Container, etc.] - Does not matter

**Additional context**

FaharAbbasRizvi:
This is fixed and tested in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-01-03/

PR #5679, opened 2022-12-27
Ensure submenus for trigger context menu are enabled/disabled correctly #5627

nikhil-mohite:


nikhil-mohite:
This will also fix #5626

Issue #5647, opened 2022-12-14
Row Count notification should not disappear automatically 


akshay-joshi:
1) Row count notification should not be disappeared automatically.
2) Row count should get updated on the Properties panel   

**To Reproduce**

1. Select any table with records of more than 2000.
2. Make sure the Properties tab is selected.
3. Right-click and select the 'Count Row' option from the context menu.

![Countrow](https://user-images.githubusercontent.com/1632882/207604944-e567d8af-0d58-4c80-a182-feca864bdb8e.png)

PR #5693, opened 2022-12-30
Fixed an issue where row count notification was disappearing automatically. #5647

pravesh-sharma:


web/pgadmin/static/js/helpers/Notifier.jsx
```diff
@@ -119,34 +120,35 @@ AlertContent.propTypes = {
 
 
 let Notifier = {
-  success(msg, autoHideDuration = AUTO_HIDE_DURATION) {
-    this._callNotify(msg, MESSAGE_TYPE.SUCCESS, autoHideDuration);
+  success(msg, autoHideDuration = AUTO_HIDE_DURATION, persist = PERSIST_SNACK_BAR) {
```
adityatoshniwal, line 123: We don't need these changes. Just pass the autoHideDuration as null. 
https://github.com/iamhosseindhv/notistack/blob/v1.0.10/src/index.d.ts#L125

web/pgadmin/browser/static/js/node_view.jsx
```diff
@@ -210,6 +210,7 @@ export function getNodeView(nodeType, treeNodeInfo, actionType, itemNodeData, fo
         key={itemNodeData?._id}
         formType={formType}
         getInitData={initData}
+        updatedData={{rows_cnt: itemNodeData?.rows_cnt}}
```
adityatoshniwal, line 213: I think this should be conditional and for now, this data should be passed only for table, view and places where count row is applicable.

adityatoshniwal (changes requested):

PR #5724, opened 2023-01-10
Made min password length same across the application

pravesh-sharma:


adityatoshniwal:
@pravesh-sharma The PR is in draft state.

web/config.py
```diff
@@ -192,6 +192,9 @@
 # Hashing algorithm used for password storage
 SECURITY_PASSWORD_HASH = 'pbkdf2_sha512'
 
+# Minimum password length
+SECURITY_PASSWORD_LENGTH_MIN = 6
```
adityatoshniwal, line 196: Will this apply when adding new user from users dialog?

adityatoshniwal (changes requested):

web/config.py
pravesh-sharma: users dialog is already having validation on 6 characters.

adityatoshniwal: OK.

PR #5753, opened 2023-01-18
Fix orphan database connections resulting in inability to connect to …

khushboovashi:
…databases. #5567


yogeshmahajan-1903 (changes requested):
Server is not reconnected when user close tab and reopens.
Fix sonarlint issue in heartbeat.py.

yogeshmahajan-1903 (approved):
Changes look good to me.

Issue #5746, opened 2023-01-17
pgAdmin 4 Database: field setting.value too short


cklein05:
**The field `value` of table `setting` in pgAdmin's database may be too short**

I'm seeing these errors in PostgreSQL's log: (this pgAdmin uses PostgreSQL for it's database)

```
[...] ERROR:  value too long for type character varying(1024)
[...] STATEMENT:  INSERT INTO setting (user_id, setting, value) VALUES (3, 'Browser/Layout', '{"floating":[], [...]
```
Seems like the length of the JSON value for setting `Browser/Layout` may exceed 1024 characters (in my case it consists of 1077 characters). Actually, this setting never gets saved in that case.

**To Reproduce**

1. Statement fails whenever a property of the `Browser/Layout` setting has changed
2. e. g. Clicking on one of the tabs _Dashboard_, _Properties_, _SQL_, etc.

**Expected behavior**

Column should be wide enough to hold even long JSON values.

**Solutions**

This pgAdmin installation uses PostgreSQL for it's database, so changing the columns width to (e. g.) 2048 resolves the problem (of course). You could consider changing the `setting` table's `value` column width to a much larger value. Alternatively, you could use something like an _unbounded_ column (`text` in PostgreSQL, SQLite should have something similar - aren't SQLite's datatypes no more than syntactic sugar anyway?)

nikhil-mohite:
The browser layout is getting rest to the default value with an external database (PostgreSQL).
 1. Open the older pgAdmin 4 version and change some browser layouts.
 2. Now open the latest pgAdmin 4 version, and check the browser layout changes are now reset to default and that changes done earlier were not preserved. (snapshot build https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-01-25/)

nikhil-mohite:
Resolved, Verified on the latest snapshot build https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-01-31/
OS: macOS Monterey 12.6
Mode: Desktop (Expretal DB Postgres)

Issue #5732, opened 2023-01-12
Info about kerberos authentication to server not exported or imported


alexeyg9:
While exporting or importing servers and server groups definitions flag "Kerberos authentication?" not exported or imported, see functions load_database_servers and dump_database_servers in file web/pgadmin/utils/__init__.py

nikhil-mohite:
Resolved, verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-01-27/
Platform: macOS Monterey 12.6
Mode: Desktop

Issue #5525, opened 2022-11-14
Evaluate Flask-Migrate 4.0.0


musicinmybrain:
[Flask-Migrate](https://pypi.org/project/Flask-Migrate/) has just released version 4.0.0.

Please consider the following:

- Review the [change log](https://github.com/miguelgrinberg/Flask-Migrate/blob/v4.0.0/CHANGES.md#flask-migrate-change-log) for this release.
- Review the use of Flask-Migrate in [`web/pgadmin/setup/db_upgrade.py`](https://github.com/pgadmin-org/pgadmin4/blob/master/web/pgadmin/setup/db_upgrade.py) (the only place it is used in this project).
- Decide whether the dependency specifications can be loosened to allow version 4.x:

https://github.com/pgadmin-org/pgadmin4/blob/c5e9aa6357aee37986ce25b3775f6aa6111e84d7/DEPENDENCIES#L30

https://github.com/pgadmin-org/pgadmin4/blob/c5e9aa6357aee37986ce25b3775f6aa6111e84d7/requirements.txt#L16

musicinmybrain:
Has anyone had a chance to take a look a this? It seems like it should be fine, but it’s not immediately obvious to me how to test it.

dpage:
I don't see any pressing need to upgrade from the release notes - certainly there is no mention of any security fixes that I can spot. Why do you feel this needs to be prioritised?

musicinmybrain:
I’m delaying upgrading the `python-flask-migrate` package in the development version (Rawhide) of Fedora Linux because the `pgadmin4` package pins the older version. I can keep doing that, but preferably for weeks or months rather than indefinitely.

The version pin could be loosened as a downstream patch if necessary, but I’m not sure how to actually test compatibility.

dpage:
Akshay, please ensure this is handled in the next round of dep updates. Thanks.

musicinmybrain:
> Akshay, please ensure this is handled in the next round of dep updates. Thanks.

I appreciate it!

akshay-joshi:
On Mon, Jan 16, 2023 at 3:48 PM Dave Page ***@***.***> wrote:

> Akshay, please ensure this is handled in the next round of dep updates.
> Thanks.
>
   Sure will check

anthraxx:
Justed wanted to note that on Arch Linux we now run into the same issue, compatibility with 4.0.x would be greatly appreciated.

dpage:
[...]

Issue #4728, opened 2021-10-26
Include suport to set custom connection options (RM #6961)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/6961***
*Originally created by **Israel Lins Albuquerque** at 2021-10-26 13:45:07 UTC.*

This may able to change any option on connection string used by libpq.

rafapaisley:
This feature would be very useful for us as well...
Being able to specify "target_session_attrs=primary" / "target_session_attrs=standby" as in psql or "targetServerType=primary" / "targetServerType=seconday" as in jdbc would be much appreciated

dvasdekis:
Please see issue here: https://github.com/pgadmin-org/pgadmin4/issues/5547
As identified by [adityatoshniwal](https://github.com/adityatoshniwal), adding parameters like `--cluster%3Dgolden-mermaid-3114` would allow connection to CockroachDB Cloud, a free distributed cloud database.
Thanks!

nikhil-mohite:
Getting error on register server with Password file

1. Open the register server dialog, provide basic info name, Host address and enable Kerberos Authentication? (Connect Now? should be enabled )
3. In the Parameters tab add `Password file` field and provide value
4. now click on the save button, getting the following error
<img width="703" alt="image" src="https://user-images.githubusercontent.com/69510038/215097908-8626c320-20be-4be5-9221-915e5ae415be.png">
<img width="794" alt="image" src="https://user-images.githubusercontent.com/69510038/215098004-587f1cbd-7e1d-4617-b13c-7a36f799b177.png">

akshay-joshi:
The issue is fixed.

yogeshmahajan-1903:
While upgrading from pgAdmin 4.15 to latest snapshot, migration is failing with below error.

PFA SS.
<img width="1790" alt="Screenshot 2023-01-30 at 11 42 30 AM" src="https://user-images.githubusercontent.com/16016019/215722877-d3c19ab6-99ce-438a-b5fa-a148c65435ed.png">
```
2023-01-30 11:59:58,604: INFO   pgadmin:        ########################################################
2023-01-30 11:59:58,604: INFO   pgadmin:        Starting pgAdmin 4 v6.19...
2023-01-30 11:59:58,604: INFO   pgadmin:        ########################################################
[...]

PR #5769, opened 2023-01-23
1) Added support for setting PostgreSQL connection parameters. #4728

akshay-joshi:
2) Fixed an issue where Kerberos authentication to the server is not imported/exported. #5732 
3) Increase the length of the value column of the setting table. #5746 
4) Upgrade Flask-Migrate to 4.0.0. #5525


akshay-joshi:
Impacted areas:

- Migration of the pgAdmin configuration database. server table gets updated, check for SQLite and external database both. 
- SSL Tab, Passfile, Host Address, and Connection timeout from the server dialog.
- Connect/Disconnect/Reset to the PG/EPAS server.
- The connection String should be visible on the Properties tab.
- Process(Backup, Maintenance, Restore), PSQL module
- Import/Export servers from GUI and Command line with old exported JSON files and new files.
- Documentation.

web/migrations/versions/f656e56dfdc8_.py
```diff
@@ -0,0 +1,98 @@
...
+        batch_op.drop_column('hostaddr')
+        batch_op.drop_column('passfile')
+        batch_op.drop_column('connect_timeout')
+
+
+def upgrade():
+    migrate_connection_params('server')
+    migrate_connection_params('sharedserver')
+
+    with op.batch_alter_table("setting") as batch_op:
```
adityatoshniwal, line 89: Please add one comment here on what is the operation for. I also recommend using Text data type to avoid length issues.

web/pgadmin/browser/server_groups/servers/static/js/variable.ui.js
```diff
@@ -67,24 +73,64 @@ export default class VariableSchema extends BaseUISchema {
     });
   }
 
+  getPlaceHolderMsg(variable) {
+    let msg = '';
+    if (variable?.min_server_version && variable?.max_server_version) {
+      msg = gettext(variable?.max_server_version + ' <= Supported version >= ' + variable?.min_server_version );
```
adityatoshniwal, line 79: Please use the syntax `gettext('%s <= Supported version >= %s', var1, var2)`

web/pgadmin/misc/cloud/utils/__init__.py
```diff
@@ -48,9 +48,8 @@ def _create_server(data):
         name=data.get('name'),
         maintenance_db=data.get('db'),
         username=data.get('username'),
-        ssl_mode='prefer',
         cloud_status=data.get('cloud_status'),
-        connect_timeout=30,
+        connection_params={'sslmode': 'prefer', 'connect_timeout': 30}
```
adityatoshniwal, line 52: I think `{'sslmode': 'prefer', 'connect_timeout': 30}` can be made common const.

web/pgadmin/model/__init__.py
```diff
@@ -71,6 +73,23 @@ def process_result_value(self, value, dialect):
             return value
 
 
+class PgAdminJSONString(types.TypeDecorator):
+    """
```
adityatoshniwal, line 77: Probably add one line info here.

web/pgadmin/tools/psql/__init__.py
```diff
@@ -394,6 +394,18 @@ def get_conn_str_win(manager, db):
     manager.export_password_env('PGPASSWORD')
     db = db.replace('"', '\\"')
     db = db.replace("'", "\\'")
+
+    hostaddr = manager.get_connection_param_value('hostaddr')
```
adityatoshniwal, line 398: `get_connection_param_value` can be modified to return connection param dict if param name is not passed.
For eg - `manager.get_connection_param_value()` will return all the conn params. And then the returned dict can be used to get params. This way, multiple `get_connection_param_value` calls can be avoided.

web/pgadmin/tools/psql/__init__.py
```diff
@@ -404,33 +416,32 @@ def get_conn_str_win(manager, db):
...
+    if get_complete_file_path(sslrootcert):
         conn_attr = " {0} sslrootcert='{1}'".format(
-            conn_attr, get_complete_file_path(manager.sslrootcert))
+            conn_attr, get_complete_file_path(sslrootcert))
 
-    if get_complete_file_path(manager.sslcrl):
+    if get_complete_file_path(sslcrl):
         conn_attr = " {0} sslcrl='{1}'".format(
-            conn_attr, get_complete_file_path(manager.sslcrl))
+            conn_attr, get_complete_file_path(sslcrl))
```
adityatoshniwal, line 445: We don't pass all the params to psql like we do pgAdmin psycopg2 connect? Shouldn't we use `create_connection_string` here as well?

adityatoshniwal (changes requested):
The code changes looks good to me except few mentioned.

web/migrations/versions/f656e56dfdc8_.py
akshay-joshi: Comment added.

web/pgadmin/browser/server_groups/servers/static/js/variable.ui.js
akshay-joshi: Done

web/pgadmin/tools/psql/__init__.py
akshay-joshi: For this specific file, I have removed the function completely so no need to fix the above as in the other places only one function call is there. I am keeping this as it is.

web/pgadmin/tools/psql/__init__.py
akshay-joshi: Fixed. Call create_connection_string

adityatoshniwal (approved):

Issue #5794, opened 2023-01-30
Replace the ChartJS charting lib


adityatoshniwal:

adityatoshniwal:
https://www.npmjs.com/package/echarts Looks like the lib to go with. It is comparatively faster than Chart.js and has all the features which Chart.js provides.

adityatoshniwal:
Please test on windows for CPU usage.

khushboovashi:
Tested on candidate build, working fine on Mac and Windows 12 Server.

PR #5821, opened 2023-02-03
Use uplot for Dashboard graphs to reduce CPU usage. #5794

adityatoshniwal:


khushboovashi (commented):
- Tooltip should be shifted to left if the data-point is at the most right position. Currently tooltip is getting cut for the most right data-points.
- I believe lazy loading of chartjs is not possible, so not implemented or I have missed that part.
- Code looks good to me.

adityatoshniwal:
@khushboovashi I will work on the tooltip fix.
I think it should be possible to lazy load chartjs, but considering the time left for release I thought of doing it later separately.

Issue #5685, opened 2022-12-29
Grant column permission to a view doesn't show up in the sql properties


geohuz:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Create a view and assign the permission:
```
GRANT SELECT(title) ON TABLE api.user_articles TO app_user;
```

however this column permission doesn't show up in the SQL tab.

**To Reproduce**

create any view and grant column permission

**Expected behavior**

The ```GRANT SELECT(title) ON TABLE api.user_articles TO app_user;``` must be shown on the SQL tab of the view.

**Error message**

If possible, include the full, unedited text of any error messages displayed.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [OSX monterey]
 - Browser [chrome]

**Additional context**

The server is running postgresql version 15.

adityatoshniwal:
Hi @geohuz,

Can you try taking DB backup in plain format and the check backup sql file? If the GRANT statement is present in the backup SQL file then we can say that its a bug in pgAdmin. Backup uses PostgreSQL's pg_dump.
<img width="743" alt="image" src="https://user-images.githubusercontent.com/31855755/209913664-32e96e6d-124d-44a5-a2f2-e78a772e3721.png">

adityatoshniwal:
Related to #5269

nikhil-mohite:
Resolved, and verified on candidate build 6.21: https://developer.pgadmin.org/builds/2023-03-01-1/
OS: macOS Monterey 12.6
Mode: Desktop

PR #5838, opened 2023-02-08
Ensure that Grant column permission to a view is visible in the SQL tab. #5685

akshay-joshi:


adityatoshniwal (approved):
The code looks good to me.

Issue #5758, opened 2023-01-18
pgAdmin 6.19 can't resize left (vertical) windows


ddjuric1:
In pgAdmin 6.19 we can't resize left (vertical) windows where connections,databases,schemas and object are.
Looks like the left to right 1/3 ratio is fixed as a fix for the flickering issue. 
Could it that ratio of 1/3 be parametrized? It would be very convenient.

adityatoshniwal:
@ddjuric1 It can be resized. Did you lock your layout using File->Lock Layout->Full lock?

ddjuric1:
@adityatoshniwal you right. It can be resized. But, I tried to unlock it trough Preferences options, and it did,t do that way (probably after restarting pgAdmin it do). It works if you unlock it through File->Lock Layout immediately. 

Thx

Regards

adityatoshniwal:
I can confirm its a bug. Changing the preference should immediately reflect.
Thanks.

anilsahoo20:
Verified. Tested on Snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-02-16/

Environment: macOS Monterey 12.6.3(Desktop Mode)

anilsahoo20:
Verified on candidate build 6.21: https://developer.pgadmin.org/builds/2023-03-01-1/
OS: macOS Monterey 12.6.3
Mode: Desktop

PR #5844, opened 2023-02-09
Fixed an issue where lock layout menu was not in sync with preferences. #5758

pravesh-sharma:


web/pgadmin/browser/templates/browser/js/utils.js
```diff
@@ -20,7 +20,7 @@
   {% endif %}label: "{{ item.label }}", applies: ["{{ key.lower() }}"],
   priority: {{ item.priority }},
   enable: "{{ item.enable }}",
-  {% if item.checked is defined %}checked: {% if item.checked %}true{% else %}false{% endif %},
+  {% if item.checked is defined %}checked: {% if item.name == 'mnu_lock_'+current_ui_lock %}true{% else %}false{% endif %},
```
adityatoshniwal, line 23: There should be one more condition for `{% if item.checked %}`

web/pgadmin/preferences/static/js/components/PreferencesComponent.jsx
```diff
@@ -574,6 +574,23 @@ export default function PreferencesComponent({ ...props }) {
             gettext('Later')
           );
         }
+        /*
```
adityatoshniwal, line 577: Change the comment:
`Sync the lock layout menu with preferences`

adityatoshniwal (changes requested):
Otherwise, it looks good to me.

web/pgadmin/browser/templates/browser/js/utils.js
pravesh-sharma: iten.checked will either be true or false and  was earlier used  to set checked property to true or false, I think adding item.checked will not have any effect. please let me know if I am missing any point here.

Issue #5842, opened 2023-02-09
Additional logging for login for keeping track of who is using shared pgadmin instance


paulmilbank:
**Describe the solution you'd like**

I would like additional logging around the logins so I can keep track of users who are logging in to a shared PGAdmin instance

**Describe alternatives you've considered**

I currently replace the internal.py and oauth2.py files here https://github.com/pgadmin-org/pgadmin4/tree/master/web/pgadmin/authenticate with ones that contain the additional logging as part of our container build process. This breaks every few releases as I need to update my changes against the latest pgadmin code.

**Additional context**

I am happy to make a PR for this if that would be useful. I have added my logging at info level and the default level seems to be warning. I believe this means most users will not see the additional logging.

adityatoshniwal:
Hi @paulmilbank,

You can raise the PR. We can review and merge it.

Thanks.

paulmilbank:
Thanks very much, PR raised now.

yogeshmahajan-1903:
This is working fine.

PR #5845, opened 2023-02-09
Add additional logging for successful logins and user creation

paulmilbank:
I'm not sure where the best place to add this is for the Kerberos source as it doesn't define a login method like the other auth sources do. 

This is to resolve issue https://github.com/pgadmin-org/pgadmin4/issues/5842


adityatoshniwal (approved):
Looks good to me.

akshay-joshi:
@paulmilbank Please fix PEP8 issue.

PR #5853, opened 2023-02-11
Remove dependency on simplejson package.

mcepl:
I don't think anybody needs compatiblity with Python 2.6 any more.

The only difference in API of the standard library json methods and the ones from simplejson is attribute `encoding`, which has been removed in the standard library (for good reasons). On the other hand, the standard library now accepts both `str` and `bytes`, so encoding is not necessary that much (and if it is necessary, it should be done in the class specific encoder anyway).

The only change in API is that `web.pgadmin.utils.ajax.make_json_response()` function lost its `encoding` parameter. It is never used in all pgadmin4 anyway and given that `DataTypeJSONEncoder` produces pure Python `str` anyway, I don't think it is a big loss.

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -40,7 +40,6 @@ Flask-SocketIO                                                   5.2.0          
 WTForms                                                          3.0.1            BSD-3-Clause                         https://wtforms.readthedocs.io/
 passlib                                                          1.7.4            BSD                                  https://passlib.readthedocs.io
 pytz                                                             2021.3           MIT                                  http://pythonhosted.org/pytz
-simplejson                                                       3.18.3           MIT License                          https://github.com/simplejson/simplejson
 speaklater3                                                      1.4              UNKNOWN                              https://github.com/ThomasWaldmann/speaklater
 sqlparse                                                         0.4.3            BSD-3-Clause                         https://github.com/andialbrecht/sqlparse
 psutil                                                           5.9.3            BSD-3-Clause                         https://github.com/giampaolo/psutil
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,6 @@ Flask-SocketIO<=5.2.0
 WTForms==3.*
 passlib==1.*
 pytz==2021.*
-simplejson==3.*
 speaklater3==1.*
 sqlparse==0.*
 psutil==5.9.3
--- a/web/pgadmin/browser/server_groups/__init__.py
+++ b/web/pgadmin/browser/server_groups/__init__.py
@@ -9,7 +9,7 @@
 
 """Defines views for management of server groups"""
 
-import simplejson as json
+import json
 from abc import ABCMeta, abstractmethod
 
 from flask import request, jsonify, render_template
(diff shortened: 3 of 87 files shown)


khushboovashi:
@mcepl, 

Looks good to me, only some of the files/modules are still left like clouds, publications etc.
Also, remove encoding keyword argument from the make_json_response in sqleditor/__init__py file. Please check another instances too for the same.

Thanks,
Khushboo

mcepl:
> Looks good to me, only some of the files/modules are still left like clouds, publications etc. Also, remove encoding keyword argument from the make_json_response in sqleditor/__init__py file. Please check another instances too for the same.

Whoops, I missed those. Fixed.

khushboovashi:
> > Looks good to me, only some of the files/modules are still left like clouds, publications etc. Also, remove encoding keyword argument from the make_json_response in sqleditor/__init__py file. Please check another instances too for the same.
> 
> Whoops, I missed those. Fixed.

Thanks for the quick fix, but I still get error (make_json_response() got an unexpected keyword argument 'encoding') while trying to execute the query in the Query Tool.  So, you need to remove the 'encoding' parameter from make_json_response function in sqleditor/__init__.py file.

mcepl:
Those call were fixed. Are those attributes `encoding` in `test_import_export_create_job_unit_test.py` and `test_backup_create_job_unit_test.py`? Should I remove them as well?

khushboovashi:
> Those calls were fixed. Are those attributes `encoding` in `test_import_export_create_job_unit_test.py` and `test_backup_create_job_unit_test.py`? Should I remove them as well?

No, They are required, so do not remove those. 

With this patch, I am facing one issue, while saving/updating the non ascii data from the Sql Editor, I am getting the error saying  "Object of type bytes is not JSON serializable".

You can reproduce this issue by creating  a database with latin9 or latin8 (other than UTF8) encoding. Then try to save the non-ascii characters (ex: äöüßs).

mcepl:
> With this patch, I am facing one issue, while saving/updating the non ascii data from the Sql Editor, I am getting the error saying "Object of type bytes is not JSON serializable".

If there is a need for explicit encoding (which object have allowed non-UTF8 content?) with `.encode()/.decode()` methods of `str` and `bytes`.

khushboovashi:
> > With this patch, I am facing one issue, while saving/updating the non ascii data from the Sql Editor, I am getting the error saying "Object of type bytes is not JSON serializable".
> 
> If there is a need for explicit encoding (which object have allowed non-UTF8 content?) with `.encode()/.decode()` methods of `str` and `bytes`. Where did the crash happened?

Try the steps which I have mentioned to reproduce it.

mcepl:
> Try the steps which I have mentioned to reproduce it.

I have to admit a shameful thing: I have nothing to do with PostgreSQL, I don’t have it even installed. I am a maintainer of Python packages for SUSE/openSUSE (all 2000+ of them, together with two more colleagues), and this PR is driven mostly by my intent to get rid of `simplejson` package in our distro, I know almost nothing about PostgreSQL.

Issue #5833, opened 2023-02-06
user not removed from sqlite db pgadmin4.db - user_mfa table


mar555-ginger:
Hello, 

This was found on pgadmin 6.19, centos 7.9

Steps to reproduce: 
1. create a new user 
2. log in as that user
3. enable 2FA for the newly created user (at this point an entry will show up within user_mfa table) 
4. delete the user 
5. the entry will still be present in pgadmin4.db -> user_mfa table 

That lead for example to an interesting issue, like creating a new user, and the 2FA activation email being sent to the previous user email (which was deleted). 

 Thx

PR #5880, opened 2023-02-20
Fixed an issue where user mfa entry was not getting delete after deleting a user. #5833

pravesh-sharma:


web/pgadmin/tools/user_management/__init__.py
```diff
@@ -597,6 +597,9 @@ def delete_user(uid):
         SharedServer.query.filter(SharedServer.servergroup_id.in_(sg)).delete(
             synchronize_session=False)
 
+        # Delete MFA setting for current user.
+        UserMFA.query.filter_by(user_id=uid).delete()
```
adityatoshniwal, line 601: Shouldn't this be a delete cascade? user id must be having a FK on user table. So it should get deleted automatically on delete in user table.

adityatoshniwal (changes requested):

adityatoshniwal (changes requested):
The review was marked resolved without any change.

adityatoshniwal (changes requested):
Please use FK reference with delete cascade.

web/pgadmin/tools/user_management/__init__.py
```diff
@@ -597,6 +597,9 @@ def delete_user(uid):
...
+
```
adityatoshniwal, line 602: Use foreign key.

adityatoshniwal (changes requested):

web/pgadmin/tools/user_management/__init__.py
pravesh-sharma: Made the required changes, please review.

adityatoshniwal (approved):

PR #5915, opened 2023-03-06
Changes include:

adityatoshniwal:
- Update Flask, Flask-SQLAlchemy, Flask-Babel, Flask-Security-Too, Flask-SocketIO, pytz, psutil, SQLAlchemy, bcrypt, cryptography, eventlet, Authlib, requests python packages
- Remove pinned dnspython, Werkzeug packages from requirements.txt

### DIFF
--- a/web/config.py
+++ b/web/config.py
@@ -859,80 +859,8 @@ AUTO_DISCOVER_SERVERS = True
 #############################################################################
 SERVER_HEARTBEAT_TIMEOUT = 30  # In seconds
 
-##########################################################################
-# Local config settings
-##########################################################################
-# User configs loaded from config_local, config_distro etc.
-user_config_settings = {}
-
-
-# Function to Extract settings from config_local, config_distro etc.
-def get_variables_from_module(module_name):
-    module = globals().get(module_name, None)
-    variables = {}
-    if module:
-        variables = {key: value for key, value in module.__dict__.items()
-                     if not (key.startswith('__') or key.startswith('_'))}
-    return variables
-
-
-# Load distribution-specific config overrides
-try:
-    import config_distro
-    config_distro_settings = get_variables_from_module('config_distro')
-    user_config_settings.update(config_distro_settings)
-except ImportError:
-    pass
-
-# Load local config overrides
-try:
-    import config_local
-    config_local_settings = get_variables_from_module('config_local')
-    user_config_settings.update(config_local_settings)
-except ImportError:
-    pass
-
-# Load system config overrides. We do this last, so that the sysadmin can
-# override anything they want from a config file that's in a protected system
-# directory and away from pgAdmin to avoid invalidating signatures.
-system_config_dir = '/etc/pgadmin'
-if sys.platform.startswith('win32'):
-    system_config_dir = os.environ['CommonProgramFiles'] + '/pgadmin'
-elif sys.platform.startswith('darwin'):
-    system_config_dir = '/Library/Preferences/pgadmin'
-
-if os.path.exists(system_config_dir + '/config_system.py'):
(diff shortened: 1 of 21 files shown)


web/config.py
```diff
@@ -912,6 +912,14 @@ def get_variables_from_module(module_name):
             file_dir_name = os.path.basename(locals().get(setting))
             locals().update({setting: os.path.join(data_dir, file_dir_name)})
 
+# To use psycopg3 driver, need to specify +psycopg in conn URI
+if 'CONFIG_DATABASE_URI' in user_config_settings:
+    CONFIG_DATABASE_URI = user_config_settings['CONFIG_DATABASE_URI']
+    if CONFIG_DATABASE_URI.startswith('postgresql:'):
+        user_config_settings['CONFIG_DATABASE_URI'] = \
+            'postgresql+psycopg:{0}'.format(
+                CONFIG_DATABASE_URI[CONFIG_DATABASE_URI.find(':') + 1:])
```
akshay-joshi, line 921: We are removing psycopg2, so this logic is not required I guess. Instead, update the comments for "CONFIG_DATABASE_URI" in the same file. If in the future the string is changed then won't need to change the logic.

akshay-joshi (changes requested):
Documentation checks are failing.

web/config.py
adityatoshniwal: Even if we use only psycopg3, the `+psycopg` is required by SQLAlchemy to use psycopg3.

dpage: Logic like that shouldn't be in config.py. That file is supposed to have an absolute minimum of code in it - ideally, nothing more than platform specific conditionals to set default values accordingly ("if Windows then foo else bar" type stuff).

adityatoshniwal: @dpage config is imported at multiple places directly. Although CONFIG_DATABASE_URI is used at only one or two places, but if it is required anywhere then it has to be modified like above before use.
I can move it somewhere else if you say so.

dpage: Then maybe we need a wrapper of some kind that we import instead, then can load and manipulate settings as we need. Ultimately, config.py is supposed to be a configuration file, not application code. It just happens to use Python syntax...

adityatoshniwal: @dpage Or may be we create a new config_defaults.py which will have all the configurations of config.py and config.py (the wrapper) can have all the manipulations? This will make it work with minimum code changes.

akshay-joshi: I am still inclined toward removing this logic completely and updating the uses like "postgresql+psycopg://username:password@host:port/database" in the comment of the config.py which is already there. Just need to add "+psycopg" in each example.

adityatoshniwal: > I am still inclined toward removing this logic completely and updating the uses like "postgresql+psycopg://username:password@host:port/database" in the comment of the config.py which is already there. Just need to add "+psycopg" in each example.

@akshay-joshi This will immediately break working pgAdmin for external DB users upon upgrade. Also, users do not need to know `+psycopg` syntax, that is specific to SQLAlchemy. pgAdmin should take care of it.

adityatoshniwal: @dpage Creating evaluated_config.py as a wrapper.

adityatoshniwal: @dpage If we want to keep config.py clean then we can move the manipulation code to some other file and just call it inside config.py?

adityatoshniwal: pushed the change - https://github.com/pgadmin-org/pgadmin4/pull/5915/files#diff-07a6e9f47f63fa72551fe22f3b87ad2935a0806843fd56351b3e5a6c7bc57c6dL862

PR #5965, opened 2023-03-16
Use correct python binary for Ubuntu 18 and EL-8.

akshay-joshi:


pkg/linux/build-functions.sh
```diff
@@ -18,6 +18,19 @@ _setup_env() {
     if [ -n "${APP_SUFFIX}" ]; then
         APP_LONG_VERSION=${APP_LONG_VERSION}-${APP_SUFFIX}
     fi
+
+    # Setting up the correct Python version for Ubuntu 18 and EL-8, which have
+    # new Python versions installed parallel to the old ones.
+    OS_VERSION=$(grep "^VERSION_ID=" /etc/os-release | awk -F "=" '{ print $2 }' | sed 's/"//g' | awk -F "." '{ print $1 }')
+    SYSTEM_PYTHON_PATH='/usr/bin/python3'
+    PYTHON_BINARY='python3'
+    if [ "${OS_VERSION}" == 18 ]; then
```
dpage, line 27: As this is in the shared functions file, we can't assume it's being run on Ubuntu or EL. The test should look more like:

if [ "$1" == "redhat" ] && [ "${OS_VERSION}" == "8" ]; then

dpage: Additionally, you're comparing a string to an int. Quote the 18 :-)

pkg/debian/build.sh
```diff
@@ -38,7 +37,7 @@ Version: ${APP_LONG_VERSION}
 Architecture: ${OS_ARCH}
 Section: database
 Priority: optional
-Depends: python3, libpq5 (>= 11.0), libgssapi-krb5-2
+Depends: ${PYTHON_BINARY}, libpq5 (>= 11.0), libgssapi-krb5-2
```
dpage, line 40: How safe is this, considering PYTHON_BINARY is being set in a child function?

pkg/redhat/setup.sh
```diff
@@ -37,10 +37,10 @@ curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | sudo tee /etc/yu
 echo "Installing build pre-requisites..."
 yum groupinstall -y "Development Tools"
 
-if [ "${OS_VERSION}" == 7 ]; then
-    yum install -y expect fakeroot httpd-devel postgresql15-devel python3-devel nodejs yarn rpm-build rpm-sign yum-utils krb5-devel
-    pip3 install sphinx
-    pip3 install sphinxcontrib-youtube
+if [ "${OS_VERSION}" == 8 ]; then
```
dpage, line 40: This should also be properly quoted to avoid comparing a string and an int (I realise the old code was also broken).

dpage (changes requested):

pkg/debian/build.sh
akshay-joshi: We have set the PYTHON_BINARY  in the '**_setup_env**' function where other variables are set like {SERVERROOT}, {DISTROOT}, {APP_NAME} etc. I thought if those are being set, so PYTHON_BINARY should also be set.

dpage: Good point. No problem then.

Issue #5834, opened 2023-02-07
pgAgent: The Job is not being dropped through properties 


FaharAbbasRizvi:
- OS: [All]
 - Version: [6.20]
 - Mode: [Desktop and Server]
 - Package type: [All]
 
The Job is being dropped through properties of pgAgent while the user can drop it through right click

Steps to reproduce the behaviour:

1. Install pgAgent
2. Launch pgadmin4
3. Go to the Job node
4. create a new job
5. job created
6. Click on Properties
7. Click on Delete/Drop
8.  Click yes to drop the node
9.  job is not dropped

**Expected behaviour**

The job should be dropped successfully from properties 

**Error message**

Job is still displayed if the user tries to drop through properties 

Note:
Please note that this is also reproducible in older releases.
<img width="1011" alt="Screen Shot 2023-02-07 at 4 10 20 PM" src="https://user-images.githubusercontent.com/43006122/217231244-897b7f25-fe04-45b0-b2c1-ad7ff623d9c6.png">
<img width="1014" alt="Screen Shot 2023-02-07 at 4 10 33 PM" src="https://user-images.githubusercontent.com/43006122/217231264-aec3503e-2c9e-4f24-8d77-64e7b65524a9.png">

FaharAbbasRizvi:
This is tested in the latest snapshot build.
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-03-20/

PR #5966, opened 2023-03-17
Fixed issue where pgAgent jobs were not getting dropped from properties tab. #5834

pravesh-sharma:


web/pgadmin/misc/properties/CollectionNodeProperties.jsx
```diff
@@ -133,7 +133,9 @@ export function CollectionNodeView({
         _type: row.original._type,
       }));
     } else {
-      selRows = selRowModels.map((row) => row.original.oid);
+      let nodeObj = pgBrowser.Nodes[itemNodeData?._type.replace('coll-', '')];
+      let schema = nodeObj.getSchema?.call(nodeObj, treeNodeInfo, itemNodeData);
```
adityatoshniwal, line 137: `schema` is already been set at one more place. Use the same schema obj. Do not recreate the logic.

adityatoshniwal (changes requested):

web/pgadmin/misc/properties/CollectionNodeProperties.jsx
pravesh-sharma: The only place schema is being used is inside the useEffect where table columns are being created, so I reused that logic. I'll check again if we have reference to schema at any other place which can be reused.

adityatoshniwal: The useEffect schema can be re-used. Create a reference to it using `const schemaRef = useRef()` and then `schemaRef.current = <code>`. Then you can use `schemaRef` directly.

pravesh-sharma: Please review now.

adityatoshniwal (approved):

web/pgadmin/misc/properties/CollectionNodeProperties.jsx
```diff
@@ -265,7 +266,7 @@ export function CollectionNodeView({
           );
         });
     }
-  }, [itemNodeData, node, item, reload]);
+  }, [itemNodeData, node, item, reload, schemaRef.current]);
```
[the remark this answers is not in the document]
adityatoshniwal: You can mark this as resolved.

PR #5983, opened 2023-03-23
Run SonarQube scans on pushes to the master branch.

dpage:


adityatoshniwal:
@dpage How is it different from what is currently happening? SonarQube is run even now on commit to master.

dpage:
It's no different in practice. The intent is:

1) To include the action in the main source tree, so others can run it too without needing our Jenkins infrastructure and jobs.
2) To reduce the number of jobs run on the build farm; specifically at the moment I'm trying to get all code and dependency analysis into Github; it makes sense to keep those tasks there, with the code, and use the build farm for, building.

Once this is merged, I'll remove the Jenkins SonarQube job.

akshay-joshi:
@dpage 

If the scan failed, will it blocks the PR to squash and merge?

Issue #5934, opened 2023-03-10
Cannot save row when only columns changed have default value and are blanked.


simon-kempster:
**Describe the bug**

I have a table where various columns have a default value, both integer and character types.  If I remove the existing value from a column, it shows "[default]" in the grid display.  Trying to save the changes results in an error.

**To Reproduce**

Steps to reproduce the behaviour:
1. Go to Query Tool.
2. Select some rows from a table that has some columns defined with default values.
3. Edit a column with a default value by clearing its existing value.
4. Try to save the changes to the table.
5. Error shown: syntax error at or near "WHERE" LINE 2: WHERE

**Expected behaviour**

On saving the row, the column with the default value should be assigned the default value.

**Error message**

Error text as shown above.  From the log, there is a UPDATE command, example:
UPDATE table SET WHERE primary_key = value;
Clearly, the column(s) to be updated are missing, hence the error.

**Screenshots**

Not applicable.

**Desktop (please complete the following information):**
 - OS: Windows Server 2019 1809 build 17763.4010
 - Version: 6.21
 - Mode: Desktop
 - Package type: Windows EXE.

**Additional context**

If any columns without a default value are modified at the same time, then the row is saved correctly.  The generated UPDATE statement refers only to the columns without the default values.

anilsahoo20:
Verified on latest snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-03-28/
Environment: macOS Monterey 12.6.3

PR #6006, opened 2023-03-24
Ensure that default values are set only for insert statement if user does not provide any values, in case of updating existing values to blank it should be set to null. #5934

pravesh-sharma:


web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/insert.sql
```diff
@@ -4,7 +4,7 @@ INSERT INTO {{ conn|qtIdent(nsp_name, object_name) | replace("%", "%%") }} (
 {% if not loop.first %}, {% endif %}{{ conn|qtIdent(col) | replace("%", "%%") }}{% endfor %}
 ) VALUES (
 {% for col in data_to_be_saved %}
-{% if not loop.first %}, {% endif %}%({{ pgadmin_alias[col] }})s{% if type_cast_required[col] %}::{{ data_type[col] }}{% endif %}{% endfor %}
+{% if not loop.first %}, {% endif %}{% if data_to_be_saved[col] == 'set_default'%} DEFAULT {% else %}%({{ pgadmin_alias[col] }})s{% if type_cast_required[col] %}::{{ data_type[col] }}{% endif %}{% endif %}{% endfor %}
```
adityatoshniwal, line 7: Why are you changing the backend code? This was handled in front end and can be fixed in front end itself.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/insert.sql
pravesh-sharma: I had discussion with akshay and nikhil regarding this. We need at least one value to be provided from front end for insert to work so, if a user does not provide any values insert will fail so we need to handle it in backend only.

adityatoshniwal: @pravesh-sharma OK understood.
But I think this code might not work in case where:
1. I have a char varrying column with some default value
2. I add a row, and set that column data to `set_default` in front end and save.
3. It will take default value instead of `set_default`.
Can you verify this scenario?

adityatoshniwal (changes requested):

adityatoshniwal (approved):

PR #6024, opened 2023-03-27
Nomenclature changes:

adityatoshniwal:
1. Rename "Properties..." context menu option of object to "Edit Object...".
2. Rename "Browser" tree to "Object Explorer".


akshay-joshi (changes requested):
Document and Screenshot updates are required.

PR #6107, opened 2023-04-09
Implement an active-only-sessions view, per request on Github 3831 (RM 5844)

cdwiegand:
This has been bugging me for years, I just wanted to only show active sessions sometimes to then filter/sort by database, and identify my problem queries quickly.


cdwiegand:
Having to work through significant issues with JSX (or whatever frontend language that is) and the python, so closing the PR for now until I have something that doesn't blow up for me.

Issue #2078, opened 2018-09-17
Statusbar to see the actual database (+comment?) (RM #3652)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/3652***
*Originally created by **Durum Dara** at 2018-09-17 11:56:21 UTC.*

Dear Developers!

It would be very nice if we can see the actual selected database name (+ comment) in a statusbar.

The user can disable/enable it in preferences.

It could be a normal statusbar at the bottom (like in LibreOffice, Firefox, etc.) which shows the actual database (+ comment). 

Many times I didn't see what is the actual selected database (if I scrolled up / down, or the metadata list is very-very long - too many tables, generator, etc.).

Thank you for it!

FaharAbbasRizvi:
I am facing a problem while creating a sequence:
2023-05-03 12:25:15,729: ERROR	pgadmin:	'description'
Traceback (most recent call last):
  File "/usr/pgadmin4/venv/lib64/python3.9/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/pgadmin4/venv/lib64/python3.9/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/usr/pgadmin4/venv/lib64/python3.9/site-packages/flask/views.py", line 107, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)
  File "/usr/pgadmin4/web/pgadmin/browser/utils.py", line 311, in dispatch_request
    return method(*args, **kwargs)
  File "/usr/pgadmin4/web/pgadmin/browser/server_groups/servers/databases/schemas/sequences/__init__.py", line 157, in wrapped
    return f(self, *args, **kwargs)
  File "/usr/pgadmin4/web/pgadmin/browser/server_groups/servers/databases/schemas/sequences/__init__.py", line 246, in nodes
    description=row['description']
KeyError: 'description'

FaharAbbasRizvi:
@adityatoshniwal Comments are not displayed in the materialized view as well.

FaharAbbasRizvi:
<img width="1018" alt="Screen Shot 2023-05-03 at 12 47 32 PM" src="https://user-images.githubusercontent.com/43006122/235859009-d742f202-d8aa-411e-85b9-2a5086e59c2e.png">

[...]

PR #6153, opened 2023-04-19
Show object breadcrumbs path along with its comment on object hover. …

adityatoshniwal:
…#2078


akshay-joshi (changes requested):
1. Rename any node and hover it. Comments are not visible. 
2. In case of a long name or comment ellipses (...) should be visible.
3. Create an index/column (any child node under the table) of a table with any name, then rename it to '>>>>' and hover over the node, Name is showing incorrect.
4. In High Contrast Theme if we hover over the icon of any object node then the label is disappearing.

Issue #5922, opened 2023-03-08
Clear PK on Copy-and-Paste of table rows


usernamesareaholdoverfromthe90s:
Test case:

1. Drill down to any table (with existing rows) that has an auto-generated PK (e.g., determined by a Sequence)
2. Right-click on table, and select "View/Edit Data" | "All rows"
3. Select any row in the results list
4. Click the "Copy" icon (above the results list)
5. Click the "Paste" icon (above the results list)
6. Find the newly-pasted row (see bug #5921)

Note, that the copied row's PK has been included in the newly-pasted row.

So the first thing you have to do, EVERY TIME you copy and paste a row, is select the PK in the new row, and delete it.  Otherwise, when you click "Save Data Changes" you'll get a small error notification that the row could be saved because the PK is already in use.

Could you please change the paste behaviour to blank out the PK, since it always needs to be blanked out, anyway?

FaharAbbasRizvi:
This is resolved in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-04-27/

PR #6166, opened 2023-04-21
Ensure that when copy pasting a row primary key column is cleared and if column is serial type then copied value should be cleared and default value is used. #5922

### LINKED ISSUE #3027: copy and paste on table with column with type 'SERIAL' (RM #4782)
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/4782***
*Originally created by **André Burkhardt** at 2019-09-26 08:39:47 UTC.*

In pgAdmin III, when copy and
paste a row with columns with Type SERIAL, a messagebox with question pops up:

"This table contains columns from type 'SERIAL'. Use values from within
clipboard for this column? YES/NO"

If select NO, the column with the SERIAL is cleared after
insert. When saving the changes, the field is filled with next value
from sequence.

In pgAdmin 4 the dedection of such SERIAL-columns is not implemented, so
that you have to clear the column with SERIAL manually. Otherwise there are
double values in that column.

Is it on roadmap to implement this useful feature in pgAdmin 4?

pravesh-sharma:
This also fixes #3027


web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -953,6 +953,15 @@ def poll(trans_id):
                 if columns_info is not None:
                     # Only QueryToolCommand or TableCommand can be editable
                     if hasattr(trans_obj, 'obj_id') and trans_obj.can_edit():
+                        st, res = fetch_sequences_data(trans_obj)
```
adityatoshniwal, line 956: Why do we need separate SQL query for sequence type? This should be taken care in `get_columns_types` and in the one SQL.

web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/sequences.sql
```diff
@@ -0,0 +1,10 @@
+SELECT DISTINCT att.attname as name, att.attnum as OID, seq.seqtypid
```
adityatoshniwal, line 1: This can be clubbed with get_columns_types. I don't think we need a separate SQL for this.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -1153,6 +1162,37 @@ def fetch_all_from_start(trans_id, limit=-1):
     )
 
 
+def fetch_sequences_data(trans_obj):
+    """
+    This methos is used to fetch the owned sequence by the table
```
dpage, line 1167: Methos?

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
```diff
@@ -560,12 +562,17 @@ export class ResultSetUtils {
...
+      // if column has default value
       if(col.has_default_val) {
         columnVal = undefined;
-      } else if(rawCopiedVal === null) {
+      } else if(rawCopiedVal === null || col.isPK) {
         columnVal = null;
       }
+    } else if (col.has_default_val && col.seqtypid) {
+      // if column has default value and is serieal type
+      columnVal = undefined;
```
dpage, line 575: It seems to me that this should behave as pgAdmin III did, and ask the user what they want to do. What if I specifically want to paste the original value(s)? That is perfectly valid thing to do in some circumstances.

khushboovashi: In which situation, the user wants to paste the original value(s)? It may be a very rare situation and for that do we need to ask every time while pasting whenever we have a serial type?

web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/sequences.sql
adityatoshniwal: BTW, why didn't we use `pg_get_serial_sequence` to check if it is a serial column?

pravesh-sharma: Had discussion with Khushboo for this. If we use that function we will have to call it for all the columns, but using the query we get all the details at once.

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
dpage: Copying data from one table to another, maybe from a test system to prod or vice versa. Or when testing something, to re-populate some deleted data by copying original values back from. a spreadsheet or similar.

I wouldn't object to having a "Don't ask me again" option or similar.

web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/sequences.sql
adityatoshniwal: If you use that function in the nodes.sql query of the column by just adding `pg_get_serial_sequence`, you will get for all the columns at once right?

pravesh-sharma: As per the documentation `pg_get_serial_sequence ( table text, column text )` we will be requiring to pass the table name and column name in text format.

web/pgadmin/tools/sqleditor/__init__.py
pravesh-sharma: I have made changes in the function you have mentioned and it's working, I just modified the nodes.sql to fetch the sequences data.

adityatoshniwal: I cannot see the latest code somehow. Did you commit?

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
khushboovashi: Or we can have the option "Copy Serials" the same as "Copy with headers".

adityatoshniwal: > Or we can have the option "Copy Serials" the same as "Copy with headers".

with default as false, I agree. Better than popup.

web/pgadmin/tools/sqleditor/__init__.py
pravesh-sharma: I haven't, will commit it with front end changes. If you want I can commit that now.

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
pravesh-sharma: @dpage I have added an option in paste, attached screenshot

Screenshot 2023-04-24 at 2 06 50 PM

By default this option will be unchecked and if a user selects this option and then paste serial values will be pasted. Any suggestion/thoughts ?

dpage: Is "serials" the best wording here? Maybe "Paste with SERIAL/IDENTITY values?"

pravesh-sharma: Updated screen 
Screenshot 2023-04-24 at 2 52 45 PM

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
```diff
@@ -559,12 +561,25 @@ export class ResultSetUtils {
     return columns;
   }
 
-  processClipboardVal(columnVal, col, rawCopiedVal) {
-    if(columnVal === '') {
-      if(col.has_default_val) {
-        columnVal = undefined;
-      } else if(rawCopiedVal === null) {
+  processClipboardVal(columnVal, col, rawCopiedVal, pasteSerials) {
+    if (col.isPK) {
```
adityatoshniwal, line 565: Do we really need to check PK? If a PK is having a serial type then it will automatically taken care with serial handling code. If it is not a serial then we should not do anything with copied value.

web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/sequences.sql
adityatoshniwal, line 1: This file should be removed if not required.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
pravesh-sharma: We have two issues and in one of them user wants to clear the primary key when copy pasting that is the reason primary key check is there.

adityatoshniwal: As per the user `any table (with existing rows) that has an auto-generated PK (e.g., determined by a Sequence)`

pravesh-sharma: https://github.com/pgadmin-org/pgadmin4/issues/5922
If there is not default value or sequence being used then PK will be empty on paste.

adityatoshniwal: Why should the PK be empty if there no serial or sequence?
Also, both the tickets mention to clear only when there is a serial or auto generated val.

PR #6225, opened 2023-05-01
Implement an active-only-sessions view, per request on Github 3831 (RM 5844)

cdwiegand:
This has been bugging me for years, I just wanted to only show active sessions sometimes to then filter/sort by database, and identify my problem queries quickly.

I took feedback from prior PR, changed from a new tab to a checkbox within the CustomHeader of PgTable. Idea is to allow users to only show non-Idle connections, so we can then sort by other columns (e.g. database, connection time, etc.) without cluttering our display with Idle connections resulting from connection poolers or some ORMs/libaries.

### DIFF
--- a/web/pgadmin/dashboard/static/js/Dashboard.jsx
+++ b/web/pgadmin/dashboard/static/js/Dashboard.jsx
@@ -7,11 +7,12 @@
 //
 //////////////////////////////////////////////////////////////
 // eslint-disable-next-line react/display-name
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import gettext from 'sources/gettext';
 import PropTypes from 'prop-types';
 import getApiInstance from 'sources/api_instance';
 import PgTable from 'sources/components/PgTable';
+import { InputCheckbox } from '../../../static/js/components/FormComponents';
 import { makeStyles } from '@material-ui/core/styles';
 import url_for from 'sources/url_for';
 import Graphs from './Graphs';
@@ -151,6 +152,7 @@ export default function Dashboard({
   const [msg, setMsg] = useState('');
   const [tabVal, setTabVal] = useState(0);
   const [refresh, setRefresh] = useState(false);
+  const [activeOnly, setActiveOnly] = useState(false);
   const [schemaDict, setSchemaDict] = React.useState({});
 
   if (!did) {
@@ -794,6 +796,14 @@ export default function Dashboard({
     }
   }, [nodeData, tabVal, did, preferences, refresh, props.dbConnected]);
 
+  const filteredDashData = useMemo(()=>{
+    if (tabVal == 0 && activeOnly) {
+      // we want to show 'idle in transaction', 'active', 'active in transaction', and future non-blank, non-"idle" status values
+      return dashData.filter((r)=>(r.state && r.state != '' && r.state != 'idle'));
+    }
+    return dashData;
+  }, [dashData, activeOnly, tabVal]);
+
   const RefreshButton = () =>{
     return(
       <PgIconButton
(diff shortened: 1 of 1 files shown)


adityatoshniwal:
@cdwiegand Excellent work.
A question, shouldn't we simply filter active sessions in the front end data itself? That way we'll not need an extra API to manage and we can get switch between active or all session instantaneously by just using the checkbox.

cdwiegand:
I don't see a way for me to filter that in the <PgTable> using a JavaScript function - is there an example of that in pgadmin? That's why I used an API endpoint to do it. I tried using query strings but flask wouldn't pass it.

web/pgadmin/dashboard/static/js/Dashboard.jsx
```diff
@@ -833,6 +839,26 @@ export default function Dashboard({
...
+          e.preventDefault();
+          setActiveOnly(e.target.checked);
+          setRefresh(!refresh);
+        }}
+        value={activeOnly}
+        controlProps={CustomActiveOnlyHeaderLabel}
+      ></InputCheckbox>);
+  };
+    
   return (
```
adityatoshniwal, line 862: To filter on front end:
```
const finalDashData = useMemo(()=>{
  if(tabVal == 0 && activeOnly) {
    return dashData.filter((r)=>(r.state=='active'));
  }
  return dashData;
}, [activeOnly, tabVal])
```
and pass finalDashData to all the tables.

web/pgadmin/dashboard/static/js/Dashboard.jsx
```diff
@@ -833,6 +839,26 @@ export default function Dashboard({
...
+  const CustomActiveOnlyHeader = () => {
+    return (
+      <InputCheckbox
+        label={gettext('Active sessions only')}
+        labelPlacement="end"
+        className={classes.searchInput}
+        onChange={(e) => {
+          e.preventDefault();
+          setActiveOnly(e.target.checked);
+          setRefresh(!refresh);
```
adityatoshniwal, line 855: You'll not need setRefresh once the active only is filtered at frontEnd;

web/pgadmin/dashboard/static/js/Dashboard.jsx
```diff
@@ -833,6 +839,26 @@ export default function Dashboard({
     );
   };
 
+  const CustomActiveOnlyHeaderLabel =
```
adityatoshniwal, line 842: No need to create variable. Pass the object directly

web/pgadmin/dashboard/__init__.py
```diff
@@ -353,12 +357,26 @@ def get_data(sid, did, template, check_long_running_query=False):
...
+    if active_only:
+        res['rows'] = filter_active_only(res['rows'])
+
     return ajax_response(
         response=res['rows'],
         status=200
     )
 
 
+def filter_active_only(rows):
```
adityatoshniwal, line 369: As suggested, lets filter active only at front end;

web/pgadmin/dashboard/static/js/Dashboard.jsx
cdwiegand: Yeah I tried that on my side, and I could not get that to work reliably for me, which is why I ended up creating it as a variable so that I could get getText() to work correctly. I will review these changes, incorporate them into my branch, and then update my pull request. Thank you for your help with this.

cdwiegand:
I've updated my branch and squashed the commit/force pushed to update to only one commit, it works for me from a clean browser to check or uncheck and handle it entirely client-side.

web/pgadmin/dashboard/static/js/Dashboard.jsx
cdwiegand: this helped immensely, a little more digging and I had it working well. I ended up adding dashData to the array of monitored objects as I couldn't tell if that would inherently trigger (and my browser seems to not clear cache reliably, so it was going to take a long time to determine it manually).

adityatoshniwal:
@cdwiegand Very nice! The changes looks good to me. Will run the git workflows and merge if all good.

Issue #5777, opened 2023-01-24
Browser tree state is not remembered when reopening PGAdmin


natha1818:
Browser tree state is not recalled when reopening PGAdmin v6.19.
In PGAdmin v6.18 (and probably earlier), when you started PGAdmin it would expand browser tree to the location you left when quiting PGAdmin the last time.

Tested on following config:
PGAdmin version 6.19
OS: Windows 10 Pro version 22H2 (OS build 19045.2486)

Remark: similar to #4898

nikhil-mohite:
@natha1818 ,

can you please share a short video for this?

natha1818:
Here you have...

https://user-images.githubusercontent.com/28142574/215085671-158d4dc4-49ac-462c-af5b-082e55996f18.mp4

FaharAbbasRizvi:
This issue is still reproducible on the latest snapshot build.

https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-04-20/

PR #6270, opened 2023-05-10
Fixed issue where browser tree state is not remembered when reopening pgAdmin. #5777

pravesh-sharma:


khushboovashi (commented):
You should make this change along with `if (!pgBrowser.tree.hasParent(item) || !(this.parent in treeHierarchy))' this line.

pravesh-sharma:
> You should make this change along with `if (!pgBrowser.tree.hasParent(item) || !(this.parent in treeHierarchy))' this line.

I have made the changes.

PR #6276, opened 2023-05-11
Handle cases when psycopg returns bytes instead of a string.

dpage:
This will happen if the test database was initialised using SQL_ASCII encoding.

See https://github.com/psycopg/psycopg/issues/561


dpage:
Just saw a more appropriate fix suggested by Daniele. Will update.

Issue #6080, opened 2023-04-03
Windows 10 not showing pgAdmin icon on taskbar


nikhil-mohite:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Open pgAdmin 4 on windows it is not showing the pgAdmin 4 icon on the taskbar

**To Reproduce**
This issue is related to the latest version of NWjs (0.74) issue [https://github.com/nwjs/nw.js/issues/8039](https://github.com/nwjs/nw.js/issues/8039)

<img width="121" alt="image" src="https://user-images.githubusercontent.com/69510038/229447685-1d9260cf-7043-4bbd-a387-9ebe147feb3b.png">

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: latest snapshot: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-04-03/
 -  Mode: Desktop

garagusopdfs:
This happens also in the SDK, and in version 0.75

adityatoshniwal:
Please try the workaround suggested here - https://github.com/nwjs/nw.js/issues/8039#issuecomment-1528510777

FaharAbbasRizvi:
This is tested in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-05-18/

PR #6283, opened 2023-05-15
Fixed an issue where win 10 was not showing pgAdmin icon on taskbar. #6080

pravesh-sharma:


adityatoshniwal (changes requested):
Where are the changes for upgrading NW.js? We have pinned it at multiple places.

Issue #6147, opened 2023-04-18
Heartbeat is getting  logged, though no server is connected in pgAdmin


yogeshmahajan-1903:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

 Heartbeat is getting  logged, though no server is connected in pgAdmin

**To Reproduce**

1.Connect to server.
2.Check Network tab of dev tool to confirm heartbeat calls.(api call /log)
3.Now disconnect server.
4.Observer network tab.
5.'/log' heartbeat calls are on going. 

**Expected behavior**

When no server is connected, heartbeat api call should be stopped.

**Error message**

**Screenshots**

<img width="1792" alt="Screenshot 2023-04-18 at 2 09 00 PM" src="https://user-images.githubusercontent.com/16016019/232722025-135b39e9-1432-4c39-bf32-c7b6a87180e1.png">

**Desktop (please complete the following information):**
pgAdmin dev env.

**Additional context**

Issue is with below lines from browser.js
```
obj.Events.on(
        'pgadmin:server:disconnect', stop_heartbeat.bind(obj)
      );
```

nikhil-mohite:
Resolved, and verified on the latest snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-05-25/
OS: macOS
Mode: Server (Whl)

PR #6289, opened 2023-05-16
Ensure heartbeat signals are stopped on server disconnection.#6147

yogeshmahajan-1903:


web/pgadmin/browser/server_groups/servers/static/js/server.js
```diff
@@ -205,6 +205,12 @@ define('pgadmin.node.server', [
                   d = t.itemData(i);
                   t.removeIcon(i);
                   d.connected = false;
+
+                  // Generate the event that server is connected
```
khushboovashi, line 209: Please correct the tiny typo, connected should be disconnected.

yogeshmahajan-1903: Done

Issue #5048, opened 2022-05-04
Show only non-empty nodes/categories/groups in pgAdmin-4 (RM #7357)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/7357***
*Originally created by **Jan Zidek** at 2022-05-04 15:38:29 UTC.*

In the IntelliJ Idea's Database View, you can hide all the empty categories/nodes/groups by unchecking the `Show Empty Groups` switch.

In the following example, in my database the only existing objects are tables and functions. I am not disturbed by a lot of unused categories.

**Is it possible to achieve a similar effect in pgAdmin?** 

All I found in the settings is `Preferences` | `Nodes` | `Show/Hide nodes in the browser tree`, but these settings are static, they do not reflect the presence of objects in the categories.
!https://i.stack.imgur.com/uPpzU.png!

#### pgAdmin - always all categories
!https://i.stack.imgur.com/bisYV.png!

See also https://dba.stackexchange.com/q/311642/55104

FaharAbbasRizvi:
@adityatoshniwal 

If the user disabled "Show empty object collections?" and in the view child nodes are still displayed.

adityatoshniwal:
> @adityatoshniwal
> 
> If the user disabled "Show empty object collections?" and in the view child nodes are still displayed.

Please share some screenshot.

FaharAbbasRizvi:
<img width="1418" alt="Screen Shot 2023-05-25 at 10 52 08 AM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/15a8a374-db5e-463d-a397-cb5d6cddda53">

adityatoshniwal:
@FaharAbbasRizvi Test everything and then assign it to me with all the issues.

FaharAbbasRizvi:
@adityatoshniwal 

I have checked the pgAgent is still available if we disable the Show empty object collections? from the preferences, is this expected?

FaharAbbasRizvi:
@adityatoshniwal 

The schema node is still visible if the user drops the public schema and the node is empty if the user disables the *Show empty object collections?* from the preferences.

<img width="950" alt="Screen Shot 2023-05-25 at 3 21 03 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/d64a2164-ff6a-4dd3-bba0-575cebb09fc6">

FaharAbbasRizvi:
@adityatoshniwal 

[...]

PR #6294, opened 2023-05-18
Add an option to hide/show empty object collection nodes. #5048

adityatoshniwal:


adityatoshniwal:
> * Show a popup to refresh the browser tree when the value of "Show empty object collections" is changed to take effect.
Hmm. We do not show any popup if `Show System Objects` is changed. Will need to include those as well.
> * The following empty collection nodes are still visible even with the value of "Show empty object collections" set to false:
>   
>   1. Event Triggers
>   2. Indexes
>   3. RLS Policies
>   4. Rules
>   5. Triggers
>   6. Constraints
>   7. Types
>   8. Functions
>   9. Resource Groups
>   10. Synonym
>   11. Package -> Variables
>   12. Compound Triggers
> * On expanding the Subscriptions node got the below error:
>   Error Message: syntax error at end of input
>   LINE 2: WHERE sub.subdbid =
> * On expanding the Schemas node got the below error:
>   Error Message:missing FROM-clause entry for table "ct"
>   LINE 4:     AND ct.oid is NULL
> 
> **EPAS server getting the below error:**  AttributeError: type object 'SynonymView' has no attribute 'BASE_TEMPLATE_PATH'
> 
> On Refreshing the Schema node got following error:
> 
> Error Message:column "proisagg" does not exist LINE 11:     proisagg = FALSE HINT:  Perhaps you meant to reference the column "pr.prolang".
> 
> Error Message:column "s.synname" must appear in the GROUP BY clause or be used in an aggregate function LINE 5: ORDER BY synname;

I had skipped few nodes like columns, indexes, Resource Groups intentionally as I thought its good to show them. Will hide them.
However, I do not agree with hiding Package -> Variables because it will create confusion. Variables are part of package and even if empty should be shown.

Thanks.

PR #6297, opened 2023-05-19
POC: Feature/gitlab claim check

EvertonSA:
Hi colleagues,

I really like the project you guys are working on but I needed an additional authorization checking based on gitlab "Maintainer", "Developer" and "Owner" roles.

these are the additional oauth2 parameters I'm looking at:

```
          'OAUTH_GITLAB_AUTHZ_ENABLED': True,
          'OAUTH_GITLAB_GROUPS_OWNER_ROLE': '',
          'OAUTH_GITLAB_GROUPS_MAINTAINER_ROLE': '',
          'OAUTH_GITLAB_GROUPS_DEVELOPER_ROLE': '',
```

this way we can check if the user is a maintainer on a certain gitlab namespace and allow dev pgadmin to developer roles and prd pgadmin to maintainer roles.

---

this is a POC, I do not intend to enhance. I'm submitting this as I needed this and it has not been decided what to do with it. See https://github.com/pgadmin-org/pgadmin4/issues/4467

Issue #3166, opened 2019-12-03
Vault integration for dynamic secrets (RM #4989)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/4989***
*Originally created by **Tyler Hess** at 2019-12-03 19:05:32 UTC.*

I would like to be able to configure a vault instance instead of static database credentials for a database connection.

Basic workflow:
Vault Token:
-   A valid token
- -      root token
-  -     long ttl session token
-   Option to create a token based off valid vault login credentials.
-  -     oauth
-    -   ldap
-      - local login
-     -  etc 
Vault URL:
-   Much like a connection but would be used to hit vault api
Vault Path:
-   base path of where to start looking for database configurations and connections.
Look up connection details from vault api:
-   can look up what databases are connected to vault
- -      manual or automatically
-   permissions are locked down to permissions of the token
-   available databases could be populated through api calls
- -      either directed calls or automatic probing with lists and reads
-   connection details pulled from vault database configuration
- -      ip address / domain name
- -      port
- -      database name
-   dynamic credentials can be generated to connect to each database as needed.
Use that token to automatically generate temporary credentials and use those credentials to connect to the database.
 

Non-essential Features:
Having a warning when your session credentials are about to time out and the option to renew those credentials would be nice but not necessary.
Vault paths could need to be specified instead of dynamically generated
Manual search options could be proxied through pgadmin
-   Only needed if databases are not auto populated.
-   api calls used to call lists and reads to search for the correct paths so you don't have to pivot tools back and forth to vault while searching for a given path.
-   options to save api calls / results for future use. EG (you find the path needed to generate the db creds you want, you have an option to save that as a categorized call or use those results to connect to the database.)

62mkv:
any chance to prioritize this @akshay-joshi ? now that even some PR exists ... 🙏

adityatoshniwal:
> any chance to prioritize this @akshay-joshi ? now that even some PR exists ... 🙏

Hi @62mkv,

The PR raised is just a POC and cannot be merged. The author has clearly mentioned that he will not be working on it for any more changes. The pgAdmin team does not have bandwidth to work on this as there are many other important issues/features. If anyone wish to enhance the PR or work on the implementation then it is welcome.

Thanks.

dpage:
[...]

PR #6298, opened 2023-05-19
POC: Get password from Hashicorp Vault

EvertonSA:
Following https://github.com/pgadmin-org/pgadmin4/pull/6297, I needed a way to integrate pgadmin4 with hashicorp vault 

the following requirements are needed:

- vault jwt auth/policies configured with same provider as configured on oauth2 of pgadmin (does not necessarily needs to be gitlab, but auth role should be named database-generic)
- vault database STATIC role preconfigured and working
- k8s based deployment with shared server definitions using values.yaml, see https://github.com/rowanruseler/helm-charts/blob/10b73ffe7ae4bea566fc0a9df74168009f399b1f/charts/pgadmin4/values.yaml#L70
- add env var with VAULT_ADDR

image

-------

This is a POC. I'm "hacking" the code based on passexec feature, trying to change as little as possible. I rely on shared server definition and I need username and comment to be saved on init. I use db shared server definition comments to trigger vault get password. I do not intend to maintain neither enhance this code. I do not expect you to accept this MR. I'm submitting as a personal achievement and also hoping to help you to implement this better. See https://github.com/pgadmin-org/pgadmin4/issues/3166

Ideally, vault dynamic roles are preferred. But it was too hard to implement. I actually succeeded, but the code changes were too big.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -54,3 +54,4 @@ azure-identity==1.12.0
 google-api-python-client==2.*
 google-auth-oauthlib==1.0.0
 Werkzeug==2.2.3
+hvac==1.1.0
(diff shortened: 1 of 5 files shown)


adityatoshniwal:
A good reference if feature is picked for implementation. Thank you.
Closing.

Issue #5123, opened 2022-06-06
Standard desktop secret store integration for App Bundles (RM #7449)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/7449***
*Originally created by **Tamir Daniely** at 2022-06-06 08:31:35 UTC.*

Support standard desktop secret stores (OSX keychain, Windows Credential Store, Gnome Keyring...) as alternative configuration for master password mechanism in desktop app targets.

FaharAbbasRizvi:
@nikhil-mohite 
In the case of fresh mode, I think, the Cancel button should be available, please note that if I provide any random password it says incorrect master password, this needs to be updated. I have to provide the password again and again.

<img width="1256" alt="Screen Shot 2023-05-26 at 12 21 09 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/427740a4-34cd-426a-ba2b-28405ac415e1">
<img width="1253" alt="Screen Shot 2023-05-26 at 12 28 14 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/4fc29cc2-10c9-4c0c-b68f-547cadeea3c1">

<img width="1159" alt="Screen Shot 2023-05-26 at 12 43 27 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/9e0f2f33-cfc8-4c61-b475-6dcc1977e252">

FaharAbbasRizvi:
@nikhil-mohite 

I can see the master password dialog in the latest snapshot, I have tested on Ubuntu 23.04.
<img width="1257" alt="Screen Shot 2023-05-29 at 8 46 11 AM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/287277b2-9cd4-4f5f-8265-ade0482580a7">

nikhil-mohite:
> @nikhil-mohite
> 
> I can see the master password dialog in the latest snapshot on fresh installaion, I have tested on Ubuntu 23.04. <img alt="Screen Shot 2023-05-29 at 8 46 11 AM" width="1257" src="https://user-images.githubusercontent.com/43006122/241626693-287277b2-9cd4-4f5f-8265-ade0482580a7.png">

Hi @FaharAbbasRizvi, If pgAdmin is unable to use the OS password manager then it will use the Master password, this is expected behavior.

adityatoshniwal:
> > @nikhil-mohite
[...]

PR #6321, opened 2023-05-24
Improved the error messages text. #5123

nikhil-mohite:


web/pgadmin/static/js/Dialogs/index.jsx
```diff
@@ -156,27 +156,33 @@ function masterPassCallbacks(masterpass_callback_queue) {
 export function checkMasterPassword(data, masterpass_callback_queue, cancel_callback) {
   const api = getApiInstance();
   api.post(url_for('browser.set_master_password'), data).then((res)=> {
+    let is_keyring = res.data.data.keyring_name.length > 0;
```
adityatoshniwal, line 159: Why not simply use `Boolean(res.data.data.keyring_name)`? It will handle null/undefined/empty cases.

adityatoshniwal, line 159: please use camel case as var names - `isKeyring`. snake case is used in python.

web/pgadmin/static/js/helpers/ModalProvider.jsx
```diff
@@ -260,6 +260,9 @@ export const useModalStyles = makeStyles((theme) => ({
   iconButtonStyle: {
     marginLeft: 'auto',
     marginRight: '4px'
+  },
+  noteStyle: {
+    color: theme.otherVars.errorColor
```
adityatoshniwal, line 265: So we always show modal notes in error color? And why use errorColor at all? Its a note right?

adityatoshniwal (changes requested):

web/pgadmin/static/js/Dialogs/index.jsx
nikhil-mohite: Using is_keyring as Boolean flag, and null/undefined/empty is handled at backend.

Issue #6065, opened 2023-03-31
Few query tool shortcuts not working


nikhil-mohite:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

A few shortcuts in the query tool are not working 

**To Reproduce**
List of shortcuts   (keys can be different on user setup I am using default keys):
1. Connection status -> Accesskey + t : Not performing anything 
2. Execute options -> Accesskey + x: not working and also `showing wrong shortcuts` on UI
3. Find options -> Accesskey + n : not working
4. Row limits -> Acesskey + r :  : not working
5. Download Results is working only if we click on the Data output panel and press the shortcut, but if we are on the query tool tab it is not working (But other shortcuts like delete & paste rows work even query tool tab has selected no need to click  on Data output panel)
6. Shift Panel: this is working but it just highlights the panel title (In the older version if we switch panels from quay editor to scratch pad it is getting selected and the user can directly start typing now the cursor is not showing so user need to click to write)

**Expected behavior**

Shortcuts should work and should not show the wrong shortcut keys on the UI

**Desktop (please complete the following information):**
 - OS: OSX Ventura (13.3)
 - Version: Latest snapshot https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-03-31/
 - Mode: Desktop

FaharAbbasRizvi:
This is tested in the latest snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-06-21/

PR #6414, opened 2023-06-08
Fixed issue of query tool shortcuts not working. #6065

pravesh-sharma:
- Few shortcuts of query tool were not working that have been fixed in this PR. 
- Removed connection status shortcut from preferences as per discussion with team as it was not required.
- Renamed the `Find option` shortcut to `Edit options` and using this shortcut will open edit menu in query tool.
- Point 6 (Shift panel does not give focus to scratch pad) has not been fixed tried few things to fix it but did not work.

**Impact area** - All  shortcuts in query tool should function properly including shortcuts of result grid.

### DIFF
--- a/web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
@@ -724,7 +724,8 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
     modal: modal,
     params: qtState.params,
     preferences: qtState.preferences,
-  }), [qtState.params, qtState.preferences]);
+    mainContainerRef: containerRef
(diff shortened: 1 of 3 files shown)


web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -724,6 +724,7 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
     modal: modal,
     params: qtState.params,
     preferences: qtState.preferences,
+    mainContainerRef: containerRef
   }), [qtState.params, qtState.preferences]);
```
adityatoshniwal, line 728: add containerRef in memo deps.

adityatoshniwal (changes requested):

Issue #6354, opened 2023-06-02
 Queries with temporary tables in the same transaction


woliveir:
Since version 7.1 of pgadmin4 an error occurs in queries with temporary tables in the same transaction, example for reproduction:

**To Reproduce**

1. Go to 'Tools'
2. Click on 'Query Tool'
3. Execute

```sql 
-- pgadmin4:  version 7.1 and 7.2
CREATE TEMPORARY TABLE test(today) ON COMMIT DROP AS
  VALUES(CURRENT_DATE)
;

SELECT
  *
FROM
  test
;

-- output error message
-- > The specified object could not be found.
```

**Expected behavior**

show test table rows in the Data Output.

**Screenshots**
![image](https://github.com/pgadmin-org/pgadmin4/assets/16959203/caa5c255-35f3-424c-af93-c447ed3780ab)

**Desktop:**
 - OS: Linux Ubuntu/Bionic (18.04 LTS)
 - Version: pgadmin 7.2
 - Mode: Server
 - Browser: chrome, firefox
 - Package type: Docker Container

adityatoshniwal:
Fixed with commit https://github.com/pgadmin-org/pgadmin4/commit/4746fbd34674574bec09113d0240618301fdcebf

pravesh-sharma:
Issue fixed. Verified on latest snapshot build https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-06-21/

Env - macOS Ventura (13.4) Desktop mode.

Issue #6138, opened 2023-04-15
Crash pgAdmin4 7.0 after re-migration tables


DZ8540:
**Describe the bug**
After dumping database and re-migration all tables, when you going to refresh view in table, app going to crash. 

**To Reproduce**

1. Choose someone table, click right mouse button and choose `view/edit data`, then choose `all rows`
2. Wipe or dump all tables (for example via command line), then create the same name table
3. Click on `Execute/Refresh (F5)` button in `PgAdmin4`
4. See crashing application

**Expected behavior**

Application don't going to crash.

**Error message**

Without any messages and dialogs.

**Desktop (please complete the following information):**
 - OS: OSX Ventura 13.0.1
 - Version: 7.0
 - Mode: Desktop
 - Package type: Application

pravesh-sharma:
Issue fixed. verified on snapshot build https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-04-27/

Env - macOS Ventura 13.3.1 (Desktop mode)

adityatoshniwal:
Fix with commit https://github.com/pgadmin-org/pgadmin4/commit/4746fbd34674574bec09113d0240618301fdcebf

akshay-joshi:
Reopen for Testing

anilsahoo20:
Issue fixed. verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-06-23/

Env - macOS Ventura 13.3 (Desktop mode)

PR #6438, opened 2023-06-13
Throw an appropriate error when a table for which View/Edit data is open is deleted and query is executed

nikhil-mohite:


web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -761,6 +762,22 @@ def start_view_data(trans_id):
 
     if status and conn is not None and \
             trans_obj is not None and session_obj is not None:
+
+        # Fetch the Namespace Name and object Name
+        sql_path = 'sqleditor/sql/#{0}#'.format(manager.version)
```
adityatoshniwal, line 767: Wouldn't the query execute give an error? Like table does not exist?

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/__init__.py
nikhil-mohite: We fixed this issue earlier but the fix was not right #6138.

khushboovashi:
This PR will also fix #6354 too.

web/pgadmin/tools/sqleditor/__init__.py
adityatoshniwal: I mean, if the object does not exist in the database then query tool itself will give error in the messages tab?

Issue #6420, opened 2023-06-09
Raise Notice from func/proc or code blocks are no longer displayed live


mickael-choisnet:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

When executing a code block, calling a procedure or a function that raise notices, these notices are no longer displayed in real time during execution under the "message" tab. We are blind until the end of the execution, which prints all the data at once.

**To Reproduce**

- Open a new Query tool on any existing database
- Put the focus on the "message" tab
- run the next code block
   ```sql
   DO
   $BODY$
       BEGIN
           FOR counter IN 1..2
           LOOP
               RAISE NOTICE '% - "%" loop...', timeofday(), counter;  
               PERFORM pg_sleep(5);
           END LOOP;
       END;
   $BODY$;
   ```
- Nothing will be print during execution
- All traces are displayed at the end (~10sec duration)

**Expected behavior**

When executing a code block, calling a procedure or a function that raise notices, these notices should be displayed in real time in the "message" tab as it was in pgAdmin 6.X.

**Screenshots**

- In **pgAdmin 7.3** the messages are all displayed at once at the end of the execution and not as it is executed:
![image](https://github.com/pgadmin-org/pgadmin4/assets/13468399/bad585bc-7a10-41e1-bf84-1ab8c6048607)

- With **pgAdmin 6.21** it works fine:
![image](https://github.com/pgadmin-org/pgadmin4/assets/13468399/87fe7e67-fe79-40bb-b474-47ff5c7daf89)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: 7.3
 - Mode: Desktop
 - Package type: Windows

pravesh-sharma:
Issue fixed. Verified on latest snapshot build https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-06-21/

Env - macOS Ventura (13.4) Desktop mode.

alund-catt:
This is still an issue for me in version 7.5.

pravesh-sharma:
Reopening as issue is reproducible in version 7.5

anilsahoo20:
Issue fixed. Verified on latest snapshot build https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-08-16/

Env - macOS Ventura (13.4) Desktop mode.

PR #6439, opened 2023-06-13
Fix raise notice from func/proc or code blocks are no longer displaye…

khushboovashi:
…d live #6420.


akshay-joshi (changes requested):
API test cases are failing can you please check?

PR #6471, opened 2023-06-19
Fixed an issue where View/Edit data is throwing an error if the user deleted and re-create the same table for which View/Edit data is opened

nikhil-mohite:


adityatoshniwal (changes requested):
minor changes suggested. otherwise, looks good.

Issue #6374, opened 2023-06-07
Add all supported Index storage parameter used with "WITH" clause while creating an Index.


akshay-joshi:

anilsahoo20:
Tested and verified on Snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-07-18/
Environment: macOs Ventura 13.4.1

PR #6534, opened 2023-07-06
Added all supported index storage parameters while creating an index. #6374

pravesh-sharma:


web/pgadmin/browser/server_groups/servers/databases/schemas/tables/indexes/static/js/index.ui.js
```diff
@@ -353,10 +356,94 @@ export default class IndexSchema extends BaseUISchema {
...
+        type: 'int', deps: ['amname'], disabled: state => state.amname !== 'gin' || inSchema(indexSchemaObj.node_info),
+        mode: ['create', 'edit', 'properties'],
+        group: gettext('Definition'),
+        depChange: state => {
+          if (state.amname !== 'gin') {
+            return {gin_pending_list_limit: ''};
+          }
+        }
+      },{
+        id: 'pages_per_range', label: gettext('Pages per range'), cell: 'string',
```
akshay-joshi, line 381: Help String "help_str=gettext("Number of table blocks that make up one block range for each entry of a BRIN index").

akshay-joshi (changes requested):
1) Put all the Storage Parameters in a nested fieldset with the label "With".
2) Check deduplicate_items parameter should not be included in the SQL for versions less than 13.
3) For fillfactor and pages_per_range we are resetting the parameter in the ALTER command, we should do the same for all the other storage parameters (deduplicate_items, fastupdate, autosummarize) when the value is set to False.
4) API tests are failing on v12 and v13

pravesh-sharma:
> * Put all the Storage Parameters in a nested fieldset with the label "With".
> * Check deduplicate_items parameter should not be included in the SQL for versions less than 13.
> * For fillfactor and pages_per_range we are resetting the parameter in the ALTER command, we should do the same for all the other storage parameters (deduplicate_items, fastupdate, autosummarize) when the value is set to False.
> * API tests are failing on v12 and v13
 

- I have made all the changes. However, we can not reset the values of deduplicate_items, fastupdate, autosummarize on selecting false as user may want to set the value as false.
- I ran API test cases on my machine and it is not failing.

Issue #6208, opened 2023-04-28
QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00


khteh:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**
I hit this error from pgadmin4 web connecting to an AWS RDS cluster of 2 nodes. What's the root cause? Is it due to some configurations that I need to adjust? https://docs.sqlalchemy.org/en/20/errors.html#error-3o7r

The RDS dashboard shows only 2 Connections. How does it hit the connection limit and the "overflow"?

```
[pgadmin-0] 2023-04-28 09:03:33,207: ERROR	pgadmin:	QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)
[pgadmin-0] Traceback (most recent call last):
[pgadmin-0]   File "/venv/lib/python3.10/site-packages/flask/app.py", line 1821, in full_dispatch_request
[pgadmin-0]     rv = self.preprocess_request()
[pgadmin-0]   File "/venv/lib/python3.10/site-packages/flask/app.py", line 2312, in preprocess_request
[pgadmin-0]     rv = self.ensure_sync(before_func)()
[pgadmin-0]   File "/venv/lib/python3.10/site-packages/flask_principal.py", line 477, in _on_before_request
[pgadmin-0]     identity = loader()
[pgadmin-0]   File "/venv/lib/python3.10/site-packages/flask_security/core.py", line 708, in _identity_loader
[pgadmin-0]     if not isinstance(current_user._get_current_object(), AnonymousUserMixin):
[pgadmin-0]   File "/venv/lib/python3.10/site-packages/werkzeug/local.py", line 520, in _get_current_object
[pgadmin-0]     return get_name(local())  # type: ignore
[pgadmin-0]   File "/venv/lib/python3.10/site-packages/flask_login/utils.py", line 25, in <lambda>
[pgadmin-0]     current_user = LocalProxy(lambda: _get_user())
[pgadmin-0]   File "/venv/lib/python3.10/site-packages/flask_login/utils.py", line 372, in _get_user
[pgadmin-0]     current_app.login_manager._load_user()
[pgadmin-0]   File "/venv/lib/python3.10/site-packages/flask_login/login_manager.py", line 364, in _load_user
[pgadmin-0]     user = self._user_callback(user_id)
[...]

PR #6681, opened 2023-08-14
Make SQLAlchemy engine options POOL_SIZE & MAX_OVERFLOW configurable #6208

yogeshmahajan-1903:


web/config.py
```diff
@@ -22,6 +22,14 @@
 if sys.path[0] != root:
     sys.path.insert(0, root)
 
+# The number of connections to keep open inside the connection pool.
+# A SQLITE_CONNECTION_POOL_SIZE setting of 0 indicates no limit.
+SQLITE_CONNECTION_POOL_SIZE = 7
+# The number of connections to allow in connection pool "overflow".
+# These connections that can be opened above and beyond the pool_size setting
+SQLITE_CONNECTION_MAX_OVERFLOW = 1000
```
adityatoshniwal, line 30: The default connection size of a PG server is 100. This should be at least lesser than that.
This setting will be used by external PostgreSQL dbs also. So prefixing `SQLITE_` is not appropriate. I would suggest using - `CONFIG_DATABASE_` of `CONFIG_DATABASE_URI`.

adityatoshniwal (changes requested):

web/config.py
```diff
@@ -22,6 +22,14 @@
 if sys.path[0] != root:
     sys.path.insert(0, root)
 
+# The number of connections to keep open inside the connection pool.
+# A SQLITE_CONNECTION_POOL_SIZE setting of 0 indicates no limit.
```
adityatoshniwal, line 26: forgot to change the comment 😄 
\# The config database connection pool size
\# Setting this to 0 will remove any limit.

web/config.py
```diff
@@ -22,6 +22,14 @@
...
+CONFIG_DATABASE_CONNECTION_POOL_SIZE = 5
+# The number of connections to allow in connection pool "overflow".
```
adityatoshniwal, line 28: \# The number of connections allowed to overflow beyond the connection pool size.

adityatoshniwal (changes requested):

web/config.py
yogeshmahajan-1903: Done.

Issue #6487, opened 2023-06-23
Database connection should be restored in query tool if user drop and re-create database with same name


nikhil-mohite:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Connection to the database with the database should be restored.

**To Reproduce**

1. Create database 'test_db1'
2. open query tool for the newly created database
3. now execute some queries
4. Now delete the 'table_db1' (Delete form database collection node using  Delete(Force)) option
5. Re-create the database with the same name 'test_db1'
6. Now try to execute the query in the already opened query tool 
7. It shows an error in Messaged Tab: 
<img width="992" alt="image" src="https://github.com/pgadmin-org/pgadmin4/assets/69510038/dc0e955f-1e75-48fc-b4c0-6ecb226c9ad9">

**Expected behavior**

Query should be fet executed without any error

**Desktop (please complete the following information):**
 - OS: macOS (Ventura 13.3)
 - Version: V-7.3
 - Mode: Desktop

pravesh-sharma:
Issue fixed.

PR #6726, opened 2023-08-29
Fixed  restoration of query tool database connection after dropping and re-creating the database with same name. #6487

anilsahoo20:


nikhil-mohite (changes requested):
1. The Query tool is not working if the database name contains special characters (', "). Also, check autocomplete with this.
2. Please check `Open in new tab` for the Query tool. (Check database with special characters like ',", /,\), it is throwing error `NOT FOUND`
3. SchemaDiff is also failing if we drop and re-create the database with the same name.

nikhil-mohite (changes requested):
1. The Query tool is not working if the database name contains special characters (', "). Also, check autocomplete with this.
2. Please check Open in new tab for the Query tool. (Check database with special characters like ',", /,), it is throwing error NOT FOUND
3. SchemaDiff is also failing if we drop and re-create the database with the same name.

nikhil-mohite (changes requested):

web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -432,7 +433,7 @@ def _init_sqleditor(trans_id, connect, sgid, sid, did, **kwargs):
         return True, internal_server_error(errormsg=str(e)), '', ''
 
     try:
-        conn = manager.connection(did=did, conn_id=conn_id,
+        conn = manager.connection(did=did, database=dbname, conn_id=conn_id,
```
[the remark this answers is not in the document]
anilsahoo20: Made the changes

web/pgadmin/tools/schema_diff/__init__.py
```diff
@@ -750,17 +750,12 @@ def get_schemas(sid, did):
     This function will return the list of schemas for the specified
     server id and database id.
     """
```
khushboovashi, line 752: Any specific reason to remove exception handeling?

khushboovashi (changes requested):
Please fix the below scenario:

1. Start a query tool and run select pg_sleep(50);
2. Now Delete the database and Recreate it.
3. Click continue on the warning message, it will start the execution.
4. Cancel the query through tool bar. 

It will throw No database Found error.

web/pgadmin/tools/schema_diff/__init__.py
anilsahoo20: There was one scenario given by Nikhil, which was basically giving issue in Browser tree schema collection object. The collection object was giving error on refresh of the node and Database connection gets disconnected, But Now if I test the same scenario, the issue is not reproducible and Also Nikhil confirmed that the issue is not reproducible after trying multiple times, so I am reverting to the earlier code.

Issue #6400, opened 2023-06-07
Add "USING method" while creating a Materialized View.


akshay-joshi:

FaharAbbasRizvi:
If a user modifies the access method nothing happens, the behaviour should be consistent like the table  
<img width="1318" alt="Screen Shot 2023-09-12 at 5 22 25 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/cef47dff-2fbe-4ee7-9394-2608809b84fe">
<img width="1320" alt="Screen Shot 2023-09-12 at 5 22 45 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/5749d969-f5ff-45f6-9094-f60c54fb130f">

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-09-18/
Environment: MacOs ventura 13.4.1

Issue #6378, opened 2023-06-07
Add "USING <method>" option while creating a table. 


akshay-joshi:

FaharAbbasRizvi:
@pravesh-sharma 

If the user tries to create a new table and selects the default method heap then syntax error is displayed:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-09-12/
<img width="1302" alt="Screen Shot 2023-09-12 at 5 08 45 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/9b4c3ab5-e3fc-4ad8-ba0f-54db63ef808b">
<img width="1311" alt="Screen Shot 2023-09-12 at 5 08 57 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/d93925fc-f3cb-409b-8445-8ef845d17dba">

anilsahoo20:
Access method once given for partitioned child table while creating table is not coming in properties if we go through table. else it is visible if we go through Partitions node object.(For all PG versions)

MSQL is not getting generated for edit table -> partition child table to any other access method.(For PGv15 and above as alter access method is available in PGv15 and above)
<img width="1728" alt="Screenshot 2023-09-18 at 4 22 02 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/122535205/474b9d37-989b-40ff-9307-710cfaba623f">
<img width="1728" alt="Screenshot 2023-09-18 at 4 23 15 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/122535205/2cfca8b7-5ac7-43e4-b564-09e6ffccd52e">
<img width="1728" alt="Screenshot 2023-09-18 at 4 23 52 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/122535205/104c63c7-207d-4330-8fc7-1d27731db414">
<img width="1728" alt="Screenshot 2023-09-18 at 4 24 03 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/122535205/09dac0ad-7fa1-457f-8fd0-5b85b63d34ef">

pravesh-sharma:
@anilsahoo20 

I have fixed #1 and for the time being we have disabled access method drop down from partitions tab when opening table properties as we do not allow any change from there.

anilsahoo20:
Tested and verified on candidate build: https://developer.pgadmin.org/builds/2023-09-18-1/
[...]

PR #6735, opened 2023-09-04
Added USING method while creating table. #6378

pravesh-sharma:
Added USING method while creating materialized view. #6400


akshay-joshi (changes requested):
1) Using statement should be in the next line.
Screenshot 2023-09-05 at 1 41 10 PM

2) The user should be able to provide an access method when creating partitions of a partitioned table.
Screenshot 2023-09-05 at 1 45 41 PM

PR #6747, opened 2023-09-11
Fixed issue where login form doesn't appear if internal auth source is removed. #6378

pravesh-sharma:
The condition added is same as we had in older login form before removing jQuery and bootstrap.


web/pgadmin/static/js/SecurityPages/LoginPage.jsx
```diff
@@ -8,7 +8,8 @@ import PropTypes from 'prop-types';
 
 export default function LoginPage({userLanguage, langOptions, forgotPassUrl, csrfToken, loginUrl, authSources, authSourcesEnum, oauth2Config, loginBanner, ...props}) {
   const [form, setForm] = useState(({email: '', password: '', language: userLanguage}));
-  const showLoginForm = authSources?.includes('internal');
+  const showLoginForm = !((authSources?.includes('oauth2') || authSources?.includes('kerberos')) && authSources?.length == 1 || (authSources?.includes('oauth2')
```
adityatoshniwal, line 11: please add some comment on what exactly is being checked here.

adityatoshniwal (changes requested):

Issue #6674, opened 2023-08-11
If user change any data type of any column in foreign table then column name "none" is displayed in SQL


FaharAbbasRizvi:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

**To Reproduce**

1. create the following SQL:
CREATE FOREIGN DATA WRAPPER fdw1;
CREATE SERVER fr
    FOREIGN DATA WRAPPER fdw1;
CREATE USER MAPPING FOR postgres SERVER fr;
CREATE FOREIGN TABLE IF NOT EXISTS public.tab1(
    abcd "char" NULL
)
    SERVER fr;

2. Now go into properties on tab1
3.  select column
4. change the datatype to bit
5. Go to SQL
6. following SQL is displayed:

ALTER FOREIGN TABLE IF EXISTS public.tab1
    RENAME COLUMN abcd TO None;

ALTER FOREIGN TABLE IF EXISTS public.tab1
    ALTER COLUMN None DROP NOT NULL;

ALTER FOREIGN TABLE IF EXISTS public.tab1
    ALTER COLUMN None TYPE bit;

ALTER FOREIGN TABLE IF EXISTS public.tab1
    ALTER COLUMN None SET STATISTICS -1;

**Expected behavior**

the data type should be changed successfully in the foreign table

**Screenshots**

<img width="1326" alt="Screen Shot 2023-08-11 at 11 07 06 AM" src="https://github.com/pgadmin-org/pgadmin4/assets/43006122/83379e89-acc5-4a9b-8d2f-86c9143353b4">

**Desktop (please complete the following information):**
 - OS: [ALL]
 - Version: [7.5]
 - Mode: [All]
 - Package type: [ALL]

**Additional context**

akshay-joshi:
Tested on snapshot build 2023-09-28

Issue #6373, opened 2023-06-07
Add 'GENERATED ALWAYS AS..' option while creating column constraints for Foreign Table.


akshay-joshi:

pravesh-sharma:
Tested on the latest source code feature is working as expected.

Env-
OS - macOS Ventura 13.6
Mode - Desktop

Issue #640, opened 2016-09-27
Foreign Table Module new functionality for PG 11 and above (RM #1759)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/1759***
*Originally created by **Karen Blatchley** at 2016-09-27 11:55:51 UTC.*

 New functionality added in PG 11 for this module as below:

- ALTER [ COLUMN ] col_name SET STORAGE { PLAIN | EXTERNAL | EXTENDED | MAIN }
- DISABLE TRIGGER [ trigger_name | ALL | USER ]
- ENABLE TRIGGER [ trigger_name | ALL | USER ]
- ENABLE REPLICA TRIGGER trigger_name
- ENABLE ALWAYS TRIGGER trigger_name
- SET WITH OIDS
- SET WITHOUT OIDS

To add this functionality we need to add Columns, Triggers and Constraints as separate Nodes under the Foreign Table Node.

akshay-joshi:
Found the following issues:
1) Statistics is disabled in edit mode.
2) Only change the datatype of any column from the table dialog. Check MSQL it will create unnecessary syntax for "DROP NOT NULL" and "SET STATISTICS" is shown.
3) Create any foreign table with one column and then open the properties dialog of that table add a new column with the 'bigint' datatype and check the SQL tab, the unnecessary 'NULL' keyword is there. Check the same with the columns properties dialog, syntax is correct.
4) Create any foreign table. Add a new column, set comment, Constraints, Variables, and options, and check the SQL tab. No SQL is visible for the above-mentioned parameters.
5) Unable to change any parameter of the column from the Foreign tables properties dialog. Create any foreign table with one or two columns. Open the properties dialog and go to the columns tab. Try to edit any property of a column like a comment, Constraints, Variables, or Options nothing works.
6) Changing the Type is not working for the normal table as well.
7) Trigger SQL is missing when selecting a foreign table and clicking on the SQL tab.
8) Correct constraint type is not selected for the column when the table's properties dialog is opened.
9) The column's SQL is not correct when selecting any foreign table and clicking on the SQL tab. Create a foreign table with one generated column, check the SQL tab. NULL keyword is added to all the columns.

yogeshmahajan-1903:
> Found the following issues:
> 
> 1. Statistics is disabled in edit mode.
Fixed.

[...]

PR #6751, opened 2023-09-11
Implement column module and its features for foreign table.#640, #6373, #6674.

yogeshmahajan-1903:


khushboovashi (commented):
ALTER FOREIGN TABLE IF EXISTS public.“FT”
    ADD COLUMN col2 bigint GENERATED ALWAYS AS ((1 + 1)) STORED;

Please fix 2 parentheses issue for column SQL.

yogeshmahajan-1903:
> ALTER FOREIGN TABLE IF EXISTS public.“FT” ADD COLUMN col2 bigint GENERATED ALWAYS AS ((1 + 1)) STORED;
> 
> Please fix 2 parentheses issue for column SQL.

This is because expression entered is (1+1) and we wrap expression value in ().

PR #6772, opened 2023-09-19
Mitigate RCE version lookup

gronke:
Fixes a Remote Code Execution vulnerability in Postgres version lookup.

The previous pull-request #6764 is insufficient to mitigate the vulnerability. Command substitution does not require double-quotes for instance:

```python
 >>> subprocess.getoutput(r'echo "{0}"'.format("Hello $(whoami)").replace('"', '""'))
Hello gronke
```

Instead we use Popen and resolve the binary path. Execution occurs without shell context.


adityatoshniwal:
HI @gronke,

I am closing this in favor of https://github.com/pgadmin-org/pgadmin4/pull/6775. I have made code common and other improvements.

Thanks.

PR #6798, opened 2023-09-27
Fix API test cases for backup dialog

nikhil-mohite:


akshay-joshi (changes requested):
PEP8 and test cases are still failing please check.

Issue #5749, opened 2023-01-17
Column marked as primary key once deleted, unable to assign old columns or new created columns as primary key


anilsahoo20:
In any DB table, while defining the table and it's columns, if we mention any column as primary key. Later on if we delete the primary key column, then the primary key assignment is not working with old existing columns and new created columns for the same table.

1. Right click on any database table having primary key
2. click on properties
3. go to columns tab
4. delete the column that is mentioned as primary key
5. try to assign any other column as primary key or create a new column and assign it as primary key
6. Primary key toggle switch is not working
<img width="899" alt="Screenshot 2023-01-17 at 5 04 18 PM" src="https://user-images.githubusercontent.com/122535205/212888798-9f24f872-fc17-4acb-9f73-95b84ede895c.png">

**Expected behavior**

The new primary key must be the mentioned to clicked column or new column that is created after deleting old primary key column.

**Desktop :**
 - OS: OSX monterey
 - Version: 12.5
 - Mode: Desktop
 - Browser: Chrome
 - Package type: Python

anilsahoo20:
Hi @pravesh-sharma 
Found one issue while testing on the latest build.
On delete of any column, the msql is not getting generated. But msql was generated earlier. 
<img width="840" alt="Screenshot 2023-10-16 at 10 45 23 AM" src="https://github.com/pgadmin-org/pgadmin4/assets/122535205/5fe90e35-077c-4891-aec0-433cc1eeb31a">

anilsahoo20:
Tested and verified on candidate build:https://developer.pgadmin.org/builds/2023-10-16-1/
Environment: macOSX Ventura 13.4.1

PR #6805, opened 2023-09-29
Fixed an issue where user was not able to assign new/old columns as primary key once column with primary key is deleted. #5749

pravesh-sharma:


web/pgadmin/browser/server_groups/servers/databases/schemas/tables/static/js/table.ui.js
```diff
@@ -136,6 +136,20 @@ export class ConstraintsSchema extends BaseUISchema {
           }));
           return {columns: state.columns};
         }
+        /* If column or primary key is deleted */
+        if(actionObj.type === SCHEMA_STATE_ACTIONS.DELETE_ROW) {
```
adityatoshniwal, line 140: add condition to check column and pk changed.

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/static/js/table.ui.js
```diff
@@ -136,6 +136,20 @@ export class ConstraintsSchema extends BaseUISchema {
...
+          let deletedColumn = _.differenceBy(actionObj.oldState.columns,state.columns,'name');
```
adityatoshniwal, line 141: A column can be renamed. By this logic, the renamed column will be considered as deleted.

adityatoshniwal (changes requested):

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/static/js/table.ui.js
pravesh-sharma: @adityatoshniwal  In the case of column rename the action is `set_value` so this condition will not get called. I have tested by renaming a column and this condition was skipped.

adityatoshniwal: we have a key called `cid` which is used internally to identify unique rows. please use `cid` instead of `name`

adityatoshniwal (changes requested):
image
I deleted column of pk, all cols enabled. but when i delete the constraint then cols are not enabled.

PR #6818, opened 2023-10-04
Bugfix: pgAdmin imports servers to the wrong accounts in the server version with webserver auth enabled

### LINKED ISSUE #6815: pgAdmin imports servers to the wrong accounts in the server version with webserver auth enabled

**Describe the bug**

After importing the .json file with the server, sometimes (not always) it is imported to another user's account, and the list of servers in my account is still empty. The problem never appears when creating a server manually.

**To Reproduce**

The main thing that I think may be related to the problem is that I am using the following configuration:
AUTHENTICATION_SOURCES = ['webserver']

Users are authorized using basic auth, and the accounts are retrieved by a PHP script. **Such accounts in the pgAdmin sqlite database do not have an email, only a username.**

**Expected behavior**

Servers are always imported under the account of the user who imports them.

**Error message**

No error message - after import, a message appears stating that the operation was successful.

**Screenshots**

Select from pgAdmin sqlite "server" table:

Captura desde 2023-10-04 11-12-10

The last server was imported on user 12 and as you can see it was added on user 2. User 2 confirmed to me that he sees the server imported on my account in his account.

 - OS: Rocky Linux 9
 - Version: 7.6 (docker latest)
 - Mode: Server
 - Browser (if running in server mode): Firefox
 - Package type: Docker

**Additional context**

amarel123:
Closes #6815


web/pgadmin/utils/__init__.py
```diff
@@ -805,9 +805,16 @@ def _does_user_exist(user, from_setup):
     This function will check user is exist or not. If exist then return
     """
     if isinstance(user, User):
-        user = user.email
-
-    new_user = User.query.filter_by(email=user).first()
+        if user.email:
```
khushboovashi, line 808: pgAdmin supports multiple authentication at a time, so it is possible to have same email ids and usernames for multiple users. 
So you have to check for username + Authentication source combination to get the valid user.

khushboovashi (changes requested):

web/pgadmin/utils/__init__.py
amarel123: @khushboovashi I have made changes :)

amarel123: I just didn't know what to do with email because the function accepts text or a user object. So for an object I check auth_source+username, but for text I check email.

khushboovashi: I am proposing to introduce auth_source as a user input while importing the server itself. So, you need change in setup.py itself. The fix you did does not resolve the issue.

amarel123: Sorry, I don't know pgadmin code at all. I tried to solve the problem, but I don't know if I can help more. Can you explain in more detail what we want to achieve? What should be changed in the file https://github.com/pgadmin-org/pgadmin4/blob/master/web/setup.py? What do you mean by user input?

PR #6855, opened 2023-10-12
Add default ServerGroup to user loading 0 Servers

### LINKED ISSUE #6602: Expose Clear Servers as a setup.py argument.
I would like the ability to be able to call clear servers to unregister all servers when they are no longer available.

I am writing a sidecar container for pgadmin that will wakeup on a configured interval and connect to the kubernetes cluster that pgadmin is running in.  It identifies all postgres clusters by label and builds a servers.json file.  It then invokes setup.py.  As postgres servers are removed it "unregisters" them by way of rebuilding the json file with the remaining servers and reapplying it.  However, when no postgres servers are remaining there is no good way to remove the last registration.  

**Describe alternatives you've considered**

I tried sending an empty server list:
{
    "Servers": {}
}

It doesn't error out.  However, the Server Group section of the UI and the ability to manually add servers breaks.  The web page just locks up on refresh or new page load after clicking on add new server.

**Additional context**
It is my hope to have the ability to add or remove servers from 0:n and back to 0 again as needed using an external process like a sidecar or kubernetes job.

benjaminjb:
* check for 0 ServerGroups as result of uploading empty Servers list and re-add the default ServerGroup
* remove duplicate Server delete loop

Fixes #6602


khushboovashi (changes requested):
The changes made in load_database_servers function doesn't make sense. 
If any user cleared all the servers through command line and then wants to add server from the UI, then this fix will never be executed. 

You should add this change in clear_database_servers itself.

khushboovashi:
> > The changes made in load_database_servers function doesn't make sense. If any user cleared all the servers through command line and then wants to add server from the UI, then this fix will never be executed.
> > You should add this change in clear_database_servers itself.
> 
> I've only started digging into the code, so happy to adjust, but just wanted to check:
> 
> a. does `clear_database_servers` ever run independently? It looks to me like the only way to call `clear_database_servers` is to call an import function with the replace option:
> 
> -- from the `setup.py` script [here](https://github.com/pgadmin-org/pgadmin4/blob/master/web/setup.py#L212-L216) -- from the web import function [here](https://github.com/pgadmin-org/pgadmin4/blob/master/web/pgadmin/tools/import_export_servers/__init__.py#L195-L200)
> 
> Is there some way to clear the servers from the commandline without importing?

/path/to/python /path/to/setup.py --load-servers input_file.json --replace

Import/Export server functionality has UI and CLI based both the implementation.
Ref: https://www.pgadmin.org/docs/pgadmin4/7.7/import_export_servers.html#importing-servers

> 
> b. I do like the idea of putting a check like this in `clear_database_servers` because if `clear_database_servers` runs, but the import fouls up for some reason, the user will be left in a bad state for manual creation, like you said.
> 
> BUT if we clear the servers from a user, then add a default `Servers` ServerGroup back, then import the user's JSON file, the user will then end up with (1) the servers they imported and (b) an un-asked for default `Servers` ServerGroup. That feels like a slightly odd UX.

Agreed.
> 
> Another option would be to clear user servers & create a default in `clear_database_servers`; then load the servers & potentially remove the default `Servers` in `load_database_servers` -- but that starts to feel a little fiddly.
> 
> Happy to make requested changes, but just want to clarify the expected results.

As per the current implementation, pgAdmin does not allow the default server-group to be deleted from the left tree browser.
So, that should be implemented in the clear_database_servers function and in that way both your concerns should be taken care of.

Please let me know if you still have any doubt.

web/pgadmin/utils/__init__.py
```diff
@@ -781,14 +781,15 @@ def clear_database_servers(load_user=current_user, from_setup=False):
     for server in servers:
         db.session.delete(server)
 
-    # Remove all groups
-    groups = ServerGroup.query.filter_by(user_id=user_id)
+    # Remove all servergroups except for the first
+    # This matches the UI behavior in
+    # web/pgadmin/browser/server_groups/__init__.py#delete
```
cbandy, line 786: When two pieces/places in the code need to match, I sometimes find it helpful to make a similar comment/reference at the other end.

benjaminjb:
Thx for triggering the tests, @khushboovashi !

Looking at the failures briefly this morning, looks like 
1. the `run-python-tests-pg` workflow is failing in `Install Python dependencies on Windows` in 5 of 5 tests with `Failed to build psycopg-c`
2. the `run-feature-tests-pg` workflow for PG 12 failed the `CheckFileManagerFeatureTest`

Failure 1 seems like something beyond the scope of this PR, though I didn't see other recent-ish (2-3 days old) PRs failing that. 
For Failure 2, I browsed a little to see why this test failed, but wasn't sure. It doesn't seem like it's particularly flaky, but now sure.

Issue #6817, opened 2023-10-04
Wrong code for creating subscription if "copy-data" is set to false


chaosben:
**Describe the bug**

The generated code for creating a subscription is wrong, if the radio-button is switched off: The generated code does not contain the parameter "copy_data", which means that the default value "true" is assumed.

1. Go to Explorer -> 'Subscriptions' -> Contextmenu ->'Create Subscription'
2. Goto 'With'-Tab
3. Disable 'Copy data?'

**Expected behavior**

The generated code should include 'copy_data = false'.

**Screenshots**

![image](https://github.com/pgadmin-org/pgadmin4/assets/1180488/68acb8c7-058f-4a15-ba82-2e6895e9f30f)

Generated code:
```sql
CREATE SUBSCRIPTION test_subscription
    CONNECTION 'host=testdb port=5432 user=postgres dbname=postgres connect_timeout=10 sslmode=prefer'
    PUBLICATION "all"
    WITH (connect = true, enabled = true, create_slot = true, synchronous_commit = 'off', binary = false, streaming = 'False', two_phase = false, disable_on_error = false, run_as_owner = false, password_required = true, origin = 'any');
```

---

![image](https://github.com/pgadmin-org/pgadmin4/assets/1180488/29790e87-c2dc-4973-a8ef-107621d827bc)

Generated code:
```sql
CREATE SUBSCRIPTION test_subscription
    CONNECTION 'host=testdb port=5432 user=postgres dbname=postgres connect_timeout=10 sslmode=prefer'
    PUBLICATION "all"
    WITH (connect = true, enabled = true, copy_data = true, create_slot = true, synchronous_commit = 'off', binary = false, streaming = 'False', two_phase = false, disable_on_error = false, run_as_owner = false, password_required = true, origin = 'any');
```

**Desktop (please complete the following information):**
- OS: Windows 11 22H2
- Version: 7.7
- Mode: Desktop

pravesh-sharma:
Issue fixed. Verified on snapshot build https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-11-20/

OS - macOS 14.1.1 (Sonoma)
Mode - Desktop

chaosben:
Thank you very much! 👍

PR #6889, opened 2023-10-25
Fixed CREATE SUBSCRIPTION's generated query for copy_data parameter when it is set to false. #6817

anilsahoo20:


anilsahoo20:
In the CREATE subscription dialog->WITH tab, 

Issue:
Initially create_slot is true, In connection tab-> once we put host name as localhost, create_slot becomes false, as the host name and port are same as server where the subscription we are creating, but once you change the port, still create_slot remains false. 

Expected Result:
create_slot value is by default true, but if the host name and port are same for both publication and subscription, then create_slot will be false as the slot is not needed, because of same server, but if the host and port are different for publication and subscription, the create_slot need to be true.

nikhil-mohite (approved):

Issue #5807, opened 2023-02-01
psql should assume role set in server properties


adityatoshniwal:
**Describe the bug**
psql is not taking the role set in the server properties.
When a role is set in server properties, the role is assumed when a new connection is created. Currently, this is happening for query tool but not for psql.

**To Reproduce**

1. Set the role in the server properties.
2. Open psql
3. Execute sql - SELECT current_user, session_user;

**Expected behavior**
SELECT current_user, session_user should return otherrole, postgres.

**Desktop (please complete the following information):**
 - OS: Mac OS Big Sur
 - Version: 6.19
 - Mode: Desktop

pr2git:
I report the same bug in release 6.18 and it is now fixed in release 6.20.

adityatoshniwal:
Still not working for me in v6.20.

pr2git:
Delete your current server and create it again.
The bug was that they saved the role as username, so if you create a server with 6.18 or 6.19 your are impacted by the bug, in 6.20 the usernale and role are properly saved so you need to create it again.

pr2git:
So do you explain that existing servers were working but newly added were not?
If it was a bug like you describe the existing servers won't work.
By the way I am happy that it is now fixed in 6.20. :-)

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-11-13/
Environment: macOs Ventura 13.5.1
Package: arm64 and pip wheel

PR #6906, opened 2023-10-30
Fixed an issue where psql was not taking the role used to connect in server properties. #5807

pravesh-sharma:


web/pgadmin/tools/psql/__init__.py
```diff
@@ -531,50 +561,22 @@ def get_connection_status(conn):
...
         is_connected = get_connection_status(conn)
 
-        if is_connected:
-
-            if conn.manager and conn.manager.db_info \
-                    and conn.manager.db_info[int(did)] is not None:
-
-                db_name = conn.manager.db_info[int(did)]['datname']
-                return db_name
-            elif sid:
```
nikhil-mohite, line 553: Any specific reason to remove the else?

nikhil-mohite (changes requested):

web/pgadmin/tools/psql/__init__.py
pravesh-sharma: I discussed this with Khusboo and she suggested else part would never be executed so removed.

nikhil-mohite:
Hi @pravesh-sharma,

Issues found in current PR
1. If we have a role that has special characters in it, the set role fails in the PSQL tool.
image
image

web/pgadmin/tools/psql/__init__.py
khushboovashi: The code written was not necessary. We can get the database name from the connection itself; there is no need for a server manager.

nikhil-mohite (approved):

Issue #6769, opened 2023-09-19
All the about info details should not be visible to non-admin users


adityatoshniwal:
**Describe the bug**
Currently, any user of pgAdmin server mode can see the pgAdmin DB path or other paths set by admin. Such details should be visible to admin only.

**Expected behavior**
Details like entire server config, db & log location, OS name should not be visible to non-admin user.

**Screenshots**
<img width="751" alt="image" src="https://github.com/pgadmin-org/pgadmin4/assets/31855755/1bed9cb1-71fd-4270-bfca-4ce927d33560">

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10, OSX monterey]
 - Version: 7.7
 - Mode: Server
 - Browser (if running in server mode): [e.g. chrome, safari]
 - Package type: [e.g. RPM, DEB, Python, Container, etc.]

yogeshmahajan-1903:
Working fine.

PR #6922, opened 2023-11-02
Fixed an issue where server config information was visible to non-admin users. #6769

pravesh-sharma:


adityatoshniwal (changes requested):
The info is still coming in the API. Hiding from UI is not enough.
Change the backend API to not return the config data, and in the front end simply check if the data is available then show else hide.
Update: Desktop users are always admin, so no check required for server or desktop. Admin should be enough.

Issue #6902, opened 2023-10-29
changing password on pgadmin 4 ver 7.8 fails due to 'number of data characters (9) cannot be 1 more than a multiple of 4'


grahamj157:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**
When trying to change password to a password that has a number of characters that is a multiple of 4 - i get the following error message. - 'invalid base64-encoded string:number of data characters (9) cannot be 1 more than a multiple of 4'
I am trying to enter a password of length 12 characters.

**To Reproduce**
1 start pgadmin app
2 highlight - my server under servers
3 click on -menu> object> change password
4 enter current password
5 enter and confirm new password (12 characters)

3. See the error
'invalid base64-encoded string:number of data characters (9) cannot be 1 more than a multiple of 4'

**Expected behavior**
new password accepted

**Error message**
'invalid base64-encoded string:number of data characters (9) cannot be 1 more than a multiple of 4'

**Screenshots**

**Desktop (please complete the following information):**
2 desktops - both fail
- OS: 
Linux-5.18.0-0.bpo.1-amd64-x86_64-with-glibc2.31
 - Version: pgadmin4 ver 7.8
 - Mode: Desktop
 - Browser Chromium 114.0.5735.91
 - Package type: deb

 - OS: Windows 10- 10.0.22621.SP0
 - Version: [e.g. 6.15]
 - Mode: Desktop
 - Browser Chromium 114.0.5735.91
 - Package type: [e.g. RPM, DEB, Python, Container, etc.]

**Additional context**

grahamj157:
[pgadmin4 config.txt](https://github.com/pgadmin-org/pgadmin4/files/13197174/pgadmin4.config.txt)

grahamj157:
the password i was using is abc017D2023= this passes online base64 validation tests (https://onlinebase64tools.com/validate-base64)

grahamj157:
more context: one other thing i noticed at 4 ver 7.8 is that i am no longer required to supply the master password.

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-11-14/
Environment: macOs ventura 13.5.1
package: arm64

PR #6952, opened 2023-11-08
Fixed an issue where change the server password is not working in desktop mode #6902

nikhil-mohite:


adityatoshniwal:
HI @nikhil-mohite,
The password is changed successfully, but it is not stored internally I guess. Query tool fails:
image

Issue #6920, opened 2023-11-01
ERD tool: ERD to SQL generator is missing columns in first table


kshefchek:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

When I create an ERD and use the Generate SQL tool, the first table in the file is missing columns.  For example, I have a table called "table_a" with columns foo, bar, baz, with a primary key foo, I just see

```
CREATE TABLE IF NOT EXISTS table_a
(

    PRIMARY KEY (foo)
);
```

The rest of the SQL looks correct.

**To Reproduce**

3. See the error

**Expected behavior**

```
CREATE TABLE IF NOT EXISTS table_a
(
    foo integer,
    bar character varying(255),
    baz character varying(255),
    PRIMARY KEY (foo)
);
```
**Error message**

N/A

**Screenshots**

N/A

**Desktop (please complete the following information):**
- Version: 7.8
- Application Mode: Desktop
- NW.js Version: 0.77.0
- Browser: Chromium 114.0.5735.91
- Operating System: macOS-13.6-arm64-arm-64bit

**Additional context**

N/A

adityatoshniwal:
Hi @kshefchek,
I am not able to reproduce this. Can you share a screen recording? Also, share the PostgreSQL version.

kshefchek:
@adityatoshniwal thanks for testing this out on your side! The postgres version is `PostgreSQL 14.6 on aarch64-unknown-linux-gnu, compiled by aarch64-unknown-linux-gnu-gcc (GCC) 7.4.0, 64-bit`, it's a remote aws RDS instance.  I'll create a test database and take a screen recording.

raprocks:
The issue is a little different, the columns are not there for **all** existing tables in the database.  it emits the relationships but since it does not emit the tables i cannot drop the existing ones and recreate them. as it currently stands it generates SQL to create relationships but since they are already present, an error occurs

raprocks:
if there is beta i can help with testing, fell free to tag me. i can test and provide feedback.

akshay-joshi:
@raprocks 

Can you please test on the [snapshot build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-11-17/) and confirm the fix?

raprocks:
will do! give me a few hours

raprocks:
Looks Good, Everything is generated!

georgetakla:
[...]

PR #6961, opened 2023-11-13
Fix ERD tool's SQL generator missing columns for the table. #6920

anilsahoo20:


web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/utils.py
```diff
@@ -339,7 +339,8 @@ def parse_format_columns(data, mode=None):
         final_columns = []
 
         for c in columns:
-            if 'inheritedfrom' not in c:
+            if ('inheritedfrom' not in c) or ('inheritedfrom' in c and
```
adityatoshniwal, line 342: This can be simplified as as `c.get('inheritedfrom', None) is None`

adityatoshniwal (changes requested):

PR #6977, opened 2023-11-16
Force convert OAuth2 claim to the data type

matietz:
Convert the `claim` variable to a list if it is not a list already.


adityatoshniwal:
@matietz  This is in draft state, is this ready for review?

matietz:
Yep, ready for review. Instead of changing the data type of `claims` a more verbose way to get to the desired behaviour could also be the following:

```python
    if (isinstance(claim, list) and any(item in authorized_claims for item in claim)) or claim == authorized_claims:
        reason = "Claim match found. Authorizing"
        return (True, reason)
```

adityatoshniwal:
> Yep, ready for review. Instead of changing the data type of `claims` a more explicit way to get to the desired behaviour could also be the following:
> 
> ```python
>     if (isinstance(claim, list) and any(item in authorized_claims for item in claim)) or claim == authorized_claims:
>         reason = "Claim match found. Authorizing"
>         return (True, reason)
> ```
> 
> Let me know what you think.

I think it will make the code complex. What you did previously is simple and should be fine.

PR #6999, opened 2023-11-24
Ensure focus is not changed to ssh tunnel input box when user explicitly focus in server password edit box

nikhil-mohite:


web/pgadmin/static/js/Dialogs/ConnectServerContent.jsx
```diff
@@ -46,7 +47,9 @@ export default function ConnectServerContent({closeModal, data, onOK, setHeight}
 
   useEffect(()=>{
     setTimeout(()=>{
-      firstEleRef.current && firstEleRef.current.focus();
```
adityatoshniwal, line 49: Can you try useLayoutEffect to focus ref?

adityatoshniwal (changes requested):

Issue #6717, opened 2023-08-24
Automatically created indexes not shown in treeview


asoki:
**Describe the bug**

PostgreSQL automatically creates a unique index when a unique constraint or primary key is defined for a table. These indexes are not shown in pgadmin's treeview.
This is misleading and since the index is not listed, it is not possible to see its properties or its statistics in the gui.

**To Reproduce**

1. execute: `create table public.foo(id integer not null primary key);`
2. check if postgres created the index for the pkey
3. `SELECT
    tablename,
    indexname,
    indexdef
FROM
    pg_indexes
WHERE
    schemaname = 'public' AND tablename='foo'
ORDER BY
    tablename,
    indexname;`

4. in PGAdmin treeview navigate to *server*/Databases/*dbname*/Schemas/public/Tables/foo and open the node
5. Click on Indexes to see foo's indexes

**Expected behavior**

PGAdmin should list all indexes of the selected table

**Actual behavior**

6. the index list is empty

**Screenshots**

![pgadmin-treeview-no-index](https://github.com/pgadmin-org/pgadmin4/assets/1533751/d2fdad8c-9676-4530-8481-cc18339c6dbe)

**Desktop (please complete the following information):**
 - OS: MacOS 13.3.1
 - Version: 7.6
 - Mode: Desktop
 - Package type: dmg

nikhil-mohite:
Hi @asoki ,
pgAdmin shows them in the constraints node.

Chimera78:
This seems to be a very old issue, initially reported in 2004 for pgAdmin III, as I already found this topic mentioned in the [mail-archive](https://www.mail-archive.com/pgadmin-support@postgresql.org/msg02528.html).

It makes no sense to hide automatically generated primary key indexes. Yes, it is created from a constraint, but it is also an index, so both should be shown in the tree view.

Both Oracle Developer and MS SQL Management Studio show constraints as constraints and indexes as indexes.
Indexes are also automatically created for primary keys for Oracle and MS SQL and as a developer or admin, it makes sense that the tool used to view the DB Scheme also shows this information.

The pgAdmin tree on the left however only ever shows manually created indexes, but not the automatically generated ones.
This makes no sense. Of course you can run SQL to find the indexes for specific tables (whether created manually or automatically) but that defeats the purpose.

Please finally add the automatically created indexes in the tree view under the Indexes, for better transparency (WYSIWG) instead of hiding them.

akshay-joshi:
[...]

PR #7022, opened 2023-11-30
Ensure that automatically created indexes should be shown in the treeview. #6717

akshay-joshi:


adityatoshniwal:
Please change the commit message to
`Ensure that indexes created by constraints are visible in the object explorer when "Show system objects" is enabled. #6717`

adityatoshniwal (approved):

Issue #6887, opened 2023-10-24
SQL editor not underlining location of syntax error


shadysudo:
I no longer see the location of syntax errors underlined in red in the SQL editor when the messages panel returns a syntax error.

pgadmin versions: 7.7 and 7.8

I have the same experience using the standard windows install of pgadmin4 on Windows 10 and using the docker container from dpage/pgadmin4.

The PostgeSQL database is version 12.

I am expecting customers to be underlined in red like in the documentation linked below https://www.pgadmin.org/docs/pgadmin4/7.8/query_tool.html#messages-panel

shadysudo:
![image](https://github.com/pgadmin-org/pgadmin4/assets/66808114/c604adc4-7260-4832-a239-516b356344cf)

anilsahoo20:
Tested and verified on candidate build: https://developer.pgadmin.org/builds/2023-12-11-1/
Environment: macOs Ventura 13.5.1
Package: arm64 Installer

PR #7042, opened 2023-12-07
Fixed an issue where syntax error was not highlighting for explain query. #6887

pravesh-sharma:


web/pgadmin/tools/sqleditor/utils/apply_explain_plan_wrapper.py
```diff
@@ -25,3 +25,17 @@ def apply_explain_plan_wrapper_if_needed(manager, sql):
...
+
+def return_explain_query_length(query):
+    """
+    This method returns query length if it is explain query.
+
+    Args:
+        query: sql query
+    """
+    query = query.decode()
+    if query.startswith("EXPLAIN"):
```
adityatoshniwal, line 38: What will happen if someone types a `EXPLAIN` query with an incorrect table and executes the query?

pravesh-sharma: This line only checks if the given query is `EXPLAIN` query or not, if the table is incorrect it will highlight the table name. Attached screenshot
Screenshot 2023-12-07 at 1 01 15 PM

adityatoshniwal: I asked since you've checked if the query starts with `EXPLAIN` which is a possibility without using Explain button.

nikhil-mohite (changes requested):

nikhil-mohite:
@pravesh-sharma, If possible fix the cursor position to where the user clicked before running the query, currently it getting reset to the start (col 0) of the highlighted line. (This is an older issue not introduced due to current changes)

pravesh-sharma:
> @pravesh-sharma, If possible fix the cursor position to where the user clicked before running the query, currently it getting reset to the start (col 0) of the highlighted line. (This is an older issue not introduced due to current changes)

Fixed the issue, now cursor will move to the end of the error.

Issue #7004, opened 2023-11-27
Copying query editor output with JSON produces invalid tsv


StefanBrand:
**Describe the bug**

When I copy output from the query editor that contains a JSON column, the output is unusable because it is quoted incorrectly.

**To Reproduce**

1. Execute `SELECT json_build_object('foo', 'bar', 'baz', 'qux')`
2. Copy output
4. Get `"{""foo"" : ""bar"", ""baz"" : ""qux""}"`

**Expected behavior**

Quotes in output should be escaped or the JSON string should be surrounded by single quotes.

**Screenshots**

![image](https://github.com/pgadmin-org/pgadmin4/assets/8948865/4ed5be67-69bf-45cd-90d7-887ee21ad4fd)

**Desktop (please complete the following information):**
- Version: 6.15
- Application Mode: Server
- Browser: Firefox 120.0
- Operating System: Linux-5.10.179-166.674.amzn2.x86_64-x86_64-with

StefanBrand:
Ah, thank you, this is very helpful!

yogeshmahajan-1903:
Result grid provides CSV escape output, which will be above format.([Ref](https://jsonformatter.org/csv-escape)).
To copy data in json format, double click cell and copy from json editor. Hence rejecting the issue.

PR #7069, opened 2023-12-18
Ensure json output copied from result grid is correctly quoted.#7004

yogeshmahajan-1903:


web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/CopyData.js
```diff
@@ -70,7 +70,7 @@ export default class CopyData {
   stringQuoteCell(value, column) {
     if (value && _.isObject(value)) {
       value = this.CSVOptions.quote_char + JSONBigNumber.stringify(value) + this.CSVOptions.quote_char;
-    } else if (value && column.cell != 'number' && column.cell != 'boolean') {
+    } else if (value && column.cell != 'number' && column.cell != 'boolean' && column.cell != 'Json') {
```
adityatoshniwal, line 73: A JSON is still a string. I am not sure if this will work with CSV (when multiple cells are copied). Instead, we should escape the quotes of JSON data when using double quotes.

adityatoshniwal (changes requested):

Issue #7066, opened 2023-12-18
UI Issue: browser explorer tree state is not saving.


nikhil-mohite:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

The save tree state API call is not working (API call is missing on the set interval in Preferences ) to save the current state of the object explorer tree, which will be used to auto-expand the tree on the refresh/ reopening of the pgAdmin 4 app.

**To Reproduce**

1. Open pgAdmin
2. Connect to any server, expand nodes like tables, and select any table
3. Now check the tree state API call is missing as per set interval in Preferences.
4. This will not auto-expand the tree on the next refresh or reopening of the app.

**Expected behavior**

The tree state should get saved as per the interval set in Preferences.

**Desktop (please complete the following information):**
 - OS: OSX (Sonoma ) will be reproducible on all platforms
 - Version: 8
 - Mode: Both
 - Browser (if running in server mode): [e.g. chrome, safari]
 - Package type: [e.g. RPM, DEB, Python, Container, etc.]

nikhil-mohite:
Resolved, Verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2023-12-26/
OS: macOS (Sonoma)
Mode: Desktop

PR #7079, opened 2023-12-22
Fixed an issue where object explorer tree state was not saving. #7066

pravesh-sharma:


web/pgadmin/static/js/tree/pgadmin_tree_save_state.js
```diff
@@ -57,29 +57,34 @@ _.extend(pgBrowser.browserTreeState, {
...
+
+      if (saveIntervalId)
+        clearInterval(saveIntervalId);
+
+      if (!_.isUndefined(save_tree_state_period) &&  save_tree_state_period !== -1) {
+        // Save the tree state at given save_tree_state_period
+        saveIntervalId = setInterval(this.save_state, (save_tree_state_period) * 1000);
+        this.fetch_state.apply(this);
+
+        pgBrowser.Events.on('pgadmin:browser:tree:expand-from-previous-tree-state',
```
adityatoshniwal, line 73: there is problem here. When the preference will change, a event subscription will be added here and it keep on piling up. It should be done only once, and turned off if the tree state interval is disabled.

adityatoshniwal (changes requested):

web/pgadmin/static/js/tree/pgadmin_tree_save_state.js
pravesh-sharma: We can deregister the event in case the tree interval is disabled. I read on some blogs and found that duplicates get ignored if we register an event multiple times with the same parameters. I don't know if we have a different implementation of events.

pravesh-sharma: Okay, in our case it is being added multiple times, will fix it.

web/pgadmin/static/js/tree/pgadmin_tree_save_state.js
```diff
@@ -57,29 +57,46 @@ _.extend(pgBrowser.browserTreeState, {
...
-
-      // Fetch the tree last state while loading the browser tree
-      this.fetch_state.apply(this);
+    let saveIntervalId, offExpandFromPrevState, offRemoveFromTreeState, offUpdateTreeState;
+    usePreferences.subscribe((prefStore)=>{
+      // Subscribe to listen for preferences change
+      const save_tree_state_period = prefStore.getPreferences('browser', 'browser_tree_state_save_interval')?.value;
+      if (saveIntervalId) {
+        clearInterval(saveIntervalId);
+        offEventListener();
```
adityatoshniwal, line 66: probably make `saveIntervalId = null` here after clearing to avoid errors.

web/pgadmin/static/js/tree/pgadmin_tree_save_state.js
```diff
@@ -57,29 +57,46 @@ _.extend(pgBrowser.browserTreeState, {
...
+      }
+      if (!_.isUndefined(save_tree_state_period) &&  save_tree_state_period !== -1) {
+        // Save the tree state at given save_tree_state_period
+        saveIntervalId = setInterval(this.save_state, (save_tree_state_period) * 1000);
+        this.fetch_state.apply(this);
+        onEventListener();
+      }else if (!_.isUndefined(save_tree_state_period)) {
```
adityatoshniwal, line 73: fix formatting

web/pgadmin/static/js/tree/pgadmin_tree_save_state.js
```diff
@@ -57,29 +57,46 @@ _.extend(pgBrowser.browserTreeState, {
...
-          console.warn(
-            gettext('Error resetting the tree saved state."'), error);
-        });
-    }
-
+    };
+
+    const offEventListener = () => {
+      // Deregister event listener
+      offExpandFromPrevState();
```
adityatoshniwal, line 95: probably check if `offExpandFromPrevState` is defined, then only call. You can change all these to `offExpandFromPrevState?.()`

adityatoshniwal (changes requested):

Issue #7041, opened 2023-12-06
Host change while connected results in unclear state


febner:
**Describe the bug**

When I change the host of the DB I'm currently connected to,
the attached warning appears.
When I retry after following the instructions (disconnect), my previous changes appear to be applied in the UI.
Validation doesn't let me save (as unchanged probably).

If I'd reconnect, it would be unclear as to which host I'm connecting to.
So also some security or at least safety implications in here.

**To Reproduce**

1. Connect to a DB
2. Rightclick the server connected to in the Servers window
3. Click on 'Properties'
4. Change the host
5. Close attached error
6. Disconnect DB
7. redo 2. and 3.

**Expected behavior**

Host is unchanged, as changes weren't possible to apply, as in the 'connected' state.

or

Host can't be modified while connected.

**Error message**

see screenshots

**Screenshots**

![Screenshot from 2023-12-06 16-25-32](https://github.com/pgadmin-org/pgadmin4/assets/1845406/e09738c3-fc22-4628-929c-2ef2d12f57e2)

**Desktop (please complete the following information):**
 - OS: Linux-6.6.2-201.fc39.x86_64-x86_64-with-glibc2.38
 - Version: 8.0
 - Mode: Desktop

pravesh-sharma:
<img width="1230" alt="Screenshot 2023-12-29 at 1 54 23 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/111979571/463da261-ed5f-4822-aa82-724543d0c545">
<img width="1230" alt="Screenshot 2023-12-29 at 1 55 52 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/111979571/ca4a45ea-65bd-4cc2-989c-cda5fd27658a">

For consistency, can we keep the sequence of the connection string the same? If you check screenshot 2 it is from the older version and screenshot 1 is from the snapshot build. In the update host and port comes at the end.

pravesh-sharma:
Issue resolved. Verified on snapshot build https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-01-02/

Env - macOS 14.1.2
Mode - Desktop

PR #7086, opened 2023-12-26
Fix an issue where updating node details from the properties dialog are not getting updated on the properties panel if the properties panel is active #7041

nikhil-mohite:


web/pgadmin/misc/properties/ObjectNodeProperties.jsx
```diff
@@ -196,15 +196,22 @@ export default function ObjectNodeProperties({panelId, node, treeNodeInfo, nodeD
     && formType !== 'tab') {
     schema.filterGroups = [gettext('Security')];
   }
+  // To check node data is updated or not
+  const staleCounter = useRef(0);
 
   const key = useMemo(()=>{
```
adityatoshniwal, line 202: lets add one `useEffect` with `nodeData?._id` as dependency and set `staleCounter.current = 0` in that just to be safe.

adityatoshniwal (changes requested):

Issue #7078, opened 2023-12-21
Unable to cancel or terminate active queries from Dashboard


shadysudo:
**Describe the bug**

When I click on the cancel button next to connection with an actively running query in the sessions section of the dashboard, I get the message "You are not allowed to cancel the main active session".  I get a similar message when clicking the terminate button for a connection with an active query.  If I am logged in as a super user, I get that same message when trying to cancel active queries from run by other users as well as my own.  Non super users get the message when trying to cancel their own active queries.

Clicking the terminate button on connections without active queries will close the connection as expected.

**To Reproduce**

1) Open a new query tool window in pgAdmin4
2) Execute "select pg_sleep(1000);"
3) Click on Dashboard
4) Check "Sessions" tab under "Server activity" section then look for the connection with the active query from the query tool.
5) Click on cancel button and cancel the active session
6) The message "You are not allowed to cancel the main active session" will pop up at the bottom of the screen
7) Click on the terminate button to terminate the connection
8) The message "You are not allowed to terminate the main active session" will pop up at the bottom of the screen

**Expected behavior**

The active query should be canceled and in the query tool you will see "ERROR:  canceling statement due to user request"

**Error message**

"You are not allowed to cancel the main active session"
"You are not allowed to terminate the main active session"

**Screenshots**
![image](https://github.com/pgadmin-org/pgadmin4/assets/66808114/29ba0065-136f-4072-92c4-23933de9550b)

**Desktop (please complete the following information):**
 - OS: Windows 10 Desktop and Linux (dpage/pgadmin4 docker container)
 - Version: pgAdmin version 8.1
 - Mode: Desktop and Server
 - Browser Chrome
 - Package type: windows installer and docker container

**Additional context**

The database that I am connecting to is PostgreSQL version 12.17
I have seen this behavior in pgAdmin4 for several versions.  I believe it used to work correctly in pgAdmin4 version 6.

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-01-02/
Package: pip wheel, arm64
Environment: macOs ventura 13.5.1

PR #7090, opened 2023-12-29
Fixed an issue where user was not able to cancel or terminate active queries from dashboard. #7078

pravesh-sharma:


web/pgadmin/dashboard/static/js/Dashboard.jsx
```diff
@@ -695,6 +695,7 @@ function Dashboard({
         ? treeNodeInfo.server.user.can_signal_backend
         : false;
 
+    let active_sessions = dashData.filter((data)=>data.state === 'active');
```
adityatoshniwal, line 698: shouldn't this filter also include database name comparison with maintenance_database?

adityatoshniwal (changes requested):

web/pgadmin/dashboard/static/js/Dashboard.jsx
pravesh-sharma: I have copied this logic from the old code. If we put a filter for maintenance_database, we will not be able to terminate/cancel the query running on the query tool of the maintenance database. Also, we can not terminate the connection made with the maintenance database (default connection), the backend will throw the error which is handled.

adityatoshniwal: as per the logic, it is trying to check if there are more than one connection for the connection being terminated. But what if there are multiple DB connections?

pravesh-sharma: If you check line 732, the condition is if we have more than one active connection then we should be able to terminate.

adityatoshniwal: What I want to ask is - if there are 2 active connections in the dashboard with both connected to different databases, will it allow terminating the maintenance one?

PR #7099, opened 2024-01-04
Added check to avoid keyring call to get server password if save_password flag is not set

nikhil-mohite:


web/pgadmin/browser/__init__.py
```diff
@@ -724,18 +724,19 @@ def set_master_password():
 
         try:
             all_server = Server.query.all()
+            saved_password_servers = [server for server in all_server if
+                                      server.save_password]
             # pgAdmin will use the OS password manager to store the server
             # password, here migrating the existing saved server password to
             # OS password manager
-            if keyring.get_password(
+            if len(saved_password_servers) > 0 and keyring.get_password(
```
adityatoshniwal, line 732: so even if one server has saved password then it will call get password for all?

adityatoshniwal (changes requested):

web/pgadmin/browser/__init__.py
nikhil-mohite: No for all, it is for specific server only.

adityatoshniwal:
@nikhil-mohite  all the python tests are failing for this.

adityatoshniwal:
@nikhil-mohite  The tests are still failing for ubuntu.

Issue #6085, opened 2023-04-03
Servers => duplicate Server context menu


nigzak:
**Describe the solution you'd like**
There should be a context menu for cloning a server configuration
Server <right click> => clone server

**Describe alternatives you've considered**
No one

**Additional context**
If you working with different usernames ("rights") in the database you have to manually clone the servers to avoid you always working with the highest rights

ex:
admin-account
readonly-account

If I want to configure stuff I need to use admin account
If I want to make a "select ..." I can use the readonly-account - for this I have 1:1 to clone the admin server and user different username/password

If I have more rights it can get to more duplicated servers.
This is as today messy because I have to configure all servers manually multiple times ...

HINTS: 

- would also be nice to clone a "shared server" to a "local server" (like a "copy template")
- certificates should also be copied if configured, this should include COPY if it is not existing in dedicated account ("shared server clone"; ex: root server certificate)

johnmreynolds:
This is definitely one of the biggest annoyances I have with pgadmin. I found this from searching for whether there is any way of doing this.

pravesh-sharma:
Tested and is working as expected. Verified on candidate build https://developer.pgadmin.org/builds/2024-01-08-1/

Env 
OS - Rocky 8
Mode - Desktop and server

PR #7106, opened 2024-01-05
Added copy server support, allowing the duplication of existing servers with the option to make certain modifications. #6085

akshay-joshi:


web/pgadmin/browser/server_groups/servers/__init__.py
```diff
@@ -1763,7 +1763,7 @@ def reload_configuration(self, gid, sid):
 
         if conn.connected():
             # Execute the command for reload configuration for the server
-            status, rid = conn.execute_scalar("SELECT pg_reload_conf();")
+            status, _ = conn.execute_scalar("SELECT pg_reload_conf();")
```
khushboovashi, line 1766: execute_void can be used if SQL is returning nothing.

web/pgadmin/browser/server_groups/servers/static/js/server.js
```diff
@@ -135,6 +154,11 @@ define('pgadmin.node.server', [
           data: {
             data_disabled: gettext('SSH Tunnel password is not saved for selected server.'),
           },
+        }, {
+          name: 'copy_server', node: 'server', module: this,
+          applies: ['object', 'context'], callback: 'show_obj_properties',
+          label: gettext('Copy Server'), data: {action: 'copy'},
```
khushboovashi, line 160: The label should be 'Copy Server...' as it is a dialog.

web/pgadmin/browser/server_groups/servers/__init__.py
akshay-joshi: The above code is very old, I have fixed the SonarQube code smell. Leaving it for now.

Issue #6557, opened 2023-07-11
Use COOKIE_DEFAULT_PATH in session cookie


Joe136:
**Related Issues**
#3026 : A technical explanation of the Issue. The issue was closed without a solution

**Describe the solution you'd like**

The path for session cookies is not set. If the website is hosted in a subdirectory like https://example.org/pgadmin4/ the session cookie will be sent to the root path and other directories / services as well (e.g. /static/js/, /api/, /gitlab/, /nextcloud/, /docker-registry/, /support/).

Add `path=config.COOKIE_DEFAULT_PATH,` in https://github.com/pgadmin-org/pgadmin4/blob/def1f31ba8372cf8c814729d9a0a906e99facfd2/web/pgadmin/utils/session.py#L313-L321

Optionally define a new option `SESSION_COOKIE_PATH` or use `X-Script-Name` from HTTP Header #3026 (reverse proxy).

Setting the path will lower security issues, reduce the size of HTTP requests to other urls and ease debugging in the WebConsole (no PgAdmin cookies in requests which are not related to PgAdmin).

In a hosting setup with subdirectories instead of subdomains, this is an important security setting.

**Describe alternatives you've considered**

Possibly rewrite response cookie in reverse proxy.

**Additional context**

In my setup, the main page is hosted in the root path; other services like API server runs in subdirectories.
PgAdmin is proxied in /pgadmin4/ for development to ease access to database for debugging.
When analyzing the HTTP requests in the WebConsole, the session cookie will occur in all requests, e.g. to static js/css/images, api calls, login-related pages and additional services.
Even this small setup forces me to detect three origins of cookies: NextJS server part, API server and PgAdmin. The number increases with every service, that don't use cookie paths.

Gruummy:
I also already recognized that and it would really be helpfull.

From my point of view... if the pgadmin base path is reconfigured with the header "x-script-name", then this base path should be used to add the the session cookie.

It simply would then work out of the box without the need to adjust this setting explicitly.

Btw. same should be considered for the configuration of the NON session cookie

Both types of cookie should have the dedicated path configured with if the optional request header contains for example "x-script-name: /pgadmin4"

adityatoshniwal:
> I also already recognized that and it would really be helpfull.
> 
> From my point of view... if the pgadmin base path is reconfigured with the header "x-script-name", then this base path should also be used to be added to the session cookie.
> 
> It simply would then work out of the box without the need to adjust this setting explicitly.
[...]

PR #7134, opened 2024-01-18
Use COOKIE_DEFAULT_PATH or SCRIPT_NAME in session cookie path. #6557

adityatoshniwal:


khushboovashi (commented):
I haven't tested yet but SCRIPT_NAME would give output like 'pgadmin' and we need path here, so it should be '/pgadmin4' .

adityatoshniwal:
> I haven't tested yet but SCRIPT_NAME would give output like 'pgadmin' and we need path here, so it should be '/pgadmin4' .

HI @khushboovashi,
I tested with SCRIPT_NAME as `pgadmin` and `/pgadmin`, both works fine for session cookie path.

Issue #7153, opened 2024-01-24
Fatal Error: "The pgAdmin 4 server could not be contacted"


DaniiC00kie:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Used pgadmin v8.0 for about 2 months, now it seems to have problems at starting up. (The Interface won't open, it stopps with an exception)

**To Reproduce**

// not quite sure where this bugs comes from
1. Open pgadmin (version 8.0 and up) 

**Expected behavior**

The User Interface should open (please see screenshots where it's stuck=

**Error message**

pgAdmin Runtime Environment
--------------------------------------------------------
Python Path: "C:\Users\Danii\AppData\Local\Programs\pgAdmin 4\python\python.exe"
Runtime Config File: "C:\Users\Danii\AppData\Roaming\pgadmin\runtime_config.json"
pgAdmin Config File: "C:\Users\Danii\AppData\Local\Programs\pgAdmin 4\web\config.py"
Webapp Path: "C:\Users\Danii\AppData\Local\Programs\pgAdmin 4\web\pgAdmin4.py"
pgAdmin Command: "C:\Users\Danii\AppData\Local\Programs\pgAdmin 4\python\python.exe -s C:\Users\Danii\AppData\Local\Programs\pgAdmin 4\web\pgAdmin4.py"
Environment: 
  - ALLUSERSPROFILE: C:\ProgramData
  - APPDATA: C:\Users\Danii\AppData\Roaming
  - CHROME_CRASHPAD_PIPE_NAME: \\.\pipe\crashpad_10044_UOQMGEZCHGPJVKTJ
  - CHROME_RESTART: NW.js|NW.js ist abgestürzt. Jetzt neu starten?|LEFT_TO_RIGHT
  - CommonProgramFiles: C:\Program Files\Common Files
  - CommonProgramFiles(x86): C:\Program Files (x86)\Common Files
  - CommonProgramW6432: C:\Program Files\Common Files
  - COMPUTERNAME: DESKTOP-M7CJU4S
  - ComSpec: C:\Windows\system32\cmd.exe
  - DriverData: C:\Windows\System32\Drivers\DriverData
  - EFC_2784: 0
  - FPS_BROWSER_APP_PROFILE_STRING: Internet Explorer
  - FPS_BROWSER_USER_PROFILE_STRING: Default
  - HOMEDRIVE: C:
  - HOMEPATH: \Users\Danii
[...]

Issue #7125, opened 2024-01-11
[desktop] Doesn't launch in windows 11


slushpuppy:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Hi, I am unable to start pgadmin unless i delete C:\Users\<username>\AppData\Roaming\pgadmin folder

**To Reproduce**

1. install pgadmin
2. start pgadmin/fill in server details
3. reboot the comp
4. pgadmin fails to start. no loading splash screen or anything. nwjs is running in task manager tho

**Expected behavior**

**Error message**

If possible, include the full, unedited text of any error messages displayed.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: windows 11
 - Version: [e.g. 6.15]
 - Mode: desktop
 - Browser (if running in server mode): [e.g. chrome, safari]
 - Package type: Python 3.12.1

**Additional context**

adityatoshniwal:
Hi @slushpuppy,
Can you please share the screenshot of error and logs please?

slushpuppy:
Hi @adityatoshniwal i checked C:\Users<username>\AppData\Roaming\pgadmin folder but logs are not being logged there, is there a way I can run it in verbose mode in cli?

nikhil-mohite:
Hi @slushpuppy ,

Can you please try killing the process from the task manager and please share the screen recording if possible?

MrBns:
Hi, this issue is closed. But still  I am facing this problem..

Issue #6296, opened 2023-05-18
Can't connect to pgadmin4 -- internal server error


Rufooooo:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

On updating to version 7.1, I can no longer connect to the server to login.  The apache error message says 
                      File "/usr/pgadmin4/venv/lib64/python3.11/site-packages/psycopg/pq/__init__.py", line 106, in mport_from_libpq
                      [wsgi:error]` [pid 37382:tid 37537] [remote 192.168.1.75:35318]     raise ImportError(
                      [wsgi:error]` [pid 37382:tid 37537] [remote 192.168.1.75:35318] ImportError: no pq wrapper available.

   
**To Reproduce**
1. Install (or reinstall) pgadmin4 7.1
(I am using postgresql 12 on linux (fedora 37), pgadmin4 rpms installed from pgdg and pgadmin4 repos.)
2. type into browser address of pgadmin4 server
3. See the error.

**Expected behavior**
See the login screen for pgadmin4

**Error message**
[Thu May 18 09:40:13.792982 2023] [wsgi:error] [pid 37382:tid 37560] [remote 192.168.1.75:45278] mod_wsgi (pid=37382): Failed to exec Python script file '/usr/pgadmin4/web/pgAdmin4.wsgi'.
[Thu May 18 09:40:13.793030 2023] [wsgi:error] [pid 37382:tid 37560] [remote 192.168.1.75:45278] mod_wsgi (pid=37382): Exception occurred processing WSGI script '/usr/pgadmin4/web/pgAdmin4.wsgi'.
[Thu May 18 09:40:13.793769 2023] [wsgi:error] [pid 37382:tid 37560] [remote 192.168.1.75:45278] Traceback (most recent call last):
[Thu May 18 09:40:13.794926 2023] [wsgi:error] [pid 37382:tid 37560] [remote 192.168.1.75:45278]   File "/usr/pgadmin4/web/pgAdmin4.wsgi", line 38, in <module>
[Thu May 18 09:40:13.794960 2023] [wsgi:error] [pid 37382:tid 37560] [remote 192.168.1.75:45278]     from pgAdmin4 import app as application
[Thu May 18 09:40:13.794970 2023] [wsgi:error] [pid 37382:tid 37560] [remote 192.168.1.75:45278]   File "/usr/pgadmin4/web/pgAdmin4.py", line 102, in <module>
[...]

PR #7168, opened 2024-01-30
Fix PgAdmin not launching on Win32

kristof-mattei:
Fixes https://github.com/pgadmin-org/pgadmin4/issues/7153
Fixes https://github.com/pgadmin-org/pgadmin4/issues/7125

Relates to https://github.com/pgadmin-org/pgadmin4/issues/6296

This will not fix another issue in Linux: https://github.com/pgadmin-org/pgadmin4/issues/7153#issuecomment-1911616554. For that we'll need to add a Linux case.

How does this fix the bug?

Environment variables are not consistent on Windows and are case insensitive.

On a clean machine, the user env `PATH` is spelled `Path`. The system env var is spelled `path`.

I've removed the second check which used the outdated `==` (idea for a lint?).

Then I uppercase the `key` to ensure that we catch `Path`, `path`, `PATH`, `pAtH` and all other variations.

Some output in PowerShell to prove they're case insensitive: 

```
❯ $env:VAR_NAME = "UPPER_CASE"

~
❯ $env:var_name = "lower_case"

~
❯ echo $ENV:VAR_NAME
lower_case

~
❯ echo $ENV:VaR_nAmE
lower_case
```

### DIFF
--- a/runtime/src/js/pgadmin.js
+++ b/runtime/src/js/pgadmin.js
@@ -80,7 +80,7 @@ function startDesktopMode() {
       process.env[key] = updated_path;
     }
 
-    if (platform() === 'win32' && (key === 'PATH' || key == 'Path')) {
+    if (platform() === 'win32' && key.toUpperCase() === 'PATH') {
       let _libpq_path = path.join(path.dirname(path.dirname(path.resolve(pgadminFile))), 'runtime');
       process.env[key] = _libpq_path + ';' + process.env[key];
     }


adityatoshniwal:
Thank you for fixing this @kristof-mattei

kristof-mattei:
> Thank you for fixing this @kristof-mattei

Pleasure is all mine. It's the least I can to do show my gratitude for this awesome tool.

Issue #6785, opened 2023-09-22
Formatting inserts empty lines in specific case


bveldkamp:
**Describe the bug**
I have a specific case where formatting the sql in a query window inserts an empty line inside a function between function arguments.  The 2nd argument is also moved too much to the right. 
I have never changed any formatting preferences, those are all default

**To Reproduce**
1. Paste this sql:
```
SELECT W.OBJECTID,
	G.SHAPE
FROM POINTS W
LEFT JOIN POLYGONS G ON ST_CONTAINS(G.SHAPE, W.SHAPE)
```
2. Format the SQL (Ctrl+Shift+K)
3. The result looks like this (scrolll to the right):
```
SELECT W.OBJECTID,
	G.SHAPE
FROM POINTS W
LEFT JOIN POLYGONS G ON ST_CONTAINS(G.SHAPE,

																									W.SHAPE)
```
 
**Expected behavior**
Code should be formatted as:
```
SELECT W.OBJECTID,
	G.SHAPE
FROM POINTS W
LEFT JOIN POLYGONS G ON ST_CONTAINS(G.SHAPE, W.SHAPE)
```
or maybe line up the arguments on separate lines:
```
SELECT W.OBJECTID,
	G.SHAPE
FROM POINTS W
LEFT JOIN POLYGONS G ON ST_CONTAINS(G.SHAPE, 
                                    W.SHAPE)
```

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version: 7.6
 - Mode: Desktop

anilsahoo20:
Related to issue #5083.

pravesh-sharma:
This will be fixed with #5083

anilsahoo20:
Tested and verified on candidate build: https://developer.pgadmin.org/builds/2024-02-05-1/
Environment: Debian 11

Issue #7115, opened 2024-01-09
Query Format


JuanLuisNL:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

---- If we write a statement like this:

SELECT pais, apr_idioma FROM estadisticas WHERE COALESCE(apr_idioma,'') = '' ORDER BY pais

---- When pressing Ctrl+Shift+k

SELECT PAIS,
	APR_IDIOMA
FROM ESTADISTICAS
WHERE COALESCE(APR_IDIOMA,

							'') = ''
ORDER BY PAIS 

**To Reproduce**

Press Ctrl+Shift+k

**Expected behavior**

SELECT PAIS,
	APR_IDIOMA
FROM ESTADISTICAS
WHERE COALESCE(APR_IDIOMA,'') = ''
ORDER BY PAIS

**Desktop (please complete the following information):**
 -  WIdonws
 - Version: all

nikhil-mohite:
Duplicate of #5083

Issue #6990, opened 2023-11-22
sql format err for `@>` 


gshmu:
format the sql: `select array['a', 'b'] @> array['a'];`

![image](https://github.com/pgadmin-org/pgadmin4/assets/1797676/bae3ef73-115d-4a3a-9fdc-56ff0444e748)

adityatoshniwal:
Duplicates #5083

Issue #5083, opened 2022-05-21
"FORMAT SQL"  command messes up -> operator (RM #7400)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/7400***
*Originally created by **Alexis Panagiotopouos** at 2022-05-21 06:51:55 UTC.*

If you have a query containing the '->' or '->>', FORMAt SQL command (shift+ctrl+k) produces wrong SQL.
Example:
<pre><code class="sql">
select '{}'::jsonb->'a'
</code></pre>
becomes:
<pre><code class="sql">
SELECT '{}'::JSONB - > 'a'
</code></pre>
Which is wrong sql (syntax error)

danielsamuels:
`@>` for arrays, too.

slyapustin:
Would like that to be fixed, any ideas on how to better fix that?

adityatoshniwal:
I think we should replace library used formatting the SQL and use - https://www.npmjs.com/package/sql-formatter. It has PostgreSQL dialect support and has better predictable formatting. The existing one works for simple queries, but messes complex queries.
CC: @dpage @akshay-joshi

dpage:
If it truly is better, then I have no problem switching.

pravesh-sharma:
We found that we can not change case of keywords and identifiers separately in the suggested library (which is a limitation as we support it currently) and a feature request is [pending](https://github.com/sql-formatter-org/sql-formatter/issues/237) for the same on library. So, for now we are not moving to the new library

anilsahoo20:
Tested and verified on candidate build: https://developer.pgadmin.org/builds/2024-02-05-1/
Environment: Debian 11

we-sell-bags:
> I think we should replace library used formatting the SQL and use - https://www.npmjs.com/package/sql-formatter. It has PostgreSQL dialect support and has better predictable formatting. The existing one works for simple queries, but messes complex queries. CC: @dpage @akshay-joshi

Why use a "dead" project?
The development of this formatter is currently in maintenance mode. Bugs will get fixed if feasible, but new features will likely not be added.

I have started a new SQL formatting tool: [prettier-plugin-sql-cst](https://github.com/nene/prettier-plugin-sql-cst).

    It solves several problems which can't be fixed in SQL Formatter because of fundamental problems in its arhictecture.
    It makes use of the Prettier layout algorithm, doing a better job of splitting long expressions to multiple lines.
[...]

PR #7169, opened 2024-01-31
Fixed an issue where format sql was messing up operator. #5083

pravesh-sharma:
- Also fixes https://github.com/pgadmin-org/pgadmin4/issues/6785, https://github.com/pgadmin-org/pgadmin4/issues/6990, https://github.com/pgadmin-org/pgadmin4/issues/7115.
- Using new library sql-formattor for SQL formatting.
- Added new options data type case, expression width, function case, lines between queries, logical operator new line and new line before semicolon for SQL formatting.
- Removed old options comma-first notations, re-indent aligned, re-indent, strip comments and wrap after N characters as not available in the new library.
- Capitalize casing has been removed as it was not supported by the new library using preserve casing instead.


adityatoshniwal (changes requested):
1. The backend API is not cleaned up.
2. The un-used lib is not removed from requirements.txt.
3. Commit message has no details on what new features are added and what removed in detail.
4. The commit message doesn't mention all other formatting related issues which will be fixed with this commit.
5. The commit message doesn't mention which lib is removed and which is added.

pravesh-sharma:
> 1. The backend API is not cleaned up.
> 2. The un-used lib is not removed from requirements.txt.
> 3. Commit message has no details on what new features are added and what removed in detail.
> 4. The commit message doesn't mention all other formatting related issues which will be fixed with this commit.
> 5. The commit message doesn't mention which lib is removed and which is added.
> 
> This is a very lazy PR.

I have removed the backend API. Also, I updated the commit. I did not mention the other issues in the commit as they were already linked in the original issue I have fixed. However, I have mentioned them now.

adityatoshniwal (approved):

PR #7200, opened 2024-02-08
Issue/7138 json log format

florian-sabonchi:
closes #7138

PR #7203, opened 2024-02-09
Json log format

florian-sabonchi:
closes #7200


web/pgadmin/__init__.py
```diff
@@ -265,11 +265,14 @@ def create_app(app_name=None):
         logger.addHandler(fh)
 
     # Console logging
-    ch = logging.StreamHandler()
-    ch.setLevel(config.CONSOLE_LOG_LEVEL)
-    ch.setFormatter(logging.Formatter(config.CONSOLE_LOG_FORMAT))
-    app.logger.addHandler(ch)
-    logger.addHandler(ch)
+
+    formatter = JsonFormatter(config.CONSOLE_LOG_FORMAT)
```
khushboovashi, line 269: We need both options, plain text and JSON logging. So, keep the original (plain text) as a default and the JSON logging configurable.

web/pgAdmin4.py
```diff
@@ -13,7 +13,7 @@
 
 import sys
 if sys.version_info <= (3, 9):
-    import select
+    pass
```
khushboovashi, line 16: Does this change work with Python 3.9?

khushboovashi (changes requested):

web/pgAdmin4.py
florian-sabonchi: No, I don't think so, I missed the requirements

khushboovashi (changes requested):
Looks good to me but currently this is applicable only for console logs, we can apply same for the file based logs, too.

florian-sabonchi:
> Looks good to me but currently this is applicable only for console logs, we can apply same for the file based logs, too.

sound's good to me.

khushboovashi:
> > Looks good to me but currently this is applicable only for console logs, we can apply same for the file based logs, too.
> 
> sound's good to me.

Please accomodate this change in this PR.

florian-sabonchi:
> > > Looks good to me but currently this is applicable only for console logs, we can apply same for the file based logs, too.
> > 
> > 
> > sound's good to me.
> 
> Please accomodate this change in this PR.

Yes, sure i will work on it

Issue #7221, opened 2024-02-17
Expose UNIX socket in entrypoint.sh for Docker implementation


DJviolin:
**Describe the solution you'd like**

Can you implement in `entrypoint.sh` to expose UNIX socket for a reverse proxy like Nginx, maybe defined with an environment variable like `PGADMIN_SOCK` or your liking? It will automatically take presence over listen address and port. In a Docker Compose stack, where most of the communication managed with UNIX sockets between containers, current pgAdmin container not supporting such deployment.

The official server deployment docs giving examples such Gunicorn setup: https://www.pgadmin.org/docs/pgadmin4/latest/server_deployment.html

DJviolin:
I use this possible implementation at the end of my monkey patched `entrypoint.sh`:

```sh
if [ -n "$PGADMIN_SOCK" ]; then
    BIND_ADDRESS="unix:${PGADMIN_SOCK}"
else
	if [ -n "${PGADMIN_ENABLE_TLS}" ]; then
		 BIND_ADDRESS="${PGADMIN_LISTEN_ADDRESS:-[::]}:${PGADMIN_LISTEN_PORT:-443}"
	else
		 BIND_ADDRESS="${PGADMIN_LISTEN_ADDRESS:-[::]}:${PGADMIN_LISTEN_PORT:-80}"
	fi
fi

if [ -n "${PGADMIN_ENABLE_TLS}" ]; then
    exec /venv/bin/gunicorn --limit-request-line "${GUNICORN_LIMIT_REQUEST_LINE:-8190}" --timeout "${TIMEOUT}" --bind "$BIND_ADDRESS" -w 1 --threads "${GUNICORN_THREADS:-25}" --access-logfile "${GUNICORN_ACCESS_LOGFILE:--}" --keyfile /certs/server.key --certfile /certs/server.cert -c gunicorn_config.py run_pgadmin:app
else
    exec /venv/bin/gunicorn --limit-request-line "${GUNICORN_LIMIT_REQUEST_LINE:-8190}" --timeout "${TIMEOUT}" --bind "$BIND_ADDRESS" -w 1 --threads "${GUNICORN_THREADS:-25}" --access-logfile "${GUNICORN_ACCESS_LOGFILE:--}" -c gunicorn_config.py run_pgadmin:app
fi
```

Where `PGADMIN_SOCK` is `/run/pgadmin/pgadmin.sock`. I use separate folders for sockets in Docker Compose for convenience with volumes.

[...]

PR #7224, opened 2024-02-19
Expose UNIX socket for Docker

DJviolin:
In this `entrypoint.sh` modification, we expose an UNIX socket for Docker placed at `/run/pgadmin/pgadmin.sock`. If `PGADMIN_ENABLE_SOCK` environment variable is not empty, it will create `/run/pgadmin` folder with `pgadmin:root` rights and gunicorn creating it's socket there.

I decided to not give the location from the environment variable, because user can create it anywhere in the file system, raising possible problems. Also placed in a folder, not directly in `/run`, because in a Docker Compose stack helping to create a separate named volume only for this socket.

Needs testing from others.

This pull request is originally mentoined in my issue here: https://github.com/pgadmin-org/pgadmin4/issues/7221


DJviolin:
mkdor and chown needs to happen in Dockerfile, making some changes.

DJviolin:
I moved `mkdir` from `entrypoint.sh` into `Dockerfile`, because for some reason, not always ran with `pgadmin:root` rights.

Dockerfile
```diff
@@ -200,8 +200,8 @@ RUN apk add \
     /venv/bin/python3 -m pip install --no-cache-dir gunicorn==20.1.0 && \
     find / -type d -name '__pycache__' -exec rm -rf {} + && \
     useradd -r -u 5050 -g root -s /sbin/nologin pgadmin && \
-    mkdir -p /var/lib/pgadmin && \
-    chown pgadmin:root /var/lib/pgadmin && \
+    mkdir -p /run/pgadmin /var/lib/pgadmin && \
```
khushboovashi, line 203: Do you think this will work? It will create /run/pgadmin directory only, not the /var/lib/pgadmin.

khushboovashi (changes requested):

Dockerfile
DJviolin: The creation and rights of the `/var/lib/pgadmin` folder is in the original Dockerfile, my modifications are only adding `/run/pgadmin` folder and needed rights for that commands.

I'm not running `pgadmin4` Dockerfile in my project directly, I'm importing it as a base image, I cannot test it right now. I suggest leaving it there for later testing, than it can be removed or not.

DJviolin: I tested the image now, both folder created without problem. `mkdir` can take multiple arguments for directory(ies), In my images I use it this way in `debian` and `alpine` as well:

* [MAN page](https://linux.die.net/man/1/mkdir)
* [BusyBox coreutils `mkdir` source](https://github.com/mirror/busybox/blob/2d4a3d9e6c1493a9520b907e07a41aca90cdfd94/coreutils/mkdir.c#L100)

I run Docker in Windows+WSL2 Ubuntu, if it's not creating both folders for everyone, perhabs good idea to separate the commands? Have you tested it on your side?

khushboovashi:
@DJviolin, can you please update your branch?

DJviolin:
I pulled the latest branch. Can you please verify the behavior of the mkdir command with multiple dirs? In my end, works without problem, creating multiple directories. If this is a problem, I gladly make it a separate command for the second dir. What do you suggest?

PR #7231, opened 2024-02-22
Added documentation for Dashboard's System Statistics tab. #7148

anilsahoo20:


khushboovashi (changes requested):
- The graphs will display GB, MB, Bytes depending on the number, not necessary it always displays in GB. 
- Please take the screen-shots with grey border.

anilsahoo20:
> * The graphs will display GB, MB, Bytes depending on the number, not necessary it always displays in GB. : As the values are large so it is coming in most of the places in GB, but there are some graphs where MB is coming as the value is low, e.g. Swap Memory>Used.
> * Please take the screen-shots with grey border. : Updated

PR #7239, opened 2024-02-23
fix: invalidating oauth2 session

florian-sabonchi:


web/pgadmin/authenticate/oauth2.py
```diff
@@ -69,11 +69,19 @@ def oauth_authorize():
                      methods=['GET', 'POST'])
     @pgCSRFProtect.exempt
     def oauth_logout():
+        logout_url = session['oauth2_logout_url']
```
yogeshmahajan-1903, line 72: Add check for 'oauth2_logout_url' key in session before assignment.

web/pgadmin/authenticate/oauth2.py
```diff
@@ -201,6 +209,8 @@ def get_user_profile(self):
             self.oauth2_current_client].authorize_access_token()
 
         session['pass_enc_key'] = session['oauth2_token']['access_token']
+        session['oauth2_logout_url'] = self.oauth2_config[
+            self.oauth2_current_client]['OAUTH2_LOGOUT_URL']
```
yogeshmahajan-1903, line 213: Add key check before getting value for 'OAUTH2_LOGOUT_URL'.

web/pgadmin/authenticate/oauth2.py
yogeshmahajan-1903, line 72: Add check for 'oauth2_logout_url' key in session before assignment.

yogeshmahajan-1903 (changes requested):
Keycheck is required in order not to break existing OAuth2 setups.

yogeshmahajan-1903 (approved):

PR #7265, opened 2024-03-09
Polish up the system_stats not installed message.

anthonydb:
Suggesting a minor cleanup of the message that tells users that the `system_stats` extension is not installed. If the text needs to be changed in additional files, please let me know and I'll do that as well.


adityatoshniwal:
Hi @anthonydb,
You don't need to change the .pot file. It is auto-generated.

anthonydb:
@adityatoshniwal OK, I have restored the `.pot` file to its original state.

Issue #7098, opened 2024-01-04
Add support for EDB Job Scheduler


akshay-joshi:

yogeshmahajan-1903:
This is working fine.

PR #7273, opened 2024-03-12
Added support for EDB Job Scheduler. #7098

akshay-joshi:


akshay-joshi:
> Code looks good to me. I have done basic functional testing and found one issue. While creating a schedule if you select end date and time, you will gate error and MSQL tab will not load. Screenshot 2024-03-14 at 5 02 40 PM Screenshot 2024-03-14 at 5 05 28 PM
> 
> Also, if the job is running, can we see the running status on the properties window?

I am unable to reproduce this issue. For job running status you can select the 'DBMS Job Scheduler' collection node and click on the properties tab.

PR #7279, opened 2024-03-13
Fixed some of pgAdmin CLI bugs. #7116

### LINKED ISSUE #7116: pgAdmin CLI pending bugs
Couple of issues:

1. If I run the command `python setup.py set-prefs aditya.toshniwal14@gmail.com "browser:processes:process_retain_days=true"` then it is success. Here, process retain days is a number.

2. Login with an internal user for some email and run the below cmd. Then go to UI and open the preferences dialog. App crashes. I think we should logout the user if it is inactive.
`python setup.py update-user [aditya.toshniwal14@gmail.com](mailto:aditya.toshniwal14@gmail.com) --nonadmin --inactive --json --no-console`

pravesh-sharma:
Please note that all the bugs have not been fixed hence do not close #7116.


web/pgadmin/static/js/components/FormComponents.jsx
```diff
@@ -426,9 +435,9 @@ InputText.propTypes = {
   inputStyle: PropTypes.object
 };
 
-export function FormInputText({ hasError, required, label, className, helpMessage, testcid, ...props }) {
+export function FormInputText({ hasError, required, label, className, helpMessage, testcid, labelTooltip, ...props }) {
   return (
-    <FormInput required={required} label={label} error={hasError} className={className} helpMessage={helpMessage} testcid={testcid}>
+    <FormInput required={required} label={label} error={hasError} className={className} helpMessage={helpMessage} testcid={testcid} tooltip={labelTooltip} >
```
adityatoshniwal, line 440: You've used `tooltip={labelTooltip}` at few places which is incorrect.

web/setup.py
```diff
@@ -142,10 +142,15 @@ def add_user(email: str, password: str,
...
                                    typer.Option("--active/--inactive")] = True,
                  console: Optional[bool] = True,
-                 json: Optional[bool] = False
+                 json: Optional[bool] = False,
+                 sqlite_path: Optional[str] = None,
                  ):
         """Add Internal user. """
 
+        # update the sqlite path
+        if sqlite_path is not None:
```
adityatoshniwal, line 151: Instead of repeating this code, you could try creating some common decorator.

web/setup.py
```diff
@@ -593,5 +642,9 @@ def run_migration_for_others():
         run_migration_for_sqlite()
 
 
-if __name__ == "__main__":
+def main():
     app()
+
+
+if __name__ == "__main__":
```
adityatoshniwal, line 649: Why was this required?

adityatoshniwal (changes requested):

web/setup.py
pravesh-sharma: This is to fix running the `pgadmin4-cli` command in the case of the pip wheel package.

Ref: https://github.com/pgadmin-org/pgadmin4/issues/7116#issuecomment-1968685734

PR #7318, opened 2024-03-25
Add password to data in update_user

### LINKED ISSUE #7304: setup.py update-user doesn't change password
***Please note that security bugs or issues should be reported to [[security@pgadmin.org](mailto:security@pgadmin.org)](mailto:security@pgadmin.org).***

**Describe the bug**

The `setup.py update-user` command does not update my password. After running the command with a new password, the UI does not recognize the new password for user login, but it still accepts the old password. This issue prevents me from successfully changing my passwords via command line.

**To Reproduce**

1. Create a user with the `setup.py add-user` command:

```
bash-4.4$ setup.py add-user user@example.com pass123
User Details            
+---------------------------------+
| Field       | Value             |
|-------------+-------------------|
| Email       | user@example.com |
| auth_source | internal          |
| role        | Non-admin         |
| active      | True              |
+---------------------------------+
```

2. Verify the user can log into the pgAdmin GUI using the password `pass123`.
3. Attempt to update the user's password with the `setup.py update-user` command:

```
bash-4.4$ setup update-user user@example.com --password newpass123
+---------------------------------+
| Field       | Value             |
|-------------+-------------------|
| Username    | user@example.com |
| Email       | user@example.com |
| auth_source | internal          |
| role        | Non-admin         |
| active      | True              |
+---------------------------------+
```

4. Try to log into the pgAdmin GUI using the new password `newpass123` and observe the failure.
5. Confirm that logging in with the old password `pass123` still works.
6. Update the password through the GUI to `thirdpass123` and verify that you can log in using `thirdpass123`.

**Expected behavior**

After using the `setup.py update-user` command to update a user's password, the new password should be recognized by the system, allowing me to log in with the new credential. The old password should no longer be valid.

**Error message**

No explicit error message is displayed in the terminal. However, the pgAdmin GUI login process fails with an incorrect password error when attempting to use the new password set via the `setup.py update-user` command.

**Screenshots**

pgadmin-version

- OS: [Linux Container]
- Version: [8.4]
- Mode: [Server]
- Browser (if running in server mode): [brave/chrome]
- Package type: [python/container]

**Additional context**

benjaminjb:
In the cli update_user command, the password was not being propagated to the correct fields for validate_user.

Fixes #7304

### DIFF
--- a/web/setup.py
+++ b/web/setup.py
@@ -47,7 +47,8 @@ from typing import Optional, List
 from typing_extensions import Annotated
 from pgadmin.utils.constants import INTERNAL, LDAP, OAUTH2, \
     KERBEROS, WEBSERVER
-from pgadmin.tools.user_management import create_user, delete_user, update_user
+from pgadmin.tools.user_management import create_user, delete_user, \
+    update_user as user_management_update_user
 from enum import Enum
 from flask_babel import gettext
 
@@ -261,7 +262,7 @@ class ManageUsers:
             if not uid:
                 print("User not found")
             else:
-                status, msg = update_user(uid, data)
+                status, msg = user_management_update_user(uid, data)
                 if status:
                     _user = ManageUsers.get_users_from_db(username=email,
                                                           auth_source=INTERNAL,
@@ -347,7 +348,7 @@ class ManageUsers:
             if not uid:
                 print("User not found")
             else:
-                status, msg = update_user(uid, data)
+                status, msg = user_management_update_user(uid, data)
                 if status:
                     _user = ManageUsers.get_users_from_db(
                         username=username,
--- a/web/setup.py
+++ b/web/setup.py
@@ -47,7 +47,8 @@ from typing import Optional, List
 from typing_extensions import Annotated
 from pgadmin.utils.constants import INTERNAL, LDAP, OAUTH2, \
     KERBEROS, WEBSERVER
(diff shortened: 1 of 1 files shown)


benjaminjb:
From a quick skim, it looks like the `feature tests` steps are failing in a lot of the current PRs?

adityatoshniwal:
> From a quick skim, it looks like the `feature tests` steps are failing in a lot of the current PRs?

Tests are fixed now. You can rebase your PR.

web/setup.py
```diff
@@ -47,7 +47,8 @@
 from typing_extensions import Annotated
 from pgadmin.utils.constants import INTERNAL, LDAP, OAUTH2, \
     KERBEROS, WEBSERVER
-from pgadmin.tools.user_management import create_user, delete_user, update_user
+from pgadmin.tools.user_management import create_user, delete_user, \
+    update_user as update_user_internal
```
khushboovashi, line 51: The update_user will be used to update both internal and external auth users. So, please change the update_user_internal alias or leave it as the default update_user.

khushboovashi (changes requested):

web/setup.py
benjaminjb: OK, changed -- is `user_management_update_user` better? I'm just trying to avoid a function here called `update_user` calling another function called `update_user`

Issue #5611, opened 2022-12-06
Add support for provider, deterministic, version and RULES parameter while creating Collation.


akshay-joshi:
Add new controls for provider, deterministic, and version in the **Definition** tab of the collation dialog.

anilsahoo20:
1. if control has some default value, then control must show that default value when we initially see the control if some other controls are not dependents on the current one. provider default value is libc, deterministic default value is true according to the Postgres documentation, but it is empty.
2. wrong sql gets generated for different combination of selection of control.
3. Rules control is getting disabled on selection of provider icu and deterministic true
4. Properties tab shows Rules as i, but no rule is given while creating the collation
5. After giving LC_TYPE, LC_collate in create dialog, it asks for locale, but as per docs, if locale given then the foremer 2 controls not needed, if former 2 given then locale not needed.


![Image](https://github.com/pgadmin-org/pgadmin4/assets/122535205/f4f67492-6621-44b8-81e7-476133abfb61)
![Image](https://github.com/pgadmin-org/pgadmin4/assets/122535205/cc69381a-dbc7-4bbf-9d7e-0749cfd06468)
![Image](https://github.com/pgadmin-org/pgadmin4/assets/122535205/2af506e6-192f-4395-8932-9c6113a8a53f)
![Image](https://github.com/pgadmin-org/pgadmin4/assets/122535205/d61590c1-e66c-46a4-9338-b08c5d18bc69)
![Image](https://github.com/pgadmin-org/pgadmin4/assets/122535205/52f8ae89-3cbb-40bd-a75d-cbdfa1c5f2fd)

anilsahoo20:
@JyotiEdb 
I found that the locale provide is empty in edit dialog, but in properties tab it is showing up. Can you check this.

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-04-01/
Environment: macOs ventura 13.5.1
package: arm64

PR #7319, opened 2024-03-26
Fixed issues in create collation. #5611

JyotiEdb:


web/pgadmin/browser/server_groups/servers/databases/schemas/collations/static/js/collation.ui.js
```diff
@@ -83,13 +86,45 @@ export default class CollationSchema extends BaseUISchema {
...
+        readonly: function (state) { return !obj.isNew(state); },
+        options: [{
+          label: gettext('icu'),
+          value: 'icu',
+        }, {
+          label: gettext('libc'),
+          value: 'libc',
+        }],
+        min_version: 110000,
+        deps: ['copy_collation', 'provider'],
```
anilsahoo20, line 101: Why provider added in deps?

web/pgadmin/browser/server_groups/servers/databases/schemas/collations/static/js/collation.ui.js
```diff
@@ -100,77 +135,73 @@ export default class CollationSchema extends BaseUISchema {
...
-          value: 'icu',
-        }, {
-          label: gettext('libc'),
-          value: 'libc',
-        }],
+        mode: ['properties', 'edit', 'create'],
         min_version: 120000,
-        deps: ['copy_collation'],
+        helpMessageMode: ['edit', 'create'],
+        deps: ['copy_collation', 'is_deterministic', 'provider'],
```
anilsahoo20, line 154: Why is_deterministic and provider are added as deps?

web/pgadmin/browser/server_groups/servers/databases/schemas/collations/static/js/collation.ui.js
```diff
@@ -22,6 +22,9 @@ export default class CollationSchema extends BaseUISchema {
       lc_type: undefined,
       lc_collate: undefined,
       description: undefined,
+      provider: 'libc',
+      is_deterministic: true,
+      flag_is_deterministic: false,
```
anilsahoo20, line 27: What is the use of flag_is_deterministic if we are using is_deterministic already?

anilsahoo20 (commented):
Please check,

1. The error icon is only visible on Copy collation, still when it is disabled, it should be visible for the field which is required and empty.
2. For provider=libc, the locale field must be enabled, if user enters in LC_TYPE and LC_COLLATE then locale will be disabled, if userd enters locale, the earlier 2 fields will be disabled.
3. If you select any provider and enter the dependent fields value, and then change the provider, if any field is disabled then then it should be empty and the dependent field must be enabled.(wrong MSQL getting generated)

Issue #7248, opened 2024-02-28
Rollback and commit button gets activated for sometime on getting any error for SQL query or cancelling the query.


anilsahoo20:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

If we run some query and the some error occurs or if we cancel the query in between, the rollback and commit button gets activated for some seconds.

**To Reproduce**

1. Open query tool
2. Run any wrong query
4. See the rollback and commit button gets activated for some seconds

**Expected behavior**

These buttons should not get activated.

<img width="1249" alt="Screenshot 2024-02-28 at 9 22 41 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/122535205/bd73e1f7-3e94-43b7-8fbc-f60da18e79de">

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10, OSX monterey]
 - Version: [e.g. 6.15]
 - Mode: [e.g Desktop or Server]
 - Browser (if running in server mode): [e.g. chrome, safari]
 - Package type: [e.g. RPM, DEB, Python, Container, etc.]

**Additional context**

JyotiEdb:
@anilsahoo20 
All working fine,
But when we run wrong query and auto-commit is off, we get error, and after clicking commit button its shows massage of rollback, is it expected behaviour?

anilsahoo20:
@JyotiEdb,
yes it is correct behaviour. Because there is nothing to commit as the error comes in. so the commit and rollback button will rollback always.

JyotiEdb:
Tested and verified on snapshot build:
https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-03-28/

PR #7322, opened 2024-03-27
Fix rollback and commit button activation on execute button click. #7248

anilsahoo20:


web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -950,6 +950,14 @@ def poll(trans_id):
                 is_thread_alive = True
                 break
 
+    # if transaction object is instance of QueryToolCommand
+    # and transaction aborted for some reason then issue a
+    # rollback to cleanup
+    if isinstance(trans_obj, QueryToolCommand):
```
khushboovashi, line 956: For any particular scenario, did you put this check before polling?

khushboovashi (changes requested):

web/pgadmin/tools/sqleditor/__init__.py
anilsahoo20: Yes there was one scenario, where rollback is happening after we are getting the transaction_in_error status. so the response holds the old status. so I in place of putting this check in every if or elif block, I put it before polling.

khushboovashi (approved):

PR #7328, opened 2024-03-28
Fixed an issue with empty locale provider while editing collation. #5611

JyotiEdb:


web/pgadmin/browser/server_groups/servers/databases/schemas/collations/static/js/collation.ui.js
```diff
@@ -87,7 +87,7 @@ export default class CollationSchema extends BaseUISchema {
       },
       {
         id: 'provider', label: gettext('Locale Provider'),
-        editable: false, type: 'select',mode: ['create', 'edit'], group: gettext('Definition'),
+        editable: false, type: 'select',mode: ['create'],
```
anilsahoo20, line 90: Any reason you removed, group? On General tab Locale provider field is showing up.

anilsahoo20 (changes requested):

Issue #4378, opened 2021-06-09
PgAdmin4 for Flatpak (RM #6523)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/6523***
*Originally created by **Karsten Lenz** at 2021-06-09 10:25:44 UTC.*

It would be helpful to have PgAdmin available for Flatpak, this will also include the Arch Linux based community where PgAdmin is a nightmare.
On Arch the lates working version is 4.30, there is a non working 5.25 ARU Package which whill not compile.

With Flatpak it is possible add many Distributions with just one Flatpak package.

The Hardware support from for example Manjaro is fantastic comparing to other Distributions.

Regards

lestcape:
I don't know the state of this, but should probably be revisited. Rolling release distros are not included on the build system and pip is now blocked by some distros: https://github.com/python/cpython/issues/102134#issuecomment-1445428402 While the restriction can be avoided, I think that this is already too much and the need for a solution that covers all Linux environments is becoming evident. If for some reason it's not flatpak, could it be snap or could be an appimage? I have not real preference, but at less one:

https://snapcraft.io/ 
https://appimage.org/

albertoridolfi:
Offering the software as a flatpak would lower the burden on developers and would broaden the distros in which the app will run. Instead of the devs needing to create multiple packages for multiple distributions and their multiple versions, devs would need to make just one package, and it will run everywhere.
I was even thinking of volunteering myself to create a flatpak for pgAdmin4 but its stack seem complicated to me since I don't work with nodejs nor with the web front-end tech stuff that pgAdmin4 uses.
But for pgAdmin4 devs, creating a flatpak would be a no-brainer. A lot easier than maintaining multiple builds for multiple versions of multiple distros.

https://docs.flatpak.org/en/latest/first-build.html

faveoled:
I wrote the build files: 
[org.pgadmin.pgadmin4.zip](https://github.com/pgadmin-org/pgadmin4/files/14076317/org.pgadmin.pgadmin4.zip)

If someone is interested they should provide an AppStream metadata file for the store.
E.g. using this UI: https://www.freedesktop.org/software/appstream/metainfocreator/#/guiapp

Flatpak can be tested locally this way:
```
flatpak run org.flatpak.Builder --force-clean --sandbox --user --install --ccache builddir org.pgadmin.pgadmin4.yaml
flatpak run org.pgadmin.pgadmin4
```

[...]

PR #7333, opened 2024-04-01
Add metainfo file for pgadmin4

razzeee:
This is just a basic version of a metainfo file, it could be improved in various aspects, but having something basic already in tree would already be helpful.

Mostly done for https://github.com/pgadmin-org/pgadmin4/issues/4378#issuecomment-1913611551 but can be used by any linux distro and I also heared there are non linux appstores using the format. (not sure if relevant in pgadmins usergroups)

Flathub asked if we can have this file upstream https://github.com/flathub/flathub/pull/5105#discussion_r1545472650

Ideally, your release process would also add a release in this file.

### DIFF
--- a/pkg/linux/org.pgadmin.pgadmin4.metainfo.xml
+++ b/pkg/linux/org.pgadmin.pgadmin4.metainfo.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>org.pgadmin.pgadmin4</id>
+  <name>pgAdmin 4</name>
+  <project_license>PostgreSQL</project_license>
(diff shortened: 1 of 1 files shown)


razzeee:
I should update this to include 8.5.0

adityatoshniwal:
Cherry picked in the PR https://github.com/pgadmin-org/pgadmin4/pull/7359 as both are for flatpak.

razzeee:
Just for context, while I'm doing this for flatpak, it's widely used on different Linux distribution channels and even on other platforms (but I hear those are mostly automotive for now)

adityatoshniwal:
@razzeee Shouldn't this file be autogenerated and updated based on current release?

razzeee:
I don't know your release process/tooling, but if that's possible, that would be great

PR #7359, opened 2024-04-07
Patches for running in a flatpak

razzeee:
These are the needed patches we would need to be able to run pgadmin in a flatpak.

They stem of the work in https://github.com/flathub/flathub/pull/5105/files where we're doing this (much) worse with `sed`.

I think there might be a better solution for parts of this, but I'm not sure how to get there. Having this up-streamed at least would help us for the time being.

### DIFF
--- a/runtime/src/js/misc.js
+++ b/runtime/src/js/misc.js
@@ -26,6 +26,10 @@ const createDir = (dirName) => {
   }
 };
 
+export const insideFlatpak = () => {
+  return platform() === 'linux' && fs.existsSync('/.flatpak-info');
+}
+
 // This function is used to get the python executable path
 // based on the platform. Use this for deployment.
 const getPythonPath = () => {
@@ -37,8 +41,11 @@ const getPythonPath = () => {
   case 'darwin':
     pythonPath = '../../Frameworks/Python.framework/Versions/Current/bin/python3';
     break;
-  case 'linux':
-    pythonPath = '../venv/bin/python3';
+    case 'linux':
+      pythonPath = '../venv/bin/python3';
+      if (insideFlatpak()) {
+        pythonPath = '/usr/bin/python';
+      }
     break;
   default:
     if (platform().startsWith('win')) {
--- a/runtime/src/js/pgadmin.js
+++ b/runtime/src/js/pgadmin.js
@@ -29,6 +29,10 @@ let pythonPath = misc.getPythonPath();
 let pgadminFile = '../web/pgAdmin4.py';
 let configFile = '../web/config.py';
 
+if (insideFlatpak()) {
+  pgadminFile = '/app/pgAdmin4/web/pgAdmin4.py';
+}
+
 // Override the paths above, if a developer needs to
 if (fs.existsSync('dev_config.json')) {
   try {


web/config.py
```diff
@@ -226,6 +226,8 @@
 else:
     if SERVER_MODE:
         DATA_DIR = os.path.join('/var/lib/', APP_PATH)
+    elif os.path.exists("/.flatpak-info"):
```
khushboovashi, line 229: While installing, you can override this config setting into the config_local or config_system file, so you can remove it from here. Apart from this, other changes look good to me.

khushboovashi (changes requested):

web/config.py
razzeee: Can I use system variables in there, like needed here, for `XDG_CONFIG_HOME`?

adityatoshniwal: config_distro.py is used of for package distribution related config. https://www.pgadmin.org/docs/pgadmin4/8.5/config_py.html#the-config-py-file.

razzeee: Is this done correctly? I don't think it works?

https://github.com/razzeee/org.pgadmin.pgadmin4/pull/2#issuecomment-2048260923

khushboovashi: > Is this done correctly? I don't think it works?
> 
> [razzeee/org.pgadmin.pgadmin4#2 (comment)](https://github.com/razzeee/org.pgadmin.pgadmin4/pull/2#issuecomment-2048260923)

Can you please specify, what you tried and what's not working?

razzeee: Tried the distro_config.py like this

image


And it doesn't seem to stick/lead to the same path as it did with changing the data_dir in the config.py itself

adityatoshniwal: Can you check if there is any config_local or config_system in your env which is overriding it?
Where did you put the config_distro? It should be beside config.py. try debugging web/pgadmin/evaluate_config.py where it is read and merged.

razzeee: Thank you, I think I've figured this out now

razzeee: Removed this commit as it's handled in the config of the flatpak now

razzeee:
Unsure why the tests on macOS fail

adityatoshniwal:
> Unsure why the tests on macOS fail

You can ignore that. Its a JS test case, which is failing randomly.

akshay-joshi:
@razzeee 

Can you please raise a combined PR for this and https://github.com/pgadmin-org/pgadmin4/pull/7333?

razzeee:
@akshay-joshi would it be fine to cherry pick that into this branch?

adityatoshniwal:
@razzeee You've made it compatible when running from flatpak. But the automated build process need to include flatpak build and publish it flatpak to flathub just like docker hub.

razzeee:
> @razzeee You've made it compatible when running from flatpak. But the automated build process need to include flatpak build and publish it flatpak to flathub just like docker hub.

There's no way (for me) to publish it to flathub (and thus it doesn't make much sense to build the flatpak here) - you would need to reach out and ask for credentials. "Direct uploads" are a gated feature right now and I'm not sure, if @barthalion would allow your project into that test group.

The other way would be the repo on the flathub github, to manage the manifest. But that would ideally need this to be merged.

What I was asking about, was if there was automatic tooling generating the release notes/versions.

khushboovashi:
Hi @razzeee,

We are not planning to support the Flatpak packages yet, so it doesn't make sense to add the metainfo file to the repo. However, we can commit the path-related changes. Please remove the metainfo file from the PR so we can merge the remaining changes.

razzeee:
So you're saying, I should revert the combination of the two PRs and split this up again into the two PRs like it was at the start?

khushboovashi:
> So you're saying, I should revert the combination of the two PRs and split this up again into the two PRs like it was at the start?

There is no point in raising a different PR for the meta info file because we are not going to commit that file. So, just remove that file from this PR so I can merge PATH-related changes, which are required for Flatpak.

hfiguiere:
> We are not planning to support the Flatpak packages yet, so it doesn't make sense to add the metainfo file to the repo. 

This assumption is incorrect. AppStream files are NOT specific to Flatpak in anyway. GNOME and KDE app ship them and they get installed by distro pacakge when they exists. It's then used by GNOME Software or KDE Discover or equivalent to show information about the app.

khushboovashi:
Hello @razzeee, we have reverted this commit as the pgAdmin app was crashing.
Can you please open a new PR with the updated code?

razzeee:
Where did you do this? There is nothing linked here and you did not use the revert button in github.

Found it here https://github.com/pgadmin-org/pgadmin4/commit/dfaea944f024ab90139654957ff0ec8f6c17eb8c but it has no description, no reason etc. for all I can tell, this was arbitrary.

I guess I can just revert that code and put that into a new PR, but doubtful, that you want that.

razzeee:
It also seems to have been pushed directly to master?

razzeee:
Done here https://github.com/pgadmin-org/pgadmin4/pull/7592

akshay-joshi:
Hi,

After merging the patch, pgAdmin 4 started hanging on the splash screen during the candidate build for release. Given the tight release schedule, we couldn't afford to wait for a proper fix. Therefore, we reverted the patch using "git revert <commit-id>."

We haven't investigated the cause of the hang on the splash screen due to the urgency of the release.

razzeee:
> Hi,
> 
> After merging the patch, pgAdmin 4 started hanging on the splash screen during the candidate build for release. Given the tight release schedule, we couldn't afford to wait for a proper fix. Therefore, we reverted the patch using "git revert ."
> 
> We haven't investigated the cause of the hang on the splash screen due to the urgency of the release.

This reads as in, you are not sure that it's even this commit, did you revert more? I'm trying to figure out, what your expectation for what I do in the new PR is.

akshay-joshi:
> > Hi,
> > After merging the patch, pgAdmin 4 started hanging on the splash screen during the candidate build for release. Given the tight release schedule, we couldn't afford to wait for a proper fix. Therefore, we reverted the patch using "git revert ."
> > We haven't investigated the cause of the hang on the splash screen due to the urgency of the release.
> 
> This reads as in, you are not sure that it's even this commit, did you revert more? I'm trying to figure out, what your expectation for what I do in the new PR is.

Let me debug the code and investigate why pgAdmin hangs on splash screen. I'll give review comment on new PR.

Issue #7282, opened 2024-03-14
XSS vulnerability in /settings/store endpoint


felixtoe:
**Describe the bug**

XSS vulnerability in /settings/store endpoint

**To Reproduce**

1. Use a man-in-the-middle proxy like `mitmproxy`
2. Intercept POST request to /settings/store (is e. g. being called when resizing left menu bar)
3. Modify the request body like in the attached screenshot (`... \"children\": [{\"id\":\"+3'-alert('XSS')-'\", ...`) and resume request
4. Reload pgAdmin homepage. Alert box is popping up.

Tested with pgAdmin4 8.0 and 8.3.

**Screenshots**

Are attached :)
![XSS1](https://github.com/pgadmin-org/pgadmin4/assets/26927719/c9dcdd10-e7fa-4a9e-a300-f9f934636bbf)
![XSS2](https://github.com/pgadmin-org/pgadmin4/assets/26927719/0cd05831-805e-46a7-908b-7d5aabc71727)

khushboovashi:
Issue has been fixed.

PR #7367, opened 2024-04-10
Fix the XSS vulnerability in the /settings/store endpoint.#7282

yogeshmahajan-1903:
Also fixed features failures due to sonarqube fixes.

### DIFF
--- a/web/pgadmin/browser/static/js/node.js
+++ b/web/pgadmin/browser/static/js/node.js
@@ -139,7 +139,7 @@ define('pgadmin.browser.node', [
           },
           enable: _.isFunction(self.canEdit) ?
             function() {
-              return !!(self.canEdit(arguments));
+              return !!(self.canEdit(...arguments));
             } : (!!self.canEdit),
         }]);
       }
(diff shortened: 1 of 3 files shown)


web/pgadmin/browser/static/js/node.js
```diff
@@ -159,7 +159,7 @@ define('pgadmin.browser.node', [
           },
           enable: _.isFunction(self.canDrop) ?
             function() {
-              return !!(self.canDrop(arguments));
+              return !!(self.canDrop.apply(self, arguments));
```
adityatoshniwal, line 162: we have remove this change to fix sonarqube. please do not add back.
To fix this change it to `self.canDrop(...arguments)`, arguments is array here.
there are 2 more places apart from this, please change there as well.

adityatoshniwal (changes requested):

web/pgadmin/browser/static/js/node.js
yogeshmahajan-1903: Done.

Issue #3981, opened 2020-11-24
show PostgreSQL server log (RM #6037)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/6037***
*Originally created by **Frank Limpert** at 2020-11-24 15:09:50 UTC.*

In pgAdmin3 (see attached screen shot), I can have a look inside the most recent PostgreSQL server log file. This helps me especially to tune SQL commands inside of functions using <code>auto_explain</code> et al. This feature would be especially helpful for developers not having shell access to their PostgreSQL server.

Older PostgreSQL servers might need <code>EXTENSION adminpack</code> for this. Newer PostgreSQL versions have <code>pg_read_file()</code>.

ronwildt:
@akshay-joshi Thank you for identifying my enhancement request as a duplicate #6696 What is the timeframe for implementing this feature and why was it removed from pgadmin4 in the first place, my I ask?  Thanks!

akshay-joshi:
@ronwildt 

We have completely rewritten pgAdmin 4 in React/ Python and other web technologies. pgAdmin III is c++ based standalone application. pgAdmin 4 is a web + desktop application. We will prioritize it.

ronwildt:
@akshay-joshi The info is appreciated.  When might the team be adding this to the backlog and what might be the ETA for delivery of the feature?  (looking to know for planning purposes on our end, fyi.) Thanks! :)

akshay-joshi:
@ronwildt 

Will see the feasibility and if possible will try to add this to the next release 7.7 on 21st September.

ronwildt:
@akshay-joshi That would be totally awesome!  Didn't expect the team might be able to do so that quickly.  Looking forward to it.  Have a great weekend, thx. :)

akshay-joshi:
@ronwildt

We have to check the feasibility, this feature is not straightforward and seems to be very complex. So definitely it won't be there in the next release v7.7

ronwildt:
@akshay-joshi We appreciate the update and I figured it would not be easy if not already in development.  Any idea when the feature may be planned and implemented?

akshay-joshi:
@ronwildt 

We will check the feasibility/analysis first and then we can priorities.

ronwildt:
@akshay-joshi Sounds good, looking forward to learning about ETA on inclusion in pgadmin4!  Have a great Labor Day weekend! :)

yogeshmahajan-1903:
[...]

PR #7381, opened 2024-04-16
Show PostgreSQL Server Log. #3981

khushboovashi:


akshay-joshi (changes requested):
GUI Related:

- Encountering the error "cannot access the local variable 'final_cols' where it is not associated with a value." when clicking on 'Table Based logs'.
- I think we should shift the entire "Server activity" tab near to "System Statistics"?
- We should display all available log formats and enable only those specified in the 'log_destination'.
- 'View the active session details" and 'View the log details' not working properly, and controls are not visible and overlapped.
- Clicking on the search box of each column is inadvertently sorting the column, which should not occur.
- The toggle button separator should be visible when selecting 'Text'. However, there is no separator between JSON and CSV.
- Remove the colon ":" from "LOG:" and "DEBUG:" from the values under 'Error Severity'.
- Rename the 'Error Severity' column to 'Log Mode' or 'Log Type', or a similar descriptive term. I know the term is from PostgreSQL, but LOG, and DEBUG messages are not errors.
- Rename the label 'Table Based Logs' to 'Show logs in tabular format?' or a similar descriptive term.
- On clicking the 'Table Based Logs' label switch control toggles its state.
- Timestamp/Log prefix not visible in the 'Text' mode.
- We should read all the files. Needs to be confirmed. 
- Unnessary verticle Scroll Bar is visible on the Dashboard. This is not because of this PR but can you please look into it? 
- Logs are missing in the 'Table Based logs' as it is visible in the file-based. Please check once. 
- Reduce the width of the 'Error Severity', and ''View the log details' columns (if possible)

Code Related:

- Add Documentation.
- JavaScript tests are failing.
- Fixed SonarQube issues that were newly added with this PR.
- Add proper comments in the functions wherever applicable. 
- In PgTable.jsx please correct the spelling of 'CLOUMNS'. Not in your PR but we should fix it.

Issue #7163, opened 2024-01-26
Excluding multiple tables


NeptuneTheMystic:
Hello,

In the database backup dialog, Options tab, there is a facility for specifying a table pattern for exclusion.  What is the proper entry for excluding multiple tables?  The pgAdmin documentation only indicates "table pattern."  pg_dump allows multiple occurrences of the -T switch to be used to specify more than one pattern, but it is not clear to me how to do this in pgAdmin.

adityatoshniwal:
Hi @NeptuneTheMystic,
Currently, you can provide only one pattern. This can be a new feature for a future release.

florian-sabonchi:
Hello, someone working on this?

adityatoshniwal:
> Hello, someone working on this?

@florian-sabonchi Nope. It is not assigned to anyone.

florian-sabonchi:
> > Hello, someone working on this?
> 
> @florian-sabonchi Nope. It is not assigned to anyone.

Hello, I will look on it

akshay-joshi:
@florian-sabonchi 

It seems the [Commit](https://github.com/pgadmin-org/pgadmin4/commit/134e6519897af329ca72762ec16d44b9b03bf068) mentioned does not address the specific issue regarding the '-T' or '--exclude-table' switch. It required a new control to be added in the 'Table Options' tab.

The solution you've implemented effectively addresses issues with values containing space-separated lists, such as "Exclude Schema".

Would you like to work on the feature?

florian-sabonchi:
> @florian-sabonchi
> 
> It seems the [Commit](https://github.com/pgadmin-org/pgadmin4/commit/134e6519897af329ca72762ec16d44b9b03bf068) mentioned does not address the specific issue regarding the '-T' or '--exclude-table' switch. It required a new control to be added in the 'Table Options' tab.
> 
> The solution you've implemented effectively addresses issues with values containing space-separated lists, such as "Exclude Schema".
> 
> Would you like to work on the feature?

Hello, thanks for your message, currently not. I will unassign me

anilsahoo20:
Tested and verified on candidate build: https://developer.pgadmin.org/builds/2024-04-29-1/
Environment: macOs Ventura 13.5.1
Package: arm64

PR #7435, opened 2024-04-26
Fixed issue found during testing of #7163.

pravesh-sharma:


akshay-joshi (changes requested):
Create a table in CAPITAL letters like "ABCD", it is not working.
Create a schema in CAPITAL letters and create some tables with space, with capital letters, normal tables, and test properly.

adityatoshniwal:
Isn't `Include table(s) and children` also a pattern?

Issue #6357, opened 2023-06-02
Query Tool: overwrites what I'm typing with unwanted SQL on slow connections


vladr:
**Describe the bug**

"Query Tool" has the following misfeature (using default settings): soon after opening, but not _immediately_ after opening, the "Query" edit area is overwritten with the generic SQL for dropping/(re)creating whatever is currently focused in "Object Explorer".

1. As a user, I don't care about said generic SQL; if I wanted it (which I don't), I can always right-click on the table and go to `Scripts` > `CREATE Script`.
   If I opened "Query Tool", it is because I want to run _**my own**_ SQL, not some random unsolicited default, thank you very much!
   **UPDATE:** it looks like this behavior might be controlled by the (poorly named?) "Copy SQL from main window to query tool?" option (FWIW, I've never seen a "Main window" with any SQL in it, so I saw no reason to disable the option until I started experimenting some more for this bug report);
3. To add insult to injury, when the generic SQL **_is_** inserted, it is inserted as late as several seconds after the tool actually opens (depending on network latency and server load), overwriting whatever I might have already typed there in the brief interval. #$%&!

**To Reproduce**

1. **UPDATE**: Make sure "Copy SQL from main window to query tool?" is set in the "Query Tool" options (this is the default);
2. Select an object (table, schema, ...) in the "Object Explorer" view;
3. Click on "Query Tool" button in toolbar;
4. Start typing your query right away;
   _**N.B. this step is timing-sensitive and may be best reproduced using a non-local (slow-ish) PG connection**_;
5. Observe your query being overwritten with someone's idea of a sensible default that the user never actually asked for.

**Expected behavior**

If something has already been typed by the time the query that computes the CREATE script comes back, then what has already been typed should be left alone (and the CREATE script thrown away in the trash where it belongs).

(and/or)

Rename "Copy SQL from main window to query tool?" or its uselessly redundant description ("Specifies whether or not to copy SQL to query tool from main window.") to something that corresponds to what's actually going on, so people who want to disable the behavior realize that this is perhaps the option that controls it. 

**Error message**

n/a

**Screenshots**

**Desktop (please complete the following information):**
- OS: macOS-12.6.4-x86_64-i386-64bit
- Version: 7.2
- Mode: Desktop
- Browser (if running in server mode): n/a
- Package type: DMG

**Additional context**

[...]

PR #7522, opened 2024-05-31
Disable the input till the script is loaded.#6357

RohitBhati8269:


web/pgadmin/static/js/components/ReactCodeMirror/components/Editor.jsx
```diff
@@ -193,6 +193,12 @@ export default function Editor({
             }
             if(update.docChanged) {
               onChange?.(update.view.getValue(), update.view);
+              // Update editable properties
+              editor.current?.dispatch({
```
adityatoshniwal, line 197: So you're dispatching an event on every doc change. Not at all good for performance.

web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
```diff
@@ -72,8 +71,15 @@ export default function Query({onTextSelect}) {
   const lastCursorPos = React.useRef();
   const pgAdmin = usePgAdmin();
   const preferencesStore = usePreferences();
-
   const queryToolPref = queryToolCtx.preferences.sqleditor;
+  const [isEditorDisabled, setIsEditorDisabled] = useState(false);
+  const isqueryToolAndUrl = queryToolCtx.params.is_query_tool && queryToolCtx.params.query_url;
```
adityatoshniwal, line 76: this var is not needed.

web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
```diff
@@ -72,8 +71,15 @@ export default function Query({onTextSelect}) {
...
+
+  useEffect ( () => {
+    if (queryToolCtx.preferences?.sqleditor?.copy_sql_to_query_tool || queryToolCtx.params.is_query_tool && queryToolCtx.params.query_url!== '') {
+      setIsEditorDisabled(true);
```
adityatoshniwal, line 80: where is it enabled again? It should enable again once the script it loaded.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
RohitBhati8269: it enables when the doc updated successfully.

web/pgadmin/static/js/components/ReactCodeMirror/components/Editor.jsx
RohitBhati8269: As i try to dispatch that event from another component but not worked.
Will check once again.

adityatoshniwal:
What you could do is - create on state variable in qtState - read_only default to true in QueryToolComponent. First thing what query tool does is call `getSQLScript`. If there is no SQL which is going to load then set read_only to false. Otherwise, set it to false in then of API call after EDITOR_SET_SQL.

web/pgadmin/static/js/components/ReactCodeMirror/components/Editor.jsx
RohitBhati8269: I have changed the approach as per your suggestions.
Thanks

web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -312,27 +312,31 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
       ]
     },
   };
-
-  const getSQLScript = ()=>{
+  
+  const getSQLScript = () => {
+    const setEditorDisabled = (val) => setQtState({ editor_disabled: val });
+    const setSQLAndState = (sql) => {
+      if (sql) eventBus.current.fireEvent(QUERY_TOOL_EVENTS.EDITOR_SET_SQL, sql);
```
adityatoshniwal, line 319: Remove the function setSQLAndState. Call directly.

adityatoshniwal (changes requested):

Issue #7306, opened 2024-03-22
SSH file is not able to read and change when the server is shared on server-mode


nabe1653:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

SSH file is not able to read and change when the server is shared therefore shared server is not abled to connect from other users.

e.g.
1. UserA create "Server1" with SSH file on his storage and share it. 
2. UserB can see "Server1" but it has static SSH connection option, he can not change connection option.
3. UserB faced `No password or public key available!` error when use as it. (maybe due to failed to read ssh file.)

It failed too even if UserA uses secret key from "SHARED_STORAGE" at 1st step. UserB can downalod it from storage-manager but failed to connect.

**To Reproduce**

1. Setup docker server as server-mode

    Create docker container with `config_local.py` and binding volume.

    ```yaml:sample-compose-part.yaml
   ## A part of docker-compose.yaml
    container_name: pgadmin4
    image: dpage/pgadmin4:8.4
    volumes:
      - ./bind/config_local.py:/pgadmin4/config_local.py
      - ./bind/storage:/mnt/storage
    environment:
      - PGADMIN_CONFIG_SERVER_MODE=True
    ```

    ```py:config_local.py
    ## config_local.py
    SHARED_STORAGE = [
      {'name': 'storage', 'path': '/mnt/storage', 'restricted_access': True}
    ]
    ```

2. Change bind folder permission as same as container uid

    ```sh
    ## Check uid on container and set on host (docker uses same id between host and container)
    # docker exec -it "pgadmin4" id
    uid=5050(pgadmin) gid=0(root) groups=0(root)
    # chown 5050:root ./bind/storage
    # chmod 700 ./bind/storage
    ```

3. Create shared user and register ssh server on UserA

    - Create UserB on user manager
    - Upload ssh-key on storage manager
    - Register "Server1" with ssh-tunneling option and uploaded key
    - Share "Server1"

4. Try to connect "Server1" from UserB

    - UserB can not connect "Server1"
    - UserB can input SSH options but can not save it. It become old value when reopen.

**Expected behavior**

1. Allow to use shared-storage path to share secret-key.
2. Allow to use owner's storage when shared ssh key.
3. Let UserB change ssh connection option as his connection info.

Anyway, I wish share ssh-tunneling server info.

**Error message**

`No password or public key available!`

**Screenshots**

**Desktop (please complete the following information):**
 - OS: "Rocky Linux 9.1"
[...]

PR #7535, opened 2024-06-04
Ensure user can connect to servers using SSL certificates, identity files available in the shared storage.#7306

yogeshmahajan-1903:


web/pgadmin/utils/__init__.py
```diff
@@ -226,7 +227,14 @@ def get_complete_file_path(file, validate=True):
     if current_app.PGADMIN_RUNTIME or not current_app.config['SERVER_MODE']:
         return file if os.path.isfile(file) else None
 
-    storage_dir = get_storage_directory()
+    dir_path = ''
+    if re.findall(r':/', file):
```
adityatoshniwal, line 231: increase the specificity of RE by checking if filename starts with a valid shared drive name.

adityatoshniwal (changes requested):
It fixes the SSL file select, but this looks like an issue at all the places. Selecting a shared folder doesn't work at other places. I suggest we should fix this at all the places.

web/pgadmin/utils/__init__.py
yogeshmahajan-1903: Done.

Issue #7414, opened 2024-04-22
Create script does not contain policy comment


chreck:
**Describe the bug**

In the object explorer we select a table and by right click on that table we choose Scripts > CREATE Script. In the detailed view we see the query to create the table and all its details. I.e. also the comments on the columns and the table. But I missing the comments for the defined policies.

**To Reproduce**

1. Open Query Tool
2. Copy this SQLs and execute it
```
CREATE TABLE IF NOT EXISTS public.table_comments_policies
(
	id bigint NOT NULL
)

CREATE POLICY plcy_table_comments_policies
    ON public.table_comments_policies
    AS PERMISSIVE
    FOR ALL
    TO authenticated
    USING (true)
    WITH CHECK (true);

COMMENT ON POLICY "plcy_table_comments_policies" ON "public"."table_comments_policies" IS 'All: auths';
```
3. Refresh tables
4. Right click on table table_comments_policies
6. Choose Scripts > CREATE Script

**Expected behavior**

The expected behavior is that SQLs from the create scripts command contains the `COMMENT ON POLICY` command, but it is not included.

```
-- Table: public.table_comments_policies

-- DROP TABLE IF EXISTS public.table_comments_policies;

CREATE TABLE IF NOT EXISTS public.table_comments_policies
(
    id bigint NOT NULL
)

TABLESPACE pg_default;

ALTER TABLE IF EXISTS public.table_comments_policies
    OWNER to postgres;

GRANT ALL ON TABLE public.table_comments_policies TO anon;

GRANT ALL ON TABLE public.table_comments_policies TO authenticated;

GRANT ALL ON TABLE public.table_comments_policies TO postgres;

GRANT ALL ON TABLE public.table_comments_policies TO service_role;
-- POLICY: plcy_table_comments_policies

-- DROP POLICY IF EXISTS plcy_table_comments_policies ON public.table_comments_policies;

CREATE POLICY plcy_table_comments_policies
    ON public.table_comments_policies
    AS PERMISSIVE
    FOR ALL
    TO authenticated
    USING (true)
    WITH CHECK (true);
```

**Error message**

No error message.

**Screenshots**

No screenshots attached.

**Desktop:**
 - OS: macOS 14.4.1
 - Version: 8.5
 - Mode: Desktop
 - Browser: chrome
[...]

PR #7560, opened 2024-06-11
Add comment field for rls policy dialog.#7414

RohitBhati8269:
Detail Info :- https://github.com/pgadmin-org/pgadmin4/issues/7414


adityatoshniwal (changes requested):
Test cases are missing

Issue #5932, opened 2023-03-10
auto theme based on OS preferences


AbhimanyuAryan:
I found that in the theme section, you can switch to a dark/light theme but there's no option to follow the OS colour scheme. Can we have an automatic colour scheme based on the OS scheme(MacOS)

RohitBhati8269:
Tested and verified on candidate build: [candidate-build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-06-18/)
Environment: MacOs

RohitBhati8269:
Re-tested and verified on candidate build: [candidate-build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-06-24/)
Environment: macOS/Windows

PR #7566, opened 2024-06-11
Provide option to set theme based on OS theme preference.#5932

yogeshmahajan-1903:


web/pgadmin/misc/__init__.py
```diff
@@ -84,11 +83,17 @@ def register_preferences(self):
             options=theme_options,
             control_props={
                 'allowClear': False,
+                'creatable': False,
             },
+        )
+
+        self.preference.register(
```
adityatoshniwal, line 90: there is no need for a new preference. You could add one more drop-down option in existing themes - `System`

web/pgadmin/misc/themes/__init__.py
```diff
@@ -5,9 +5,9 @@
 
 def get_all_themes():
     all_themes = {
-        "standard": {
-            "disp_name": "Standard",
-            "preview_img": "standard_preview.png"
+        "light": {
```
adityatoshniwal, line 8: you'll also have to change the default value for theme preference.

web/pgadmin/preferences/static/js/components/PreferencesComponent.jsx
```diff
@@ -292,6 +292,16 @@ export default function PreferencesComponent({ ...props }) {
           opt.selected = false;
         }
       });
+
+      let sync_with_os_id = null;
+      subNode.preferences.forEach((_el) => {
```
adityatoshniwal, line 297: you can remove this once you the same `theme` preference.

adityatoshniwal (changes requested):

web/pgadmin/misc/__init__.py
yogeshmahajan-1903: Generally, sync with os option is given separately for theme selection for example - pycharm. Providing option this way will be more noticeable than in dropdown.

web/pgadmin/misc/themes/__init__.py
yogeshmahajan-1903: Done

web/pgadmin/misc/__init__.py
adityatoshniwal: Slack, Chrome has different way of doing it. Even VS code.
The point is - having 2 variables to decide a theme can be more confusing. It should be either one.

image

image

adityatoshniwal: Firefox. Most apps do not have a separate option.

image

yogeshmahajan-1903: Done.

web/pgadmin/preferences/static/js/components/PreferencesComponent.jsx
yogeshmahajan-1903: Done.

Issue #7516, opened 2024-05-30
Configuring preferences.json in container causes AttributeError: 'NoneType' object has no attribute 'append'


justvamp:
**Describe the bug**

Configuring `preferences.json` in container causes `AttributeError: 'NoneType' object has no attribute 'append'`.

**To Reproduce**

1. Remove pgadmin volume in docker
2. Add `preferences.json` in `docker-compose.yml`:
```yml
---
networks:
  mdw: null
secrets:
  preferences.json:
    file: docker/offline_mode/pgadmin/preferences.json
  servers.json:
    file: docker/offline_mode/pgadmin/servers.json
services:
  pgadmin:
    depends_on:
      - postgres
    env_file:
      - docker/pgadmin/pgadmin.dev.env
    image: dpage/pgadmin4:8.7
    networks:
      mdw: null
    ports:
      - '5050:80'
    restart: always
    secrets:
      - source: servers.json
        target: /pgadmin4/servers.json
      - source: preferences.json
        target: /pgadmin4/preferences.json
    volumes:
      - pgadmin:/var/lib/pgadmin
```
3. Content of `preferences.json` could be any. Minimal example:
```json
{
  "preferences": {
    "browser:node:show_node_aggregate": false
  }
}
```
4. Run `docker-compose up -d --build`
5. Get the error

**Expected behavior**

Preferences are taken from `preferences.json`.

**Error message**

While initializing container I get the following error:

```
2024-05-30 17:46:56 SQLite pgAdmin config: /var/lib/pgadmin/pgadmin4.db
2024-05-30 17:46:56 ----------
2024-05-30 17:46:59 Added 1 Server Group(s) and 1 Server(s).
2024-05-30 17:47:01 ╭───────────────────── Traceback (most recent call last) ──────────────────────╮
2024-05-30 17:47:01 │ /pgadmin4/setup.py:70 in wrap                                                │
2024-05-30 17:47:01 │                                                                              │
2024-05-30 17:47:01 │    67 │   │   if kwargs and kwargs.get('sqlite_path') is not None:           │
2024-05-30 17:47:01 │    68 │   │   │   # update the sqlite path                                   │
2024-05-30 17:47:01 │    69 │   │   │   config.SQLITE_PATH = kwargs['sqlite_path']                 │
2024-05-30 17:47:01 │ ❱  70 │   │   return f(*args, **kwargs)                                      │
2024-05-30 17:47:01 │    71 │                                                                      │
[...]

PR #7567, opened 2024-06-11
Ensure preferences can be loaded using preferences.json.#7516

yogeshmahajan-1903:


web/setup.py
```diff
@@ -496,6 +496,8 @@ def set_prefs(username,
                   sqlite_path: Optional[str] = None,
                   ):
         """Set User preferences."""
+        if not pref_options:
```
khushboovashi, line 499: Can this bug be fixed by setting the optional param pref_options to []?
Like` pref_options: Annotated[Optional[List[str]], typer.Argument()] = [],`

yogeshmahajan-1903: Generally function parameters should not have mutable default arguments. Hence I set default value as None.

khushboovashi (approved):

PR #7592, opened 2024-06-18
Reapply "Detect if running in Flatpak; patch pythonPath and pgadminFi…

razzeee:
…le accordingly."

This reverts commit dfaea944f024ab90139654957ff0ec8f6c17eb8c.

As nobody knows why that has been reverted or cared to document it.

### DIFF
--- a/runtime/src/js/misc.js
+++ b/runtime/src/js/misc.js
@@ -26,6 +26,10 @@ const createDir = (dirName) => {
   }
 };
 
+export const insideFlatpak = () => {
+  return platform() === 'linux' && fs.existsSync('/.flatpak-info');
+}
+
 // This function is used to get the python executable path
 // based on the platform. Use this for deployment.
 const getPythonPath = () => {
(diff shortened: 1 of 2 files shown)


akshay-joshi:
> …le accordingly."
> 
> This reverts commit [dfaea94](https://github.com/pgadmin-org/pgadmin4/commit/dfaea944f024ab90139654957ff0ec8f6c17eb8c).
> 
> As nobody knows why that has been reverted or cared to document it.

Hi,

This is not the case as Khushboo already mentioned on the old PR that due to a crash we have reverted.

runtime/src/js/misc.js
```diff
@@ -26,6 +26,10 @@ const createDir = (dirName) => {
   }
 };
 
+export const insideFlatpak = () => {
```
akshay-joshi, line 29: Remove the **export** keyword from here, and add 
`insideFlatpak: insideFlatpak,` into the module.exports = {...} in the same file.

runtime/src/js/pgadmin.js
```diff
@@ -29,6 +29,10 @@ let pythonPath = misc.getPythonPath();
 let pgadminFile = '../web/pgAdmin4.py';
 let configFile = '../web/config.py';
 
+if (insideFlatpak()) {
```
akshay-joshi, line 32: Change this line to `if (misc.insideFlatpak()) {`

akshay-joshi (changes requested):

akshay-joshi (changes requested):
Commit message should be appropriate.

runtime/src/js/misc.js
razzeee: Changed as suggested

Issue #5345, opened 2022-09-20
Issue with download result to file from view/edit data results.


nikhil-mohite:
Platform: macOS Monterey 12.6
PgAdmin version: 6.13

Steps:
1. If we open view/edit for table.(Consider there are few rows already exist)
2. Now Add new row enter required values and click on _Save Data Changes_  button from result set toolbar.
3. Now data is save at server, click on  _Save Result to file_ button to downloaded the file.
4. If we open the file the newly added record is not present in the file.
5. If you re-run the query and click on _Save Result to file_ then newly added record is present in the file.

https://user-images.githubusercontent.com/69510038/191252550-79f4833c-fe3b-4743-9c91-7f2f50492d89.mov

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-07-03/
Package: arm64
Environment: macOs Ventura 13.5.1

khushboovashi:
This issue is fixed only for View data. In the Query tool, it is not possible to re-execute the query to fetch the data, as the last executed query could be any query (save/update/delete).

PR #7629, opened 2024-06-27
Fix issue with missing new added records in download file. #5345

RohitBhati8269:
Detail info:- https://github.com/pgadmin-org/pgadmin4/issues/5345


web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -2098,7 +2098,15 @@ def start_query_download_tool(trans_id):
...
-
+        sql = None
+        # Iterate through CombinedMultiDict to find query.
+        for key, value in data.items():
+            if key == 'query':
+                sql = value
+        if not sql:
+            sql = trans_obj.get_sql(sync_conn)
+        # Re-execute the query to ensure the latest data is included
+        sync_conn.execute_async(sql)
```
adityatoshniwal, line 2109: what if there is no change in the data? In many cases data is not even editable. This is blind execute.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
```diff
@@ -380,12 +381,15 @@ export class ResultSetUtils {
...
   saveData(reqData) {
-    return this.api.post(
+    const response = this.api.post(
       url_for('sqleditor.save', {
         'trans_id': this.transId
       }),
       JSON.stringify(reqData)
     );
+    // Set the commit flag to true
+    this.hasQueryCommitted = true;
```
khushboovashi, line 391: What if data has not been saved in case of any backend error? So, in case of an error too, this logic will fetch the data from the PG server while downloading the data

khushboovashi (changes requested):

Issue #3199, opened 2019-12-24
Copy and paste of row containing bytea column saves bytea as 'binary_data' in ascii (RM #5037)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/5037***
*Originally created by **Jason K.** at 2019-12-24 04:23:53 UTC.*

I apologize if this bug has been reported before but I have not found anything matching variations on the description.

I am using pgadmin 4.16 and have encountered a situation where copying and pasting a row containing a bytea column saves the bytea column as 'binary_data' in ascii bytes instead of the actual byte contents of the bytea column. I first observed this in pgadmin 4.8. The data continues to be displayed as '[binary data]' so it is not obvious until usage of the column data that the copy of the data is corrupt. Unless someone looks at the contents of the column bytes, they might simply think that their data was randomly damaged.

Either it should not be possible to use cut and paste on rows containing bytea columns or the cut/paste should actually copy and paste the bytes.

RohitBhati8269:
Fixed and verified on [snapshot-build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-07-10/) 
OS:- macOS Arm64
Installer: Mac (dmg package)

PR #7656, opened 2024-07-02
Disable copy and paste button in the result grid for the row containing Bytea column. #3199

anilsahoo20:


web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
```diff
@@ -787,7 +787,7 @@ export function ResultSet() {
   };
 
   const fireRowsColsCellChanged = ()=>{
-    eventBus.fireEvent(QUERY_TOOL_EVENTS.SELECTED_ROWS_COLS_CELL_CHANGED, selectedRows.size, selectedColumns.size, selectedRange.current, selectedCell.current?.length);
+    eventBus.fireEvent(QUERY_TOOL_EVENTS.SELECTED_ROWS_COLS_CELL_CHANGED, selectedRows, selectedColumns, selectedRange.current, selectedCell.current, columns);
```
adityatoshniwal, line 790: this is listened at many places, did you check other usage? For instance at one place it is negating directly. This might not work as you're sending actual object instead of size.
```
    eventBus.registerListener(QUERY_TOOL_EVENTS.SELECTED_ROWS_COLS_CELL_CHANGED, (rows, cols, range, cell)=>{
      setDisableButton('delete-rows', !rows);
      setDisableButton('copy-rows', (!rows && !cols && !cell && !range));
    });
```

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSetToolbar.jsx
```diff
@@ -142,13 +143,30 @@ export function ResultSetToolbar({query,canEdit, totalRowCount}) {
     });
   }, []);
 
+  const byteaDataTypeCellSelected = (rows, cols, range, cell, columns) =>{
```
adityatoshniwal, line 146: Why do you want to make things so complex? This can make things slower.
You can simply change the bytea editor which will never open.

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSetToolbar.jsx
```diff
@@ -218,7 +236,7 @@ export function ResultSetToolbar({query,canEdit, totalRowCount}) {
           <PgIconButton title={gettext('Copy options')} icon={<KeyboardArrowDownIcon />} splitButton
             name="menu-copyheader" ref={copyMenuRef} onClick={openMenu} />
           <PgIconButton title={gettext('Paste')} icon={<PasteIcon />}
-            shortcut={queryToolPref.btn_paste_row} disabled={!canEdit} onClick={pasteRows} />
+            shortcut={queryToolPref.btn_paste_row} disabled={buttonsDisabled['paste-rows']} onClick={pasteRows} />
```
adityatoshniwal, line 239: do not disable paste rows, it can come from any source outside pgadmin.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSetToolbar.jsx
anilsahoo20: The bytea editor is not editable so it does not open, but this function will disable the copy button, as bytea data type must be know before disabling copy button, so passing the values.

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
anilsahoo20: Done

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSetToolbar.jsx
anilsahoo20: Done

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSetToolbar.jsx
adityatoshniwal: We already had discussed that we can allow copying. There is no harm in allowing. We don't want to affect the performance just to disallow copy.

anilsahoo20: Done

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSetToolbar.jsx
anilsahoo20: Done

adityatoshniwal (approved):

Issue #7638, opened 2024-06-27
Schema Diff: Generate script button remains enabled even if not object is selected.


pravesh-sharma:
**To Reproduce**

1. Open schema diff and compare two DBs or schemas
2. Click on the check box to select any object, generate script button gets enabled.
3. Change schema or click on the compare button again, generate script is still enabled however no object is selected

**Expected behavior**
The generate script button should be disabled.

**Desktop (please complete the following information):**
 - OS: macOS Sonoma 14.5
 - Version: 8.9
 - Mode: Desktop

yogeshmahajan-1903:
Working fine.

PR #7667, opened 2024-07-04
fix disable GenerateScript button when no objects are selected in schema diff. #7638

RohitBhati8269:
Detail Info:- https://github.com/pgadmin-org/pgadmin4/issues/7638

### DIFF
--- a/web/pgadmin/tools/schema_diff/static/js/components/InputComponent.jsx
+++ b/web/pgadmin/tools/schema_diff/static/js/components/InputComponent.jsx
@@ -17,7 +17,7 @@ import { SchemaDiffEventsContext } from './SchemaDiffComponent';
 import { SCHEMA_DIFF_EVENT } from '../SchemaDiffConstants';
 
 
-export function InputComponent({ label, serverList, databaseList, schemaList, diff_type, selectedSid = null, selectedDid=null, selectedScid=null }) {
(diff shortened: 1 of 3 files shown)


web/pgadmin/tools/schema_diff/static/js/components/InputComponent.jsx
```diff
@@ -17,7 +17,7 @@ import { SchemaDiffEventsContext } from './SchemaDiffComponent';
 import { SCHEMA_DIFF_EVENT } from '../SchemaDiffConstants';
 
 
-export function InputComponent({ label, serverList, databaseList, schemaList, diff_type, selectedSid = null, selectedDid=null, selectedScid=null }) {
+export function InputComponent({ label, serverList, databaseList, schemaList, diff_type, selectedSid = null, selectedDid=null, selectedScid=null, onUpdateSelectedRowIds }) {
```
adityatoshniwal, line 20: onUpdateSelectedRowIds - the name itself is incorrect. InputComponent has no relation with selected row ids. It should be `onServerSchemaChange`. `onServerSchemaChange` should be called directly - `onServerSchemaChange()`. The caller of component should implement what to be done in `onServerSchemaChange`. Don't perform the business logic of one component in another.

adityatoshniwal (changes requested):

PR #7669, opened 2024-07-08
Introduction of custom React Hook 'useSchemaState(...)'.

asheshv:
Current implementation of the schema data & state management using the 'stateUtils' is very complicated, and added a lot of complexity.

Introducing a React custom hook `useSchemaState(...)`, which will allow us to manage the schema data & state management outside the React components, and make the code more readable and testable.

Also - it will be able to reuse the schema data & state management implementation outside the SchemaDialogView component.

TODO::
* Add testcases for the following the functionalities:
- [ ] `sessDataReducer`
- [ ] `useSchemaState`
- [ ] `prepareData`
- [ ] `validateSchema`
- [ ]  `validateCollectionSchema`


adityatoshniwal:
I also felt to simplify it further, but priorities you know.
Nevertheless, what I had in mind was using zustand to create a store for every schemview instance - https://github.com/pmndrs/zustand/blob/main/docs/guides/initialize-state-with-props.md#wrapping-the-context-provider. That way it would reduce the re-renders significantly compared to the current context based plus prop drilling. Plus every subcomponent can access state directly using zustand hooks.

Looking forward for the PR. Thanks.

asheshv:
> I also felt to simplify it further, but priorities you know.

I would like to simplify as much as possible.
> Nevertheless, what I had in mind was using zustand to create a store for every schemview instance - https://github.com/pmndrs/zustand/blob/main/docs/guides/initialize-state-with-props.md#wrapping-the-context-provider. 

Thanks for the reference.
> That way it would reduce the re-renders significantly compared to the current context based plus prop drilling. Plus every subcomponent can access state directly using zustand hooks.

I am definitely missing something, and it's calling the custom hook too many times (more time than I anticipated).
I am still a noob in React.

> 
> Looking forward for the PR. Thanks.

Thanks!

asheshv:
I've checked all feature tests and javascript tests are running fine on my local machine.

web/pgadmin/static/js/SchemaView/styled.jsx
```diff
@@ -0,0 +1,170 @@
+/////////////////////////////////////////////////////////////
```
adityatoshniwal, line 1: Please rename this to `Styled` or `StyledComponents` since it includes react components.

web/pgadmin/static/js/SchemaView/utils.js
```diff
@@ -0,0 +1,167 @@
+/////////////////////////////////////////////////////////////
```
adityatoshniwal, line 1: I feel it should be rename to `common.js` since it has more common codes than utils.

adityatoshniwal (changes requested):
Changes looks good to me.

adityatoshniwal:
the feature tests are failing with below screenshot. Looks like hostname is not filled:

image

web/pgadmin/static/js/SchemaView/styled.jsx
asheshv: Done.

web/pgadmin/static/js/SchemaView/utils.js
asheshv: Done.

asheshv:
> the feature tests are failing with below screenshot. Looks like hostname is not filled:
> 
> image

Added a timeout after clicking on the 'Connection' tab for it to be visible the panel itself properly.

adityatoshniwal:
> > the feature tests are failing with below screenshot. Looks like hostname is not filled:

> Added a timeout after clicking on the 'Connection' tab for it to be visible the panel itself properly.

It is still failing with same reason. You can check the screenshots in the artifacts here https://github.com/pgadmin-org/pgadmin4/actions/runs/10160132703

asheshv:
> > > the feature tests are failing with below screenshot. Looks like hostname is not filled:
> 
> > Added a timeout after clicking on the 'Connection' tab for it to be visible the panel itself properly.
> 
> It is still failing with same reason. You can check the screenshots in the artifacts here https://github.com/pgadmin-org/pgadmin4/actions/runs/10160132703

Yeah - I saw that. :(

asheshv:
Thank you @yogeshmahajan-1903 for the pointer.
I am able to resolve the feature tests related the server `host` input based on your input.

But - there are still other cases failing.

asheshv:
So - finally feature tests are passing, but - javascript testcases are failing as 'Reset' button is not disabled, when it is supposed to be. Need some changes to take care of it.

asheshv:
@adityatoshniwal All testcases are passing now. (one github action for feature test failed, but - to upload the test logs, and not an actual issue).

If you're satisfied, please commit the changes.

PR #7691, opened 2024-07-15
Update MfaValidatePage.jsx

jcodeio:
MFA field is set to password type, this triggers browsers to ask to save the password and I find this annoying since its a OTP


khushboovashi:
> MFA field is set to password type, this triggers browsers to ask to save the password and I find this annoying since its a OTP

OTP field should be password type. I don't think we should remove the type password.

jcodeio:
> > MFA field is set to password type, this triggers browsers to ask to save the password and I find this annoying since its a OTP
> 
> OTP field should be password type. I don't think we should remove the type password.

For what reason? these are not protected since they change, it causes the browsers to misbehave by auto filling and attempting to save. It is standard practice to show these plain text, I know of no body else who protects them.

dpage:
I agree - I don't see any reason for the OTP field to be a password.

Issue #7530, opened 2024-06-03
Smart Select option in pgAdmin 4


SheikhAziz:
Smart Select option which was introduced in pgAdmin 4 v8.4 was removed in next released pgAdmin v8.5, I was thinking that it will be back in next release but it doesn't. I don't know why that useful feature is removed from the next release of pgAdmin.
Smart select feature which is provided in most of query editor like visual studio code etc. It is like when you select a text, it will highlights all the matches. For example in pgAdmin 4 v8.4 when you select a text, it highlights all matches in the whole text in green.
Just because of this feature I'm still using v8.4. I installed latest version but downgrade it because of this feature.
I'm sharing screenshots of query example from pgAdmin 4 v8.4.

![highlight](https://github.com/pgadmin-org/pgadmin4/assets/99309995/625fd67c-9df6-4b2a-be13-65b07d7d3b1e)

Is it possible to get back this feature again in future release ?

anilsahoo20:
Note: We could add preference based Smart select.

SheikhAziz:
I think this feature is only in pgAdmin 8.4 and its removed from future release versions, because I never seen this functionality in older version and newer version except v8.4.
Reverted to v8.4 but there are some bugs in v8.4 (like :- find and replace is not working etc.)
Expecting this feature in future release.

Thank you. :)

anilsahoo20:
Feature implemented and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-07-19/
Environment: OSX ventura 13.5.1
Pacakge: arm64

SheikhAziz:
Thank you very much. 🥇

PR #7698, opened 2024-07-16
Add highlight matched selection option in query tool preference.#7530

RohitBhati8269:
Detail Info:- https://github.com/pgadmin-org/pgadmin4/issues/7530


web/pgadmin/browser/static/js/browser.js
```diff
@@ -1728,6 +1728,7 @@ define('pgadmin.browser', [
       insert_pair_brackets: pgBrowser.utils.insertPairBrackets,
       brace_matching: pgBrowser.utils.braceMatching,
       indent_with_tabs: pgBrowser.utils.is_indent_with_tabs,
+      highlight_matched_selection:pgBrowser.utils.highlightSelectedMatchedText
```
adityatoshniwal, line 1731: the left label and right label should be same except underscore

web/pgadmin/tools/sqleditor/utils/query_tool_preferences.py
```diff
@@ -208,6 +208,16 @@ def register_query_tool_preferences(self):
         )
     )
 
+    self.highlight_matched_selection = self.preference.register(
+        'Editor', 'highlight_matched_selection',
```
adityatoshniwal, line 212: again.. as per the label variable name should be `highlight_selection_matches`. Don't multiple names for same things. Keep the name consistent everywhere.

adityatoshniwal (changes requested):
Make sure the highlighted color is visible clearly on all themes.

Issue #6572, opened 2023-07-14
Please add a keyboard shortcut (Ctrl + W) for closing panel/window


s101d1:
Please add a keyboard shortcut (e.g. Ctrl + W) to close panel/window, especially the Query Tool panel.

stefansundin:
I'm on macOS and for me, Cmd+W does the same thing as Cmd+Q, i.e. quit whole the application. I expect it to work like a web browser and only close the "tab" (guess it is called a panel in pgAdmin).

So I agree with this feature request. 👍

s101d1:
@RohitBhati8269 Maybe can just assign the close panel with another shortcut key (like Ctrl + Shift + W) and make the shortcut key customizable?

RohitBhati8269:
Sure @s101d1  we can do it with other shortcut keys that are not default used by the browser and which can be overridden.

pravesh-sharma:
@RohitBhati8269  Everything seems to be working fine there is one one observation

Whenever I close a tab the focus shifts to object explorer it should either shift to the next tab or the first one.

pravesh-sharma:
Found an issue-

When writing a query in the query tool and attempting to close it using shortcut, the focus shifts to another tab. However, since there is a prompt dialog within the query tool for closing, the focus should remain on the query tool itself

pravesh-sharma:
Tested this feature on candidate [build](https://developer.pgadmin.org/builds/2024-07-23-1/) and is working as expected.

Env:
OS - Rocky 8
Mode - Desktop and server

vishalnandagopal:
Every single browser has set Ctrl + W to close a tab and Ctrl + Shift + W to close the whole browser window.

pgAdmin is doing the opposite @pravesh-sharma ?

pravesh-sharma:
@vishalnandagopal 

We are using the shortcuts mentioned to close the panel/tab (such as the Query Tool and ERD Tool) in pgAdmin. Please refer to the comment below.

> I tried to override the Ctrl + W / cmd + w shortcut but did not succeed. I did some research and found that Ctrl + W and some other shortcuts cannot be overridden due to security and user experience considerations.
> 
> References:
> 
[...]

PR #7715, opened 2024-07-18
Fix tab focus issue while deleting multiple tabs using shortcut key.

RohitBhati8269:
DetailInfo:-https://github.com/pgadmin-org/pgadmin4/issues/6572


adityatoshniwal (changes requested):
I just tested existing as well.

PR #7734, opened 2024-07-23
Fix focus issue when closing query tool with shortcut.#6572

RohitBhati8269:
DetailInfo:-https://github.com/pgadmin-org/pgadmin4/issues/6572

### DIFF
--- a/web/pgadmin/browser/static/js/keyboard.js
+++ b/web/pgadmin/browser/static/js/keyboard.js
@@ -188,15 +188,14 @@ _.extend(pgBrowser.keyboardNavigation, {
     }
   },
   _focusTab: function(dockLayoutTabs, activeTabIdx, shortcut_obj, combo){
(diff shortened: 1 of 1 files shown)


web/pgadmin/browser/static/js/keyboard.js
```diff
@@ -193,8 +193,10 @@ _.extend(pgBrowser.keyboardNavigation, {
     else if (combo.key === shortcut_obj.close_tab_panel) {
       const panelId = dockLayoutTabs[activeTabIdx].id?.slice(14);
       if (panelId) {
-        pgAdmin.Browser.docker.close(panelId);
-        activeTabIdx = activeTabIdx === dockLayoutTabs.length - 1 || activeTabIdx === 0 ? 1 : activeTabIdx + 1;
+        const warningPrompt = pgAdmin.Browser.docker.close(panelId);
```
adityatoshniwal, line 196: This is not the right way to do this, this is a hack. There other many tools in pgAdmin which closes with a warning. Putting `warningPrompt` variable all over the places is preposterous.

adityatoshniwal (changes requested):

adityatoshniwal:
This is fixed in an other commit.

PR #7738, opened 2024-07-25
Make sure apparmor profile is added only for Unbuntu 24 as it is only…

adityatoshniwal:
… one creating problem.w

### DIFF
--- a/pkg/debian/build.sh
+++ b/pkg/debian/build.sh
@@ -68,9 +68,29 @@ fakeroot dpkg-deb --build "${SERVERROOT}" "${DISTROOT}/${APP_NAME}-server_${APP_
 echo "Creating the desktop package..."
 mkdir "${DESKTOPROOT}/DEBIAN"
 
-cat << EOF > "${DESKTOPROOT}/DEBIAN/conffiles"
+# Ubuntu 24 requires apparmor profile to work.
+OS_ID=$(grep "^ID=" /etc/os-release | awk -F "=" '{ print $2 }')
+OS_VERSION=$(grep "^VERSION_ID=" /etc/os-release | awk -F "=" '{ print $2 }' | sed 's/"//g' | awk -F "." '{ print $1 }')
+
+if [ "${OS_ID}" == 'ubuntu' ] && [ "${OS_VERSION}" -ge "24" ]; then
+  cat << EOF > "${DESKTOPROOT}/DEBIAN/conffiles"
 /etc/apparmor.d/pgadmin4
 EOF
+
+  mkdir -p "${DESKTOPROOT}/etc/apparmor.d"
+  cp "${SOURCEDIR}/pkg/debian/pgadmin4-aa-profile" "${DESKTOPROOT}/etc/apparmor.d/pgadmin4"
+
+  cat << EOF > "${DESKTOPROOT}/DEBIAN/postinst"
+#!/bin/sh
+
+STATUS="$(systemctl is-active apparmor.service)"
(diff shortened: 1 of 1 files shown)


pkg/debian/build.sh
```diff
@@ -68,10 +68,28 @@ fakeroot dpkg-deb --build "${SERVERROOT}" "${DISTROOT}/${APP_NAME}-server_${APP_
 echo "Creating the desktop package..."
 mkdir "${DESKTOPROOT}/DEBIAN"
 
-cat << EOF > "${DESKTOPROOT}/DEBIAN/conffiles"
+# Ubuntu 24 requires apparmor profile to work.
+UBUNTU_CODENAME=$(grep "^UBUNTU_CODENAME=" /etc/os-release | awk -F "=" '{ print $2 }')
+if [ "${UBUNTU_CODENAME}" == 'noble' ]; then
```
dpage, line 73: Hard coded on a single codename? That means we need to change the code with every release. Please change to a numeric comparison on the version number.

pkg/debian/build.sh
```diff
@@ -68,10 +68,28 @@ fakeroot dpkg-deb --build "${SERVERROOT}" "${DISTROOT}/${APP_NAME}-server_${APP_
...
+  mkdir -p "${DESKTOPROOT}/etc/apparmor.d"
+  cp "${SOURCEDIR}/pkg/debian/pgadmin4-aa-profile" "${DESKTOPROOT}/etc/apparmor.d/pgadmin4"
+
+  cat << EOF > "${DESKTOPROOT}/DEBIAN/postinst"
+#!/bin/sh
+
+STATUS="$(systemctl is-active apparmor.service)"
+if [ "${STATUS}" == "active" ]; then
+    echo "Restarting apparmor.service to load pgAdmin profile..."
+    systemctl restart apparmor.service
```
dpage, line 87: Do we need to restart it, or can we just reload it?

web/pgadmin/about/__init__.py
```diff
@@ -123,6 +123,13 @@ def detect_browser(request):
     agent = request.environ.get('HTTP_USER_AGENT')
     os_details = parse(platform.platform()).ua_string
 
+    try:
```
dpage, line 126: How is this related to AppArmor?

pkg/debian/build.sh
dpage, line 73: Hard coded on a single codename? That means we need to change the code with every release. Please change to a numeric comparison on the version number.

dpage (changes requested):

web/pgadmin/about/__init__.py
adityatoshniwal: It is not, I found it helpful to add in the about dialog.

pkg/debian/build.sh
adityatoshniwal: OK I found a better way. This will only load pgadmin4 
```
apparmor_parser -r /etc/apparmor.d/pgadmin4
```

web/pgadmin/about/__init__.py
dpage: It should be in a separate PR then.

dpage (changes requested):

web/pgadmin/about/__init__.py
adityatoshniwal: OK. I'll remove the change.

dpage (approved):

Issue #7540, opened 2024-06-06
Windows 11 issue: Server heartbeat logging error after deleting server


usmanmuzaffar15:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**
Encountered error 'Server heartbeat logging error : Request failed with status code 404' while deleting a server from tree control

Pre req:
     Postgresql 16 is added to the server group.
1. Go to 'tree control and delete existing server'.
2. Wait for almost 20 seconds, an error message from the right bottom corner appears saying:
**'Server heartbeat logging error : Request failed with status code 404'**
I have tried followings also:
1) Delete the server after disconnecting it from context menu ---> I tried it couple of times and did not encounter the error.
2) Delete the server without disconnecting it from context menu --->  It raised the above mentioned error. I tried three to four times and every time it was encountered.

I tried with both with pgadmin 8.7 and 8.8

**Expected behavior**
No error message should be raised if a server is deleted from the server group from tree area

**Error message**
**'Server heartbeat logging error : Request failed with status code 404'**

**Screenshots**
<img width="960" alt="heartbeat" src="https://github.com/pgadmin-org/pgadmin4/assets/149988380/895903d6-77b6-427e-930b-29a40ef30b4b">

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 11]
 - Version: [e.g. 11 Home 22H2]
 - Mode: [e.g Desktop ]
 - Browser (Nill]
 - Package type: [e.g. windows installer.]

**Additional context**

akshay-joshi:
Tested on candidate build by Yogesh

automationSele:
The issue is fixed. Verified on latest build(8.11) from live. 
Thanks

PR #7787, opened 2024-08-06
Fix server heartbeat logging error after deleting server.#7540

RohitBhati8269:
Detail Info:- https://github.com/pgadmin-org/pgadmin4/issues/7540

### DIFF
--- a/web/pgadmin/browser/static/js/heartbeat.js
+++ b/web/pgadmin/browser/static/js/heartbeat.js
@@ -28,7 +28,7 @@ export function send_heartbeat(_server_id, _item) {
       .catch((error) => {
         if (error && error.message == 'Network Error') {
           pgAdmin.Browser.notifier.error(gettext(`pgAdmin server not responding, try to login again: ${error.message || error.response.data.errormsg}`));
-        } else {
+        } else if (!error.response?.data?.data?.includes('Manager not found')) {
           pgAdmin.Browser.notifier.error(gettext(`Server heartbeat logging error: ${error.message || error.response.data.errormsg}`));
         }
         stop_heartbeat(_item);
--- a/web/pgadmin/utils/heartbeat.py
+++ b/web/pgadmin/utils/heartbeat.py
@@ -25,22 +25,28 @@ def log_server_heartbeat(data):
                                                                )
 
     _server_heartbeat = getattr(current_app, '_pgadmin_server_heartbeat', {})
+    session_id = session.sid
 
-    if session.sid not in _server_heartbeat:
-        _server_heartbeat[session.sid] = {}
+    if session_id not in _server_heartbeat:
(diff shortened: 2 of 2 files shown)


web/pgadmin/browser/static/js/heartbeat.js
```diff
@@ -28,6 +28,8 @@ export function send_heartbeat(_server_id, _item) {
       .catch((error) => {
         if (error && error.message == 'Network Error') {
           pgAdmin.Browser.notifier.error(gettext(`pgAdmin server not responding, try to login again: ${error.message || error.response.data.errormsg}`));
+        } else if (error.response.data?.data?.includes('Manager not found')) {
+          console.error(error);
```
khushboovashi, line 32: No need for a console error, just bypass the error.

web/pgadmin/utils/heartbeat.py
```diff
@@ -25,22 +25,28 @@ def log_server_heartbeat(data):
...
-    if session.sid not in _server_heartbeat:
-        _server_heartbeat[session.sid] = {}
+    if session_id not in _server_heartbeat:
+        _server_heartbeat[session_id] = {}
 
     if not manager:
         stop_server_heartbeat(data)
-        return False, gettext("Manager not found. Stopped Heartbeat logging.")
+        msg = gettext("Manager not found. Stopped Heartbeat logging.")
+        current_app.logger.debug(
```
khushboovashi, line 36: It should be current_app.logger.error

khushboovashi (changes requested):

khushboovashi (changes requested):
It showing log error in some cases.
Screenshot 2024-08-06 at 4 06 15 PM

web/pgadmin/browser/static/js/heartbeat.js
```diff
@@ -26,10 +26,12 @@ export function send_heartbeat(_server_id, _item) {
         }
       })
       .catch((error) => {
-        if (error && error.message == 'Network Error') {
-          pgAdmin.Browser.notifier.error(gettext(`pgAdmin server not responding, try to login again: ${error.message || error.response.data.errormsg}`));
-        } else {
-          pgAdmin.Browser.notifier.error(gettext(`Server heartbeat logging error: ${error.message || error.response.data.errormsg}`));
+        if (!error.response?.data?.data?.includes('Manager not found')){
```
khushboovashi, line 29: I would just return from this if condition, rather than putting everything into this condition.

khushboovashi (changes requested):

web/pgadmin/browser/static/js/heartbeat.js
RohitBhati8269: If I directly return  then it will not call the  `stop_heartbeat(_item);`

Issue #7682, opened 2024-07-11
Schema Diff: Generate script Ignores Filter Conditions When Using "Select All"


WeiJie-JL:
**Describe the bug**

After generating the schema diff results in pgAdmin4, the "Select All" button ignores any filter conditions that have been applied, leading to a potentially dangerous situation where objects could be unintentionally modified or dropped. 

**To Reproduce**

1.After the diff results are displayed, uncheck the "target only" option in the top-right corner.
2.Select all items in the displayed results list. At this point, the list does not show items that are "target only."
3.Generate the change script. You will find that the script includes "DROP TABLE xxx CASCADE" statements.

**Expected behavior**

Only the visible selected content is included in the script.

**Desktop (please complete the following information):**
 - OS: both Windows 10 and OSX monterey
 - Version: 8.9
 - Mode: Desktop

**Additional context**

This behavior is completely counterintuitive and can easily lead to irreversible consequences. It is hoped that this issue can be addressed and resolved.

rtrad89:
This also happens when I select all tables, or all functions, etc. so not only on the high-level select all checkbox but also on the subgroups.

RohitBhati8269:
Issue fixed and Verified on snapshot [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-08-12/).

Env:-
OS - macOS arm
Mode - Server

WeiJie-JL:
<img width="733" alt="image" src="https://github.com/user-attachments/assets/8b159942-445e-47bf-a23f-683607b9a070">
the issue can still be reproduced on the latest version 8.10.

akshay-joshi:
@WeiJie-JL 

The fix has not been released yet, if you can check the milestone it is for v8.11.

WeiJie-JL:
> @WeiJie-JL
> 
> The fix has not been released yet, if you can check the milestone it is for v8.11.

Alright, thank you very much.

Issue #7499, opened 2024-05-22
On refreshing schema diff tool opened in new tab gives error 'Transaction not found' error


yogeshmahajan-1903:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

On refreshing schema diff tool opened in new tab gives error 'Transaction not found' error

**To Reproduce**

1. Open schema diff tool in new tab.
2. Refresh the tab.
3. Try to compare 2 databases now.
4. Error will be shown

**Expected behavior**

Error should not be shown.

RohitBhati8269:
Issue fixed and Verified on snapshot [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-08-12/).

Env:-
OS - macOS arm
Mode - Server

PR #7791, opened 2024-08-07
1) Fixed an issue where refreshing the Schema Diff tool opened in a new tab caused an error. #7499

akshay-joshi:
2) Fixed an issue where the Generate Script ignored filter conditions when a parent node was selected. #7682

### DIFF
--- a/web/pgadmin/tools/schema_diff/static/js/components/SchemaDiffComponent.jsx
+++ b/web/pgadmin/tools/schema_diff/static/js/components/SchemaDiffComponent.jsx
@@ -8,15 +8,16 @@
 //////////////////////////////////////////////////////////////
 
 
-import React, { createContext, useMemo, useRef } from 'react';
+import React, { createContext, useMemo, useRef, useEffect } from 'react';
 import { styled } from '@mui/material/styles';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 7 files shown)


web/pgadmin/tools/schema_diff/static/js/components/SchemaDiffComponent.jsx
```diff
@@ -82,6 +92,10 @@ export default function SchemaDiffComponent({params}) {
     });
   }
 
+  useEffect(()=>{
+    initializeSchemaDiff();
```
adityatoshniwal, line 96: I think we should show a spinner or something till initialise in progress.

adityatoshniwal (changes requested):

web/pgadmin/tools/schema_diff/static/js/components/SchemaDiffComponent.jsx
akshay-joshi: Sure will work on it.

akshay-joshi: As per offline discussion, it is not required.

adityatoshniwal (approved):

Issue #7688, opened 2024-07-15
ERD tool not opening the saved pgerd file


anilsahoo20:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

When we open a recently saved pgerd file on a new ERD tool tab with keyboard shortcut(Cmd+o), it gives blank tab.

**To Reproduce**

1. open erd tool
2. create a table
3. save the dialog
4. open a new erd tool panel, use keyboard shortcut to open file manager to open the file.
5. It will not draw the tables
6. Try using the cursor to open the file manager to open the pgerd file, it will work.

**Expected behavior**

It should work.

**Error message**

If possible, include the full, unedited text of any error messages displayed.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: OSX ventura
 - Version: 8.9
 - Mode: Desktop
 - Browser (if running in server mode): [e.g. chrome, safari]
 - Package type: arm64

**Additional context**

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-08-14/
Package: arm64, docker
Environment: macOs Ventura 13.5.1

PR #7804, opened 2024-08-12
Fix an issue where ERD tool not opening the saved pgerd file. #7688

RohitBhati8269:
bug fixed in this PR :-
1. Detail Info:- https://github.com/pgadmin-org/pgadmin4/issues/7688
2. The "Save" button was not disabled after deleting all objects/tables from the ERD tool.
3. The "Save" file shortcut was functional(opens a save file dialogue) even when no objects were created in the ERD tool.

### DIFF
--- a/web/pgadmin/tools/erd/static/js/erd_tool/components/ERDTool.jsx
+++ b/web/pgadmin/tools/erd/static/js/erd_tool/components/ERDTool.jsx
@@ -298,7 +298,7 @@ export default class ERDTool extends React.Component {
   async componentDidMount() {
     this.setLoading(gettext('Preparing...'));
     this.registerEvents();
-
+    this.diagramContainerRef.current?.focus();
     const erdPref = this.preferencesStore.getPreferencesForModule('erd');
     this.setState({
       preferences: erdPref,
@@ -510,6 +510,10 @@ export default class ERDTool extends React.Component {
         this.diagram.getSelectedLinks().forEach((link)=>{
           this.diagram.removeOneToManyLink(link);
         });
+        if (this.diagram.getNodesData().length === 0){
+          this.setState({dirty: false});
+          this.eventBus.fireEvent(ERD_EVENTS.DIRTY, false);  
+        }
         this.diagram.repaint();
       },
       () => {/*This is intentional (SonarQube)*/}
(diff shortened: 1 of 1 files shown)


web/pgadmin/tools/erd/static/js/erd_tool/components/ERDTool.jsx
```diff
@@ -577,7 +581,7 @@ export default class ERDTool extends React.Component {
     this.closeOnSave = closeOnSave;
     if(this.state.current_file && !isSaveAs) {
       this.saveFile(this.state.current_file);
-    } else {
+    } else if (this.diagram.getNodesData().length > 0){ {
```
adityatoshniwal, line 584: same as above. why disable saving empty files.

adityatoshniwal (changes requested):
I failed to find code changes related to commit message.

web/pgadmin/tools/erd/static/js/erd_tool/components/ERDTool.jsx
```diff
@@ -510,6 +510,10 @@ export default class ERDTool extends React.Component {
         this.diagram.getSelectedLinks().forEach((link)=>{
           this.diagram.removeOneToManyLink(link);
         });
+        if (this.diagram.getNodesData().length === 0){
```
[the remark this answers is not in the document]
RohitBhati8269: when we open a new ERD tool, the save button is initially disabled. It becomes enabled when we create objects or tables, but if we delete all the objects we've created, the save button should be disabled again, just like it is in the query tool. Since it wasn’t getting disabled automatically, that's the reason I added this check.

web/pgadmin/tools/erd/static/js/erd_tool/components/ERDTool.jsx
RohitBhati8269: Same for this, As when we open ERD tol our save button is disabled but our shortcut for save is working so It looks like a bug.

RohitBhati8269:
> I failed to find code changes related to commit message.

While debugging I found that the issue occurred due to the wrong focus, so I added a focus on erd-tool, now its working fine.

PR #7854, opened 2024-08-26
replace Flask-Security-Too with Flask-Security

dotlambda:
see https://github.com/pallets-eco/flask-security/blob/main/README.rst#notes-on-this-repo


adityatoshniwal:
@dotlambda As per the link you shared. The package will be deployed with both names Flask-Security and Flask-Security-Too. Only the git repo is changes. So I don't think this change is needed.
Also, there may be conflicts when upgrading pgAdmin as Flask-Security-Too need to be manually removed. pip will not remove on its own.

Thank you for raising the PR anyway.

Issue #7863, opened 2024-08-28
The row number in result grid is deceptive and easy to lead mistake in data operation


xzer:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

The row number in result grid is very annoying, when the table has a number id (usually the PK) as the first column, it is very confusing to user which number is the row number and which number is the id from the table. Especially when the user is scrolling to right side to view the long columns, the row number column is stick at left side, it is very much easy to mislead user to treat the row number as the id of the table data.

I report this as a bug rather than feature request, because this defect has significant impact on data operation correctness.

**To Reproduce**

1. Go to 'query panel'
2. input SQL as "select * from some_table"
3. See result grid

**Expected behavior**

The row number should be optional, there should be toggle to allow hide the row number column, and also there should be a global setting to allow hide the row number column.

**Error message**

N/A

**Screenshots**

<img width="1120" alt="image" src="https://github.com/user-attachments/assets/de28dd03-e0fe-44f0-80cb-b832052df7f4">

<img width="1111" alt="image" src="https://github.com/user-attachments/assets/dc4ce906-b0e4-4a17-b1d5-cd0c1835a45f">

**Desktop (please complete the following information):**
 - OS: OSX Sonoma
 - Version: 8.9
 - Mode: Desktop or Server

**Additional context**

xzer:
I report this as a bug rather than feature request, because this defect has significant impact on data operation correctness.

adityatoshniwal:
@xzer You can check any sql editor, all have row numbers because it is very useful in most cases. Your request is first of a kind, no one had a problem with this before. So we're taking it as feature request to allow disabling the rownumbers.

xzer:
check the dbeaver

https://dbeaver.io/wp-content/uploads/2018/03/data_edit-450x220.png

Yes, it also has row number in the result grid, but it is in clear different cell style and background color, it is very clearly indicate the number is not a part of the data. But look above picture I pasted, they are almost same.

I would say, it may be not necessary to really "hide" it, but it need to be clearly differentiated from the data cells.

xzer:
> Right now we add slight shadow to the right edge of row number.

If you go back to check my pasted screenshot, I don't think the "slight shadow" can be easily differentiate from other cell border lines.

[...]

Issue #2874, opened 2019-08-09
Alternate the background color of rows to enhance readability (RM #4598)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/4598***
*Originally created by **Luther Goh** at 2019-08-09 14:44:38 UTC.*

Table rows' background colour can be alternated in order to make them more readable.

It may be achieved by just adding css as suggested in this message:

https://www.postgresql.org/message-id/CAAj2gHwUrYz7t2nzwZgZXeTj5fbEtkACQdHWR2KEsd2_T4gU7g%40mail.gmail.com

pravesh-sharma:
The feature is implemented and is working as expected, verified on snapshot [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-10-03/).

Env:
OS - macOS 14.7 (Sonoma)
Mode - Desktop and server
Browser - Chrome

pravesh-sharma:
Hi @jon777,

We tested on dark and high-contrast themes and found no issues. Texts are readable.

PR #7911, opened 2024-09-11
Alternate row background color and make the serial number cell distinct. #2874 #7863

RohitBhati8269:
Detail Info:-
1. https://github.com/pgadmin-org/pgadmin4/issues/7863
2. https://github.com/pgadmin-org/pgadmin4/issues/2874


adityatoshniwal (changes requested):
It is actually the same as existing with only difference is alternate row backgrounds. If later we remove the alternate bg using preferences then issue of distinguishing row numbers will remain un-fixed.

PR #7927, opened 2024-09-16
Fixed the import-export 'Not null columns' isuse and UserManagement dialog issue (#7884)

asheshv:


anilsahoo20:
On import/export data for table, when we open the dialog for the 1st time, it does not shows the columns list on NOT NULL columns control, once we close the dialog and opens the dialog for the same table for the 2nd time, it shows the list of columns on NOT NULL columns control.

PR #7929, opened 2024-09-16
Introduced a 'exclude' option in the 'Field' to exclude it from the change. (Fixes #7884)

### LINKED ISSUE #7884: Single responsibility principle for the SchemaView and DataGridView
> A class should have one, and only one, reason to change.
> -- Robert C. Martin (Uncle Bob)

**`Single responsibility principle`** improves the code maintainability, because:
* **Frequency and Effects of Changes** - you need to change block for code for a single purpose.
* **Easier to Understand** - improves the readability
* **Easy to extend** - easy to add new features

Current implementation of the `SchemaView` and `DataGridView` is amazing. It is great piece of code, and does the job well.
But - at the same time, it's not easy to understand, and it is not easy to make change in it, as it could result into breaking other places as well.

When working on the 'useSchemaState' custom hook, I realised that - it is not easy or enough to move some piece of the code and introduce it as the custom hook, the current javascript and feature test-suites have definitely informed me about the broken state at a particular point of time. it made me realised that it won't be easy to introduce new features or do modification in future.

As a result - I would like to propose to use the single responsibility principle for them.

Note:  There are still some exceptions in the currently [proposed patch](https://github.com/pgadmin-org/pgadmin4/pull/7876).

References:
- https://blog.cleancoder.com/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html
- https://stackify.com/solid-design-principles/

asheshv:
Use the option 'exclude' to add field 'notNullColumns', which will be excluded from the data, in the 'Import/Export' UI Schema. It will help to force re-render the 'Not Null Columns' select control on change of 'notNullColumns' data availability.

### DIFF
--- a/web/pgadmin/static/js/SchemaView/SchemaState/common.js
+++ b/web/pgadmin/static/js/SchemaView/SchemaState/common.js
@@ -149,6 +149,9 @@ export function getSchemaDataDiff(
     };
 
     schema.fields.forEach((field) => {
+      // Never include data from the field in the changes, marked as
+      // 'excluded'.
+      if (field.exclude) return;
       /*
        * If skipChange is true, then field will not be considered for changed
        * data. This is helpful when 'Save' or 'Reset' should not be enabled on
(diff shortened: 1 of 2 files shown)


pravesh-sharma:
Tested locally and it is working as expected.

PR #7941, opened 2024-09-18
Fixing an issue introduced due to reloading of the tree due to changes in particular preferences.

asheshv:
Fix an issue - when object browser tree is being recreated due to reloading for changes in some of the preferences. Tree object returns object from the previous instance as 'selected', but - it does not have the 'treeNodeInfo' available. In this special case - we would consider that there is no node selected at that particular moment, and pass information accordingly.


asheshv:
Will create a different branch for the same.

Issue #7958, opened 2024-09-23
User management dialog search is not working.


anilsahoo20:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

1. The search box takes the full width, but it has taken half width.
2. The search box is not giving any result on searching the users that are still present.
3. Note is missing in Existing table>Edit dialog>Partitions tab.

**Screenshots**
<img width="899" alt="Screenshot 2024-09-19 at 5 25 55 PM" src="https://github.com/user-attachments/assets/6d544743-29fb-492e-9a70-e1484a98bdc1">

<img width="901" alt="Screenshot 2024-09-23 at 2 33 55 PM" src="https://github.com/user-attachments/assets/b2ae3fcf-5f2f-4661-8259-47fb18593184">

anilsahoo20:
#7955, the issue will be fixed along with this issue.

yogeshmahajan-1903:
1.Scroll is added in the parameters - [Sev-3]
<img width="720" alt="Screenshot 2024-09-20 at 9 53 48 AM" src="https://github.com/user-attachments/assets/72de1a99-6cbf-447e-8e62-ae26470846c8">

Expected - 
<img width="771" alt="Screenshot 2024-09-20 at 9 53 53 AM" src="https://github.com/user-attachments/assets/f3eaedb1-4edf-44d5-b739-9fbe1077858a">

asheshv:
[> #7955, the issue will be fixed along with this issue.
](https://github.com/pgadmin-org/pgadmin4/pull/7959)
This has not taken care of the Dashboard issue as it is not really related to SchemaView changes.

pravesh-sharma:
Issue fixed verified on snapshot [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-10-03/).

Env:
OS - macOS 14.7 (Sonoma)
Mode - Desktop.

PR #7959, opened 2024-09-23
Fixed issues reported in #7958

asheshv:
- Search wasn't working as we were using wrong row-index, while rendering it.
- Fixed the CSS issue with the alignment of the search input box.
- Pass the row object while rendering the 'DataRow' component, to improve the performance as it was calling the filtering function repeatedly for each row.
- Changed the order of the validators on a field to priortize the 'empty' string validator over the unique-id validator in a collection.

### DIFF
--- a/web/pgadmin/static/js/SchemaView/DataGridView/formHeader.jsx
+++ b/web/pgadmin/static/js/SchemaView/DataGridView/formHeader.jsx
@@ -12,6 +12,7 @@ import React, {
 } from 'react';
 import { styled } from '@mui/material/styles';
 import Box from '@mui/material/Box';
+import PropTypes from 'prop-types';
 
 import { SCHEMA_STATE_ACTIONS } from 'sources/SchemaView/SchemaState';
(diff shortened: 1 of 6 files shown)


anilsahoo20:
The search box result is wrong.

Expected search box width
Screenshot 2024-09-23 at 3 46 49 PM

Current search box width
Screenshot 2024-09-23 at 3 46 36 PM

asheshv:
> The search box result is wrong.
> 
> Expected search box width Screenshot 2024-09-23 at 3 46 49 PM
> 
> Current search box width Screenshot 2024-09-23 at 3 46 36 PM

It is not a side effect or bug, but - It has been changed like that intentionally.

anilsahoo20:
The partition tab in edit dialog of table is not showing.
Screenshot 2024-09-23 at 4 07 51 PM

anilsahoo20:
The search box result is correctly displayed.

One remaining issue still not fixed i.e. `Note is missing in Existing table>Edit dialog>Partitions tab.`

Issue #2646, opened 2019-06-05
See (and set) encrypted password (RM #4323)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/4323***
*Originally created by **Michael Shapiro** at 2019-06-05 05:51:33 UTC.*

PgAdmin3 has the ability to see the encrypted password for a login role. Can this be added to PgAdmin4?

When you select the SQL for a login role, PgAdmin3 shows.

  CREATE ROLE xxx_user LOGIN
    ENCRYPTED PASSWORD 'md5f349771b7d83b170e8377bd7d09'
    SUPERUSER INHERIT CREATEDB CREATEROLE NOREPLICATION;

Can this functionality be added to PgAdmin4? 
Also, can the dialog for a login role support setting an encrypted password?

khushboovashi:
An encrypted password for the login role is implemented in the SQL tab.

khushboovashi:
https://www.postgresql.org/docs/15/sql-createrole.html

akshay-joshi:
As per the documentation from PostgreSQL v10 and above "The password is always stored encrypted in the system catalogs. The ENCRYPTED keyword has no effect, but is accepted for backwards compatibility."

Hence closing this feature.

PR #7972, opened 2024-09-25
Add an encrypted password for create and edit roles.#2646

RohitBhati8269:
Detail Info:- https://github.com/pgadmin-org/pgadmin4/issues/2646


adityatoshniwal (changes requested):
Checking if the password starts with `scram-sha-256$` or `md5` is absurd. One can keep a password like `md54321`. You're supposed to add one more separate text box for encrypted password.

akshay-joshi:
The password is always stored encrypted in the system catalogs. The ENCRYPTED keyword has no effect, but is accepted for backwards compatibility.

We don't required it, so closing this PR.

Issue #7393, opened 2024-04-18
Query Window - SQL Text Newline is copied to the ClipBoard as LF (and not as CRLF)


durumdara:
As I experienced that if I copy the Query SQL text to the ClipBoard, that NewLine sign is only LF.

On Windows OS the NewLine is CRLF.

Some tools can handle the orphan CR or LF, like Notepad++.

But if I paste the SQL text into DevExpress controls like cxMemo (= dialogs written in Delphi),  the whole SQL is only one line!

It would be better if in Win64 compilation you use CRLF as NewLine and not only LF.

I hope that your GUI or Dev. Package allows this.

Matestro92:
This makes my work very difficult. Will the report be processed?

pravesh-sharma:
Found an issue:

On windows default should be CRLF but editor takes LF as default.

pravesh-sharma:
Working fine. Tested on snapshot [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-10-04/).

Env:
OS - Windows 11
Mode - Desktop

PR #8001, opened 2024-10-03
Fix EOL detection based on OS.

RohitBhati8269:
Detail info:- https://github.com/pgadmin-org/pgadmin4/issues/7393


adityatoshniwal (changes requested):
`navigator.platform` is deprecated.

Issue #7907, opened 2024-09-09
Support system option for sslrootcert


dgilman-hrp:
**Describe the solution you'd like**

postgres v16 and on supports using `system` for `sslrootcert` where the system root certificate storage will be used instead of setting the path to a file.

**Describe alternatives you've considered**

The workaround is to set the path manually, which is workable, but less secure and not the desired behavior.

**Additional context**

[Postgres docs here](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT)

yogeshmahajan-1903:
@dgilman-hrp 
You can set the value for Root certificate by manually typing `system`. 
<img width="697" alt="Screenshot 2024-09-20 at 5 07 30 PM" src="https://github.com/user-attachments/assets/0cb59e4b-198d-4a8f-8335-6bc6acd4c79a">

willluongo-vendia:
I am running pgadmin 8.12 and it does not seem to honor the "system" value, I get this error:
> Unable to connect to server: connection failed: connection to server at "<ip>", port 5432 failed: root certificate file "/Users/<user>/.postgresql/root.crt" does not exist Either provide the file, use the system's trusted roots with sslrootcert=system, or change sslmode to disable server certificate verification.

from these settings:
<img width="698" alt="image" src="https://github.com/user-attachments/assets/4421bcd3-52c7-4068-8764-6f6e5f2f9511">
 



However, I can provide the path from this line in pathAbout PgAdmin 4: `CA_FILE = "/Applications/pgAdmin 4.app/Contents/Resources/web/cacert.pem"`. The connection also works from psql using the `sslrootcert=system` option.

yogeshmahajan-1903:
@willluongo-vendia 

Does file - /Users/<user_name>/.postgresql/root.crt exists on you system? If yes can you please list permission for the same?
Also does pgadmin is installed with the same user which is being used to create psql  connection?

When I rename root certificate, at ~/.postgresql directory  I get the error
$ .postgresql % sudo mv root.crt root123.crt                                                                                                            
$ .postgresql % psql "host=< ip_addr > port=<port_number> dbname=postgres user=postgres connect_timeout=10 sslmode=verify-full"
psql: error: connection to server at "<ip_addr>, port< port_number > failed: root certificate file "/Users/<user_name>/.postgresql/root.crt" does not exist
Either provide the file or change sslmode to disable server certificate verification.

[...]

PR #8035, opened 2024-10-14
Ensure server connection is successful for sslrootcert=system in server configuration.#7907

yogeshmahajan-1903:


web/pgadmin/utils/driver/psycopg3/server_manager.py
```diff
@@ -675,6 +675,12 @@ def create_connection_string(self, database, user, password=None):
                     with_complete_path = True
                     value = get_complete_file_path(value)
 
+                if key == 'sslrootcert':
```
adityatoshniwal, line 678: You can remove sslrootcert from `['passfile', 'sslcert', 'sslkey', 'sslrootcert', 'sslcrl', 'sslcrldir']` since it is taken care separately.

web/pgadmin/utils/driver/psycopg3/server_manager.py
```diff
@@ -675,6 +675,12 @@ def create_connection_string(self, database, user, password=None):
...
+                    if value == 'system':
```
adityatoshniwal, line 679: can be reduced to 1 if condition `if value != 'system'` to avoid sonarqube issues.

adityatoshniwal (changes requested):

Issue #1984, opened 2018-08-01
Add menu option to close all server connections (RM #3533)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/3533***
*Originally created by **Michael Krueger** at 2018-08-01 05:47:42 UTC.*

I stumbled upon the case, where I tried to connect to one server in my servergroup, using a user that had very limited permissions to the point he was not able to query the db objects tree. PGAdmin4 generated an internal error, and I thought it to be wise that I restart PGAdmin backend. After restart I opened a new browser window, and PGAdmin (due to saved sessions) tried to restore my previously failing db connection. That in turn prohibited me from disconnecting that server manually, as I was not able to reach that server in the group. Cleaning the local session cache under .pgadmin/ I was able to get PGAdmin back to work.

To not have to go to the disk to remove the sessions, I recommend a menu option that would stop PGAdmin4 from attempting to connect to any previously connected server. Something like "Close all server connections" should to the trick. Then I expect to be able to open the server group again, to reach the failing server, to be able to change its properties.

anilsahoo20:
Tested and verified on source code.
Environment: macOs 13.5.1

PR #8080, opened 2024-11-04
Added a menu option to close all server and database connections. #1984

pravesh-sharma:


anilsahoo20 (commented):
1. After multiple database connections opened, if I click on `Disconnect from all databases` menu option, it disconnects all the databases but the Menu option must be deactivated, but currently it is still active.
2. After clicking the `Disconnect from all servers`, the focus on Servers group is going away. 
3. After clicking the `Disconnect from all servers`, if we see the object menu, it show many options and we click on any of the option, it is not working.

pravesh-sharma:
> 1. After multiple database connections opened, if I click on `Disconnect from all databases` menu option, it disconnects all the databases but the Menu option must be deactivated, but currently it is still active.
> 2. After clicking the `Disconnect from all servers`, the focus on Servers group is going away.
> 3. After clicking the `Disconnect from all servers`, if we see the context menu, it show many options and we click on any of the option, it is not working.

#2 is expected behaviour.

adityatoshniwal (changes requested):

anilsahoo20 (approved):

Issue #5099, opened 2022-05-26
Problem with Query Tool (RM #7422)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/7422***
*Originally created by **Mark Pare** at 2022-05-26 19:37:15 UTC.*

Hello, I just upgraded to v6.9 from a previous release (v5.0) and I have multiple issues in the Query Tool:

Standard notepad shortcuts used to work

CTRL-A no longer selects all data
CTRL-END no longer moves the cursor to the last row / last column
Highlighting cells to be able to issue CTRL-C on doesn't work anymore

In general, navigation in the query tool doesn't work the same anymore. Whenever I execute a query, I need to left-click in the Query window otherwise I'm in the Data Output window automatically, which is different than before.

pravesh-sharma:
Except ctrl/cmd + A i.e. select all other keys seems to work. Changes are required in react-data-grid lib used in the query tool

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2024-12-09/
Package: exe, arm64
Environment: macOs Sonoma 14.7.1 & WIndows 11

PR #8094, opened 2024-11-06
Restore Ctrl+A shortcut to select all rows and columns in results grid.

RohitBhati8269:
Detail Info:- https://github.com/pgadmin-org/pgadmin4/issues/5099


web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/index.jsx
```diff
@@ -92,13 +99,39 @@ const StyledPgReactDataGrid = styled(PgReactDataGrid)(({theme})=>({
...
       handleCopy();
     }
+    if (handleSelectAll && ((e.metaKey && isMac) || (e.ctrlKey && !isMac)) && e.key === 'a') {
+      e.preventDefault();
+      handleSelectAll();
+    }
+  };
+}
+
+function handleShortcutKeyDown({
```
adityatoshniwal, line 115: there should only be one handler. don't make it complex by adding multiple layers of callbacks. I'm finding it very difficult to understand what is the code doing. There are no comments as well.

web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/index.jsx
```diff
@@ -333,6 +390,32 @@ RowNumColFormatter.propTypes = {
   onSelectedColumnsChange: PropTypes.func,
 };
 
+function CustomCell({ row, column, onSelectedColumnsChange}) {
```
adityatoshniwal, line 393: Where are the other props gone which are passed by react-data-grid?

web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/index.jsx
```diff
@@ -344,11 +427,16 @@ function formatColumns(columns, dataChangeStore, selectedColumns, onSelectedColu
 
   for(const [idx, col] of retColumns.entries()) {
     col.renderHeaderCell = HeaderRenderer;
+    col.renderCell = (props)=>{
```
adityatoshniwal, line 430: do we really need this? react-data-grid has `onCellKeydown` prop. We should be able to use that I think.

adityatoshniwal (changes requested):
I think the solution need to be re-visited.

web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/index.jsx
RohitBhati8269: I already tried that, but I wasn't successful.
Encountered issue on selecting all rows with react-data-grid hook `onRowSelectionChange`.

web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/index.jsx
RohitBhati8269: Thanks, I will optimize it.

RohitBhati8269:
> I think the solution need to be re-visited.

I will go through with the react-data-grid doc again. If I find another best solution, I will try to apply that.
Thanks

RohitBhati8269:
> > I think the solution need to be re-visited.
> 
> I will go through with the react-data-grid doc again. If I find another best solution, I will try to apply that. Thanks

Updated the solution.

adityatoshniwal:
I have implemented this in more easy way and have committed here - https://github.com/pgadmin-org/pgadmin4/commit/c5d5fbdafd211bf2f2b3c9fe526de2e232f3f587.
Closing this.

PR #8107, opened 2024-11-07
Not setting the full dialect and driver for `CONFIG_DATABASE_URI` in case of an external DB currently breaks the docker image

sdsys-ch:
When using PostgreSQL for pgAdmin's config database, SQLAlchemy 2.0+ requires the 'psycopg' dialect vs. 'psycopg2' or 'psycopg3':

```
>>> from sqlalchemy.dialects.postgresql import psycopg; help(psycopg)
Help on module sqlalchemy.dialects.postgresql.psycopg in sqlalchemy.dialects.postgresql:

NAME
    sqlalchemy.dialects.postgresql.psycopg

DESCRIPTION
    .. dialect:: postgresql+psycopg
        :name: psycopg (a.k.a. psycopg 3)
        :dbapi: psycopg
        :connectstring: postgresql+psycopg://user:password@host:port/dbname[?key=value&key=value...]
        :url: https://pypi.org/project/psycopg/

    ``psycopg`` is the package and module name for version 3 of the ``psycopg``
    database driver, formerly known as ``psycopg2``.  This driver is different
    enough from its ``psycopg2`` predecessor that SQLAlchemy supports it
    via a totally separate dialect; support for ``psycopg2`` is expected to remain
    for as long as that package continues to function for modern Python versions,
    and also remains the default dialect for the ``postgresql://`` dialect
    series.
```

Problem:

When setting `PGADMIN_CONFIG_CONFIG_DATABASE_URI` in Docker env to either `postgresql://...` or `postgresql+psycopg3://...` we either hit `sqlalchemy.exc.NoSuchModuleError: Can't load plugin: sqlalchemy.dialects:postgresql.psycopg3` or `ModuleNotFoundError: No module named 'psycopg2'`.

This PR ensures proper DBAPI postgresql dialect handling for both initial container validation (via entrypoint) and runtime connections for an external postgres config database.

I chose this route because

1. Runtime database connections work with all URI formats through Flask-SQLAlchemy
2. No impact on pgadmin's internal postgres connections via wrapper function

Let me know if you prefer a different approach. Might have overlooked something / duplicate code isn't nice. Thanks.


web/pgadmin/__init__.py
```diff
@@ -335,7 +335,11 @@ def get_locale():
     ##########################################################################
     if config.CONFIG_DATABASE_URI is not None and \
             len(config.CONFIG_DATABASE_URI) > 0:
-        app.config['SQLALCHEMY_DATABASE_URI'] = config.CONFIG_DATABASE_URI
+        uri = config.CONFIG_DATABASE_URI
+        pg_pattern = r'^postgresql(\+psycopg3)?:'
```
adityatoshniwal, line 339: Can you move this logic to a common function probably in the file `web/pgadmin/utils/__init__.py` and then use it at both places?

adityatoshniwal (changes requested):
Otherwise, changes looks good to me.

web/pgadmin/__init__.py
sdsys-ch: @adityatoshniwal done and added doc changes. ok?

khushboovashi:
Hi @sdsys-ch, 

Do we really need this validation and correction in pgAdmin code? We can change the documentation and the config file comment properly, to avoid confusions. After that it is user's responsibility to set it properly.

Thanks,
Khushboo

sdsys-ch:
Hi @khushboovashi ,

Generally agree and don't specifically like the approach but might be relevant for backward compatibility / existing docker deployments. More of a service to the client thing, I guess.

Either way I'm good with both.

Regards,
Chris

khushboovashi:
@sdsys-ch, feel free to reopen it, if you encounter this issue again.

Issue #8095, opened 2024-11-06
Support Pg 17 builtin locale provider


duncand:
The latest pgAdmin's "create database" GUI feature lacks the new "builtin" option of Postgres 17 for locale provider and related options like "C.UTF-8" so one only create a database with the "builtin" provider using hand written SQL.

https://www.postgresql.org/docs/current/sql-createdatabase.html

This feature request is to expand the GUI to handle "builtin", and not just C/ICU that it currently implies are the only options.

Thank you.

PR #8122, opened 2024-11-14
Added feature to support built-in locale provider. #8095

pravesh-sharma:


docs/en_US/database_dialog.rst
```diff
@@ -44,8 +44,8 @@ Use the *Definition* tab to set properties for the database:
 * Select the strategy from the drop-down listbox in the *Strategy* field while
   creating a new database. This option is available from v15 and above.
 * Select the locale provider from the drop-down listbox in the *Locale Provider*
-  field to set the default collation for this database. Possible values are: icu, libc.
-  This option is available from v15 and above.
+  field to set the default collation for this database. Possible values are: icu, libc and builtin.
+  This option is available from v15 and above. **Note:** builtin option is available from v17 onwards.
```
akshay-joshi, line 48: This could be changed to: 
The icu and libc options are available from v15 and builtin option is available from v17 onwards.

web/pgadmin/browser/server_groups/servers/databases/static/js/database.ui.js
```diff
@@ -148,13 +150,21 @@ export default class DatabaseSchema extends BaseUISchema {
...
+          }, {
+            label: gettext('libc'),
+            value: 'libc',
+          }];
+          if(obj.getServerVersion() >= 170000) {
+            options.push({
+              label: gettext('builtin'), value: 'builtin',
+            });
+          }
+          return Promise.resolve(options);
```
akshay-joshi, line 166: Why do we need this? In the previous logic it was not there.

akshay-joshi (changes requested):
I did the eyeball review. Not tested.

web/pgadmin/browser/server_groups/servers/databases/static/js/database.ui.js
pravesh-sharma: We need this because the builtin locale provider is available on PG 17, so if the server version is below 17, the builtin option should not be available.

Issue #8127, opened 2024-11-14
Query tool prompts for unsaved changes even no changes have been made


natha1818:
Same as #5756
Bug is back:

Reproduce issue:

    open Query Tool
    click 'open file' icon and open a file
    -> 'save file' icon in toolbar is highlighted by default even before any changes occurred.
    click 'open file' icon to open another file
    -> prompts for 'Unsaved changes - Are you sure you want to discard the current changes?'

Tested on following config:
PGAdmin version 8.13
OS: Windows 10 Pro version 22H2 (OS build 19045.5131)

pravesh-sharma:
@RohitBhati8269  Found below issue:

If we change the EOL, the save button gets enabled. If any changes are made to the editor and it is reverted to its original state, the save button is disabled, which should not happen as the EOL has changed.

natha1818:
Also, when you open the second file, the file name on the tab doesn't change. When you open a third file, the name of the second file appears on the tab. And so on, always one file name behind.

pravesh-sharma:
@natha1818 I cannot reproduce the issue on the latest version of pgAdmin 4. Which version of pgAdmin 4 you are using?

natha1818:
PGAdmin version 8.13

OS: Windows 10 Pro version 22H2 (OS build 19045.5131)

anthonydb:
I am experiencing this issue on version 8.13 on macOS Sonoma 15.1.1

pravesh-sharma:
Working fine.

Issue #8010, opened 2024-10-04
pgAdmin fails to return results rows from FUNCTION


duncangroenewald:
pgAdmin fails to display the data rows from a FUNCTION.

See the example function below

```
CREATE OR REPLACE FUNCTION tablerows(ref1 refcursor)
RETURNS refcursor AS $$
BEGIN
OPEN ref1 FOR
WITH tbl AS
  (SELECT table_schema,
          TABLE_NAME
   FROM information_schema.tables
   WHERE TABLE_NAME not like 'pg_%'
     AND table_schema in ('makespace'))
SELECT table_schema,
       TABLE_NAME,
       (xpath('/row/c/text()', query_to_xml(format('select count(*) as c from %I.%I', table_schema, TABLE_NAME), FALSE, TRUE, '')))[1]::text::int AS rows_n
FROM tbl
ORDER BY rows_n DESC;
RETURN ref1;
END; $$ LANGUAGE plpgsql;

And this SQL is used to fetch the results from the function 

BEGIN;
SELECT tablerows('tables');
FETCH ALL FROM "tables";
COMMIT;
```

See attached image for the pgAdmin screen showing the results columns but no data rows.

Version 8.12
Application Mode Desktop
Commit: a0aa1855186bbe769182e797dc546812ed4dc720 2024-09-20
Current User pgadmin4@pgadmin.org
Electron Version 32.1.1
Browser Chrome 128.0.6613.137
Operating System macOS-14.6.1-arm64-arm-64bit

<img width="416" alt="Screenshot 2024-10-05 at 8 12 21 AM" src="https://github.com/user-attachments/assets/13d4b0e6-6442-467e-aaf2-a57f70f45da2">

pravesh-sharma:
The actual issue was that when multiple queries were executed together in the query tool, it returned results and messages from the last executed query, which were not being handled correctly. We have fixed this, and now it only shows the result and message from the last executed query.

The issue is fixed and tested on the latest snapshot [build](**url**).

PR #8147, opened 2024-11-19
Enhance query tool clarity and prevent unnecessary unsaved changes prompt. #8010 #8127

RohitBhati8269:
1. Improve query tool data output and messages for clarity. [Detail info](https://github.com/pgadmin-org/pgadmin4/issues/8010)
2. Prevent unsaved changes prompt when no changes are made in the query tool. [Detail info](https://github.com/pgadmin-org/pgadmin4/issues/8127)

### DIFF
--- a/web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
@@ -753,6 +753,8 @@ export class ResultSetUtils {
       }
       this.eventBus.fireEvent(QUERY_TOOL_EVENTS.SET_MESSAGE, tabMsg, true);
       this.eventBus.fireEvent(QUERY_TOOL_EVENTS.FOCUS_PANEL, PANELS.MESSAGES);
+      /* Clear the query data if the query has no result to display.*/
+      onResultsAvailable(null, [], []);
     }
     return retMsg;
   }
--- a/web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
@@ -10,7 +10,7 @@ import React, {useContext, useCallback, useEffect, useMemo } from 'react';
 import { format } from 'sql-formatter';
 import { QueryToolContext, QueryToolEventsContext } from '../QueryToolComponent';
 import CodeMirror from '../../../../../../static/js/components/ReactCodeMirror';
-import {PANELS, QUERY_TOOL_EVENTS} from '../QueryToolConstants';
+import {OS_EOL, PANELS, QUERY_TOOL_EVENTS} from '../QueryToolConstants';
 import url_for from 'sources/url_for';
 import { LayoutDockerContext, LAYOUT_EVENTS } from '../../../../../../static/js/helpers/Layout';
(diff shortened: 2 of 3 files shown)


web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
```diff
@@ -753,6 +753,7 @@ export class ResultSetUtils {
       }
       this.eventBus.fireEvent(QUERY_TOOL_EVENTS.SET_MESSAGE, tabMsg, true);
       this.eventBus.fireEvent(QUERY_TOOL_EVENTS.FOCUS_PANEL, PANELS.MESSAGES);
+      this.eventBus.fireEvent(QUERY_TOOL_EVENTS.SET_QUERY_DATA, null);
```
adityatoshniwal, line 756: What is this for, no comments added? And why do we need an event in the same component, which is not used in any other component?

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
```diff
@@ -1046,6 +1047,10 @@ export function ResultSet() {
       setSelectedRows(new Set());
       setAllRowsSelect('NONE');
     });
+
+    eventBus.registerListener(QUERY_TOOL_EVENTS.SET_QUERY_DATA, (data) => {
+      setQueryData(data);
```
adityatoshniwal, line 1052: You can directly call this.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
RohitBhati8269: In today's scrum call I discussed with the team about this issue Maybe you left the call.
It is a bug like:-
when we execute any `select` SQL it fetches the data and shows it in the data output tab
And then if we run `commit or alter`  SQL, it shows us a message, but still the data-output tab has the old `select` query value.
That's why I have made these changes. Currently, we are not clearing the data-output tab if a query does not have any result to display.

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
RohitBhati8269: Yes, I can call it directly as a prop.
But I thought if in future we can use that event.
Let me know if we don't want this event and I will call it directly.

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
adityatoshniwal: It was discussed in a call, but how will someone know when reading the code? Add some comments.

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
```diff
@@ -713,7 +713,7 @@ export class ResultSetUtils {
     return null;
   }
 
-  queryFinished(httpMessage, onResultsAvailable, onExplain) {
+  queryFinished(httpMessage, onResultsAvailable, onExplain, setQueryData) {
```
adityatoshniwal, line 716: why not use `onResultsAvailable` to setQueryData?

adityatoshniwal (changes requested):

Issue #6794, opened 2023-09-26
Pass host and user as environment variables to passwordexec subshell


pccowboy:
**Describe the solution you'd like**

Pass the current value of `host name` and `user name` as environment variables to the `PasswordExec.get` command.

This would allow easy use of tokens generated by commandline utilities (such as the AWS CLI)  to incorporate the database hostname, i.e.:

for the password exec command, instead of hard coding the host and user name
```
aws rds generate-db-auth-token --hostname some.cluster.us-east-2.rds.amazonaws.com --port 5432 --region us-east-2 --username my_iam_user
```

we could use
```
aws rds generate-db-auth-token --hostname $PGADMIN_HOSTNAME --port 5432 --region us-east-2 --username $PGADMIN_USERNAME
```
or something similar.

**Describe alternatives you've considered**
We could perform variable substitution on the password exec command itself, but I think that would be open to attack, where passing the specific values to the subprocess call's environment keeps things internal to pgadmin.

**Additional context**

yogeshmahajan-1903:
Working fine.

PR #8189, opened 2024-11-28
Added support for passing connection details as variables in the passexec command. #6794

pravesh-sharma:


docs/en_US/server_dialog.rst
```diff
@@ -205,6 +205,8 @@ Use the fields in the *Advanced* tab to configure a connection:
   command will be used as the SQL password. This may be useful when the password
   should be generated as a transient authorization token instead of providing a
   password when connecting in `PAM authentication <https://www.postgresql.org/docs/current/auth-pam.html>`_ scenarios.
+  You can pass server hostname, port, and DB username to the password exec command as variable by providing placeholders 
+  like ``%DB_HOSTNAME%``, ``%DB_PORT%`` and ``%DB_USERNAME%`` which will be replaced with the server connection information.
```
adityatoshniwal, line 209: I think `DB_` is redundant and can be removed. It can be same as what we have to title config. Plus, its server hostname, not db hostname.
Please add one example as well.

adityatoshniwal (changes requested):

docs/en_US/server_dialog.rst
```diff
@@ -205,6 +205,9 @@ Use the fields in the *Advanced* tab to configure a connection:
   command will be used as the SQL password. This may be useful when the password
   should be generated as a transient authorization token instead of providing a
   password when connecting in `PAM authentication <https://www.postgresql.org/docs/current/auth-pam.html>`_ scenarios.
+  You can pass server hostname, port and DB username to the password exec command as variable by providing placeholders
+  like ``%HOST%``, ``%POSR%`` and ``%USERNAME%`` which will be replace with the server connection information.
```
khushboovashi, line 209: like ``%HOST%``, ``%PORT%`` and ``%USERNAME%`` which will be replaced with the server connection information.

Issue #7330, opened 2024-03-29
readOnlyRootFilesystem to true in Kubernetes


albundy83:
**Describe the solution you'd like**
Hello,

it will be nice that files that are updated by `entrypoint.sh` file be written to a dedicated **empty** folder.
At least file `/pgadmin4/config_distro.py`.

For moment, we can't use an `emptyDir` or `persistentVolumeClaim` as there are many files in `/pgadmin4` folder.

I will be a nice step to achieve a more secure pgadmin4 Deployment by being capable to set `readOnlyRootFilesystem` to true.

**Describe alternatives you've considered**
Generate my self the `/pgadmin4/config_distro.py` as ConfigMap or Secret and mount it but we should be able to disable generation.

Specify a custom path for `/pgadmin4/config_distro.py` with ENV var.

**Additional context**

oldboys92:
also looking for a solution to this 👍

balazsjdp:
Hey, is there any solution for this?

yogeshmahajan-1903:
@albundy83 , @oldboys92 , @balazsjdp 
Does below configuration solved your issue?

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: pgadmin
  labels:
    app: pgadmin
spec:
  replicas: 1
  selector:
    matchLabels:
      app: pgadmin
  template:
    metadata:
      labels:
        app: pgadmin
    spec:
      containers:
        - name: pgadmin
          image: dpage/pgadmin4:8.13
          env:
            - name: PGADMIN_DEFAULT_EMAIL
              value: "admin@example.com"
            - name: PGADMIN_DEFAULT_PASSWORD
              value: "admin"
          ports:
            - containerPort: 80
          securityContext:
            readOnlyRootFilesystem: true  # Make the root filesystem read-only
          volumeMounts:
            - name: pgadmin-data
              mountPath: /var/lib/pgadmin
              readOnly: false
            - name: pgadmin-tmp
              mountPath: /tmp  # Mount tmp volume
              readOnly: false
            - name: pgadmin-config-volume
              mountPath: /pgadmin4/config_distro.py
              subPath: config_distro.py
      volumes:
        - name: pgadmin-data
          emptyDir: {}
        - name: pgadmin-tmp
          emptyDir: {}  # Temporary directory that Kubernetes will manage
        - name: pgadmin-config-volume
          configMap:
            name: pgadmin-config-distro

---

apiVersion: v1
[...]

PR #8214, opened 2024-12-02
Add support to deploy pgadmin in container with readOnlyRootFilesystem as true.#7330

yogeshmahajan-1903:


web/pgadmin/evaluate_config.py
```diff
@@ -47,7 +58,10 @@ def validate_config_variable(key, value):
 
 # Load distribution-specific config overrides
 try:
-    import config_distro
+    if 'CONFIG_DISTRO_FILE_PATH' in os.environ:
```
khushboovashi, line 61: What if 'CONFIG_DISTRO_FILE_PATH' not in the env? Does this logic work with the current implementation?

yogeshmahajan-1903: This works with existing implementation. This is required only for building the docker image. But let me add else block too.

khushboovashi: Yeah, else block is required. Otherwise, it will not work where we have a distro file other than Docker env.

Issue #8065, opened 2024-10-24
"Crypt Key missing" - Worker (pid:XX) was sent SIGKILL!


ghost:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Whenever a query is consuming resources, it produces an error that makes pgadmin container restart and give an SIGKILL before doing so:

```
172.25.9.248 - - [24/Oct/2024:15:37:40 +0000] "GET /sqleditor/poll/8544899 HTTP/1.1" 200 453 "https://pgadmin-pgadmin-prod.apps.dev.ocp.domain.com/sqleditor/panel/8544899?is_query_tool=true&sgid=361&sid=1856&did=16413&database_name=eutras01-prod" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Safari/537.36 Edg/130.0.0.0"
172.25.9.248 - - [24/Oct/2024:15:37:41 +0000] "GET /sqleditor/poll/8544899 HTTP/1.1" 200 453 "https://pgadmin-pgadmin-prod.apps.dev.ocp.domain.com/sqleditor/panel/8544899?is_query_tool=true&sgid=361&sid=1856&did=16413&database_name=eutras01-prod" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Safari/537.36 Edg/130.0.0.0"
[2024-10-24 15:37:41 +0000] [1] [ERROR] **Worker (pid:21) was sent SIGKILL! Perhaps out of memory?**
[2024-10-24 15:37:41 +0000] [122] [INFO] Booting worker with pid: 122
2024-10-24 15:37:48,566: INFO	pgadmin:	########################################################
2024-10-24 15:37:48,566: INFO	pgadmin:	Starting pgAdmin 4 v8.12...
2024-10-24 15:37:48,566: INFO	pgadmin:	########################################################
[...]

PR #8238, opened 2024-12-05
Ensure CSRF errors handled to return unauthorized response.#8065

yogeshmahajan-1903:


web/pgadmin/__init__.py
```diff
@@ -915,13 +917,15 @@ def all_exception_handler(e):
...
-            error = str(e.description) + 'Please refresh the page.'
-            return internal_server_error(errormsg=gettext(error))
         return e
 
-    # Intialize the key manager
+    # Send unauthorized response if CSRF errors occurs.
+    @app.errorhandler(CSRFError)
+    def handle_csrf_error(error):
+        err_msg = str(error.description) + ' You need to refresh the page.'
+        return unauthorized(errormsg=gettext(err_msg))
```
khushboovashi, line 926: gettext should be applied to the string, not to the variable.

khushboovashi (changes requested):

web/pgadmin/__init__.py
yogeshmahajan-1903: Done

PR #8239, opened 2024-12-05
add ignore partitions in schema diff

alexander-smyslov:
Often there is a situation when the only difference in the schemes is the number of partitions. It would be convenient to have an option to ignore this

### DIFF
--- a/docs/en_US/preferences.rst
+++ b/docs/en_US/preferences.rst
@@ -562,6 +562,8 @@ Use the *Ignore Tablespace* switch to ignores the tablespace while comparing the
 Use the *Ignore Whitespace* switch to ignores the whitespace while comparing
 the string objects. Whitespace includes space, tabs, and CRLF.
 
+Use the *Ignore Partitions* switch to ignores the partitions while comparing
+the partitions tables. 
 
 The Storage Node
 ****************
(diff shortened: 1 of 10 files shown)


pravesh-sharma:
@alexander-smyslov, Just to clarify, you are trying to add an option so that if the user selects it, the schema diff will not compare the partitions of the table and will not generate any diff SQL. If this is what you're trying to achieve, then the code in the PR is not having the desired effect, it is still showing the difference.

alexander-smyslov:
> @alexander-smyslov, Just to clarify, you are trying to add an option so that if the user selects it, the schema diff will not compare the partitions of the table and will not generate any diff SQL. If this is what you're trying to achieve, then the code in the PR is not having the desired effect, it is still showing the difference.

thanks for review. merge was incomplete. fixed.

pravesh-sharma:
> > @alexander-smyslov, Just to clarify, you are trying to add an option so that if the user selects it, the schema diff will not compare the partitions of the table and will not generate any diff SQL. If this is what you're trying to achieve, then the code in the PR is not having the desired effect, it is still showing the difference.
> 
> thanks for review. merge was incomplete. fixed.

The code still does not have any effect it still shows the difference between the partitions.

alexander-smyslov:
image

pravesh-sharma:
> image

If we ignore the partition, it should not be shown as different; it should be shown as identical. Also, I have another test case where the source DB has a table with 3 partitions and the target DB has the same table with 2 partitions. In this case, it should also be ignored, but this is not being handled correctly now.

akshay-joshi:
There is no response from the author; hence closing it.

Issue #7708, opened 2024-07-18
Add a side workspace, similar to those in VSCode and PyCharm, to pgAdmin 4.


akshay-joshi:
Implement separate workspace for Query Tool, PSQL, Schema Diff, etc.
Attached is the POC work done.

There are a couple of things that we need to check:

1. Persistent with the already opened tabs.
2. Drag and Drop objects in the Query Tool.
3. Classic UI vs New and their respective changes.
4. Open an empty query tool tab which is difficult as a part of POC at the moment. 

[SideWorkspace_POC.patch](https://github.com/user-attachments/files/16278542/SideWorkspace_POC.patch)

anthonydb:
@akshay-joshi I tried out the implementation of this and have a few thoughts to share from the perspective of a user:

* Overall, this is a nice feature. Less clutter is good.
* However, as a user working in the Query Tool, I often have need to reference the Object Explorer. I often drag/drop column or table names into the Query Tool.  Or I need to simply look at table structure and column names. I understand that I can click over to the "Default Workspace" to see table structure, but the ability to simply glance back and forth between queries I'm writing and the explorer is helpful.
* Thus, this implementation gives me only two choices -- 1) less clutter and more difficult access to Object Explorer or 2) changing settings to use the "classic" UI and not have access to the less-cluttered UI.

There could be a combination that would solve this issue: Simply make the Object Explorer a collapsible pane in the Query Tool workspace. When a user selects the Query Tool workspace, the explorer is closed by default. You would just need a way to signal to the user that it is available to be opened.

akshay-joshi:
@anthonydb 

I agree that Drag-and-Drop is not supported in the Workspace layout. Could you please raise a feature request for it? We'll evaluate its feasibility. Additionally, you don’t need to navigate to the Object Explorer to view table structures or column names—you can use the autocomplete feature (CTRL + SPACE) in pgAdmin.

We considered making the Object Explorer a collapsible pane in the classic design. However, the Workspace layout is designed as a context-free environment, allowing users to connect to ad-hoc servers that may not be registered in the Object Explorer."

anthonydb:
@akshay-joshi 

I have opened https://github.com/pgadmin-org/pgadmin4/issues/8293. Thanks for considering.

yogeshmahajan-1903:
Works fine.

QuotallRmw:
Firstly well done on producing such a good tool, thanks for all your work on this.

[...]

PR #8253, opened 2024-12-11
Enhanced pgAdmin 4 with support for Workspace layouts. #7708

akshay-joshi:


web/pgadmin/preferences/__init__.py
```diff
@@ -241,6 +244,10 @@ def save():
         if not res:
             return internal_server_error(errormsg=msg)
 
+        if 'cleanup_on_layout_change' in data and \
```
adityatoshniwal, line 247: Do we need this? We can decided based on key and value.

web/pgadmin/tools/schema_diff/static/js/SchemaDiffModule.js
```diff
@@ -60,7 +61,8 @@ export default class SchemaDiff {
   }
 
   launchSchemaDiff() {
-    let panelTitle = gettext('Schema Diff');
+    let panelTitle = gettext('Schema Diff - ' + SchemaDiff.panelTitleCount);
```
adityatoshniwal, line 64: First panel doesn't need to have have suffix. Suffix can be added from 2.
```
SchemaDiff.panelTitleCount > 1 ? gettext('Schema Diff (%s)', SchemaDiff.panelTitleCount) : gettext('Schema Diff')
```

adityatoshniwal (changes requested):

web/pgadmin/preferences/__init__.py
```diff
@@ -269,6 +276,12 @@ def save():
                             samesite=config.SESSION_COOKIE_SAMESITE,
                             **domain)
 
+    # if layout is changed from 'Workspace' to 'Classic', disconnect all
+    # servers.
+    if has_layout_changed:
+        disconnect_from_all_servers()
```
adityatoshniwal, line 282: Following the `separation of concerns` principle, this can be taken care in a separate API call after successful pref save.

adityatoshniwal (changes requested):

web/pgadmin/browser/server_groups/servers/utils.py
```diff
@@ -560,3 +561,116 @@ def get_replication_type(conn, sversion):
...
+        manager.release()
+
+
+def delete_adhoc_servers():
+    """
+    This function will remove all the adhoc servers.
+    """
+    try:
+        db.session.query(Server).filter(Server.is_adhoc == 1).delete()
+        db.session.commit()
```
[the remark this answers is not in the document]
akshay-joshi: I have tried but it is throwing errors for SQLite and PostgreSQL.

web/pgadmin/preferences/__init__.py
akshay-joshi: Fixed

PR #8258, opened 2024-12-12
Fix in case of the file does not exist

albundy83:


albundy83:
Fix for https://github.com/pgadmin-org/pgadmin4/issues/7330

yogeshmahajan-1903 (approved):

Issue #6513, opened 2023-06-29
Distinguishing dialogs for deleting and disconnecting from the database


chvilalukas:
**Describe the solution you'd like**

The dialogs for deleting the database and disconnecting from the database look the same at first glance. It would be useful to distinguish them more clearly (by color or icon).

**Additional context**

It's only a matter of time before I find myself selecting delete database from the menu instead of disconnecting from database. And since both dialogs look very similar I'll overlook it.
![Snímek obrazovky 2023-06-29 v 11 38 47](https://github.com/pgadmin-org/pgadmin4/assets/9366179/a20c4d8c-57bf-4f81-be3b-01df6dad809e)
![Snímek obrazovky 2023-06-29 v 11 39 01](https://github.com/pgadmin-org/pgadmin4/assets/9366179/1fbfadd4-4466-4963-a2df-49096fcf123e)

nikhil-mohite:
We should change the label of the buttons to 'Delete' or 'Disconnect' and 'No' to 'Cancel'

chvilalukas:
Thank you, that might help at least a little. But I think it would still need some visual differentiation (color or icon). PgAmin is also used on production environments, and if someone accidentally deletes the database there, that's really a bummer.

yogeshmahajan-1903:
Console error while cancelling operation 
<img width="1728" alt="Screenshot 2025-01-16 at 4 21 10 PM" src="https://github.com/user-attachments/assets/62285a66-e8c8-44c0-9fda-457c5ffc5c27" />

yogeshmahajan-1903:
Working fine.

PR #8284, opened 2024-12-23
Fix the button focus issue. #6513

RohitBhati8269:
Detail info:- https://github.com/pgadmin-org/pgadmin4/issues/6513

### DIFF
--- a/web/pgadmin/static/js/helpers/ModalProvider.jsx
+++ b/web/pgadmin/static/js/helpers/ModalProvider.jsx
@@ -13,13 +13,14 @@ import { getEpoch } from 'sources/utils';
 import { DefaultButton, PgIconButton, PrimaryButton } from '../components/Buttons';
 import Draggable from 'react-draggable';
 import CloseIcon from '@mui/icons-material/CloseRounded';
+import DeleteIcon from '@mui/icons-material/Delete';
 import CustomPropTypes from '../custom_prop_types';
 import PropTypes from 'prop-types';
 import gettext from 'sources/gettext';
 import HTMLReactParser from 'html-react-parser';
 import CheckRoundedIcon from '@mui/icons-material/CheckRounded';
 import { Rnd } from 'react-rnd';
-import { ExpandDialogIcon, MinimizeDialogIcon } from '../components/ExternalIcon';
+import { ExpandDialogIcon, MinimizeDialogIcon, DisconnectedIcon } from '../components/ExternalIcon';
 import { styled } from '@mui/material/styles';
 
 export const ModalContext = React.createContext({});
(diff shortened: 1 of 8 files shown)


web/pgadmin/static/js/helpers/ModalProvider.jsx
```diff
@@ -53,18 +54,20 @@ function renderExtraButtons(button) {
...
         {confirm &&
-          <DefaultButton startIcon={<CloseIcon />} onClick={onCancelClick} autoFocus={true}>{cancelLabel}</DefaultButton>
+          <DefaultButton startIcon={<CloseIcon />} onClick={onCancelClick} autoFocus={hasAutoFocus} >{cancelLabel}</DefaultButton>
         }
         {
           extraButtons?.length ?
             extraButtons.map(button=>renderExtraButtons(button))
             :
-            <PrimaryButton className='Alert-margin' startIcon={<CheckRoundedIcon />} onClick={onOkClick}>{okLabel}</PrimaryButton>
+            <PrimaryButton className='Alert-margin' startIcon={<CheckRoundedIcon/>} onClick={onOkClick} autoFocus={true}>{okLabel}</PrimaryButton>
```
adityatoshniwal, line 70: You have applied autofocus to this button and above Default button also. this will not work.

web/pgadmin/static/js/helpers/Notifier.jsx
```diff
@@ -68,6 +70,20 @@ AlertContent.propTypes = {
   cancelLabel: PropTypes.string,
 };
 
+function createButtonConfig(icon, label, onOkClick, closeModal, type, color=null, focus=false){
```
adityatoshniwal, line 73: why do we need such functions which solves nothing?

web/pgadmin/static/js/helpers/Notifier.jsx
```diff
@@ -182,20 +198,16 @@ class Notifier {
...
-            onOkClick();
-            closeModal();
-          },
-          color: 'error',
-        },
-      ];
+    const extraButtons = (closeModal) => {
+      // Button configurations for specific actions like "Delete" and "Disconnect".
+      const buttonConfigs = {
+        [gettext('Delete')]: () =>
```
adityatoshniwal, line 204: the buttons config was an array, why are you creating a dictionary? You're making simple code complex.

adityatoshniwal (changes requested):

web/pgadmin/static/js/helpers/ModalProvider.jsx
RohitBhati8269: As per this requirement :- 
**For the delete option, I would also:
- Make Cancel the default option.**

I added Auto-focus true on the Cancel button when there Is an extra button with a focus prop. that's why I've added conditionally Auto-focus at that place

web/pgadmin/static/js/helpers/Notifier.jsx
RohitBhati8269: Just to ensure That the code stays DRY for better readability and flexibility.
If want to remove?

web/pgadmin/static/js/helpers/ModalProvider.jsx
adityatoshniwal: And if there are no extra buttons then? hasFocus will be true and default button + primary button both will have autoFocus.

web/pgadmin/static/js/helpers/Notifier.jsx
adityatoshniwal: This is destroying the readability.

web/pgadmin/static/js/helpers/ModalProvider.jsx
RohitBhati8269: The focus does not depend on the extra button. It depends on the focus property of the extra button. And this focus property we are passing only for **Delete** button.

adityatoshniwal: As per your logic, if any extra has focus then default button will have focus?

adityatoshniwal (changes requested):

web/pgadmin/static/js/helpers/ModalProvider.jsx
RohitBhati8269: As we are checking focus props from extraButtons and
If there are extraButtons then this **PrimaryButton** won't render.
It will render a button from **renderExtraButtons**.

adityatoshniwal: I remember when extraButtons was added, intention was to show `extra` buttons apart from existing ones. Now you've changed it. Extras is no more extra. It has become the main buttons.

web/pgadmin/static/js/helpers/Notifier.jsx
```diff
@@ -175,29 +189,38 @@ class Notifier {
...
-            closeModal();
-          },
-          color: 'error',
-        },
-      ];
-    };
-    this.modal.confirm(title, text, onOkClick, onCancelClick, okLabel, cancelLabel, extraButtons);
+  confirmDelete(title, text, onOkClick, onCancelClick, okLabel = gettext('Delete'), cancelLabel = gettext('Cancel')){
+    let isDeleteButton = gettext('Delete') === okLabel;
+    this.modal.showModal(
```
adityatoshniwal, line 200: create a function confirmDelete in modal provider and call here.

web/pgadmin/static/js/helpers/Notifier.jsx
```diff
@@ -175,29 +189,38 @@ class Notifier {
...
-            onOkClick();
-            closeModal();
-          },
-          color: 'error',
-        },
-      ];
-    };
-    this.modal.confirm(title, text, onOkClick, onCancelClick, okLabel, cancelLabel, extraButtons);
+  confirmDelete(title, text, onOkClick, onCancelClick, okLabel = gettext('Delete'), cancelLabel = gettext('Cancel')){
+    let isDeleteButton = gettext('Delete') === okLabel;
```
adityatoshniwal, line 199: In a dialog type `confirmDelete` focus should by default be on Cancel. Checking based on label better be avoided.

adityatoshniwal (changes requested):

PR #8343, opened 2025-01-08
Ensure double click event is not listened on arrows beside database object in the browser tree.

yogeshmahajan-1903:


web/pgadmin/tools/search_objects/static/js/index.js
```diff
@@ -56,7 +56,10 @@ export default class SearchObjectModule {
     let isEnabled = (() => {
       if (!_.isUndefined(obj) && !_.isNull(obj)) {
         if (_.indexOf(pgAdmin.unsupported_nodes, obj._type) == -1) {
-          if (obj._type == 'database' && obj.allowConn) {
+          let t = pgBrowser.tree,
+            selected_item = t.selected(),
```
adityatoshniwal, line 60: if we're finding the selected item then what is the use of param obj. obj is the selected tree item right?

adityatoshniwal (changes requested):

web/pgadmin/tools/search_objects/static/js/index.js
yogeshmahajan-1903: obj is not always selected item. In case of double click on serve node, selected item is server but as we expand server, obj gets values of all db objects listed below server.

khushboovashi (commented):
@yogeshmahajan-1903, have you removed double click on database object too?

yogeshmahajan-1903:
> @yogeshmahajan-1903, have you removed double click on database object too?

No. Double click on arrow is removed. Double click on label/db object works.

web/pgadmin/tools/search_objects/static/js/index.js
adityatoshniwal: Then you should probably check the source of obj. obj is intended to check for particular object when selected.

Issue #7617, opened 2024-06-24
Updating column name does not update it on the constraint include column


pravesh-sharma:
**Describe the bug**
When creating a new table if the column name is changed and it is used inside the constraint include column it is not updated.

**To Reproduce**

1. Start creating a table and add columns.
2. Move to the constraint tab and create a constraint e.g. a Primary key.
3. Select columns and include columns for the primary key.
4. Move back to the columns tab and change the column names.
5. Go to the constraint tab, the column name is updated but the include column is still old.

**Expected behavior**
Column name should be updated on the include column field as well.

**Screen recording**

https://github.com/pgadmin-org/pgadmin4/assets/111979571/67506323-6900-42b6-b50e-297bcc5dcff6

**Desktop (please complete the following information):**
 - Version:8.8
 - Mode: Desktop

pravesh-sharma:
This needs to be fixed for foreign key and unique constraint.

pravesh-sharma:
Issue is fixed for primary key.

Created a new issue for foreign key and unique constraint. https://github.com/pgadmin-org/pgadmin4/issues/9060

PR #8348, opened 2025-01-09
Sync include columns with column name changes.#7617

RohitBhati8269:
Detail Info:- #7617

### DIFF
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/static/js/column.ui.js
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/static/js/column.ui.js
@@ -60,6 +60,7 @@ export default class ColumnSchema extends BaseUISchema {
       seqcycle: undefined,
       colconstype: 'n',
       genexpr: undefined,
+      isInclude: false
     });
 
     this.getPrivilegeRoleSchema = getPrivilegeRoleSchema;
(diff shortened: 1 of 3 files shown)


web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/static/js/column.ui.js
```diff
@@ -60,6 +60,7 @@ export default class ColumnSchema extends BaseUISchema {
       seqcycle: undefined,
       colconstype: 'n',
       genexpr: undefined,
+      isInclude: false
```
adityatoshniwal, line 63: I don't think this is required. The problem is with syncing the column names. Normal columns works without any extra variables.

adityatoshniwal (changes requested):

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/static/js/column.ui.js
RohitBhati8269: The **isInclude** variable helps track which columns are added to the "Include Column" field in the create table modal, especially when column names change. This works similarly to the **is_primary_key** variable that tracks columns in the model's column field. This helps maintain proper tracking of column states. Let me know if you'd prefer a different approach!

adityatoshniwal: I don't see any is_foreign_key or is_unique_key variables. I don't think it is the right direction.

adityatoshniwal: Then what is the use of `include_columns` key?

Issue #6044, opened 2023-03-29
BUG: Query inexplicably can't be modified (hint: when database connection is lost)


usernamesareaholdoverfromthe90s:
Test case:

1. Drill down to any table, and right-click on the table and select "View/Edit Data..." | "Filtered Rows..."
2. Type in some valid query
3. Wait for your connection to be lost (note: there is NOTHING that indicates the connection is(n't) lost)
5. Click on the "Filter Options" button
6. Type in a new query

BUG: You can't click the Save button and there's no (apparent) feedback as to why not.

It turns out that if you subsequently re-execute the old query, then you'll get the message that the database connection was lost and then are offered to reconnect.  Once you do that, then you can successfully edit the query, as expected.

This behaviour is shown in the attached screen capture video.

Expected behaviour:

Given that the Save button is disabled in the Filter Options, then clearly pgAdmin knows the connection is lost.   So ideally, simply clicking on the "Filter Options" button would offer to reconnect you (same as re-executing the query), and once the connection is reestablished, bring up the Filter Options menu.

Less ideal would be to provide some feedback upon clicking the Save button saying that the connection was lost.

https://user-images.githubusercontent.com/117369461/228634770-a6130990-393a-4505-82e6-9ef3b9ae3fdf.mp4

RohitBhati8269:
Faced a bug while testing:-
When password not save for any server.
1. Drill down to any table, and right-click on the table and select "View/Edit Data..." | "Filtered Rows..."
2. Type in some valid query
3. Wait for your connection to be lost / disconnect from the server.
4. Click on the "Filter Options" button
5. Type in a new query
6. click on Save (it will ask for password to reconnect the server)
7. Now on QueryTool you will not see any updated query or filter which you applied.

RohitBhati8269:
Working fine Tested and Verified on :- [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-01-30/)
Package :- wheel
OS :- macOS Sequoia 15.0.1

PR #8360, opened 2025-01-15
Fixed an issue when the server/database connection was lost; the filter dialog is not getting saved. #6044

anilsahoo20:


web/pgadmin/tools/sqleditor/static/js/components/dialogs/FilterDialog.jsx
```diff
@@ -103,20 +105,85 @@ class FilterSchema extends BaseUISchema {
 export default function FilterDialog({onClose, onSave}) {
 
   const queryToolCtx = React.useContext(QueryToolContext);
+  const eventBus = React.useContext(QueryToolEventsContext);
   const filterSchemaObj = React.useMemo(()=>new FilterSchema([]));
 
+  const connectServerModal = (modalData, connectCallback, cancelCallback) => {
+    queryToolCtx.modal.showModal(gettext('Connect to server'), (closeModal)=>{
```
adityatoshniwal, line 112: Do we really need a connect server dialog inside a filter dialog? Filter dialog could simply save the filters and query execution can ask for server connect I believe. Only case where filter dialog won't be able to save is when pgAdmin server itself is down which is fine.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/static/js/components/dialogs/FilterDialog.jsx
anilsahoo20: What happens is that when the server connection is lost and we open the filter dialog, there we are calling a filter/dialog API to get the preliminary sort and sql data, and this gives connection lost error as server is disconnected. And if password is not saved for server then to reconnect to the server we need to call the Connect server dialog for asking password, once user enters we will reconnect the server and recall the filter/dialog api and gets the data and populate it on the dialog.

web/pgadmin/tools/sqleditor/utils/start_running_query.py
```diff
@@ -82,26 +81,15 @@ def execute(self, sql, trans_id, http_session, connect=False):
 
             # Connect to the Server if not connected.
             if connect and not conn.connected():
-                view = SchemaDiffRegistry.get_node_view('server')
-                response = view.connect(transaction_object.sgid,
-                                        transaction_object.sid, True)
-                if response.status_code == 428:
+                from pgadmin.tools.sqleditor.utils import \
```
khushboovashi, line 84: Why import here?

web/pgadmin/tools/sqleditor/utils/query_tool_connection_check.py
```diff
@@ -0,0 +1,65 @@
...
+        session,
+        trans_id
+    )
+    if isinstance(session_obj, Response):
+        return session_obj
+
+    transaction_object = pickle.loads(session_obj['command_obj'])
+
+    if transaction_object is not None and session_obj is not None:
+        view = SchemaDiffRegistry.get_node_view('server')
```
khushboovashi, line 36: Please add a comment here properly regarding in which scenario we need this connection here.

web/pgadmin/tools/sqleditor/utils/start_running_query.py
anilsahoo20: As in the query_tool_connection_check function we need to verify the trans_obj and session_obj available we are importing the StartRunningQuery class there and due to a circular import, I am importing the function `query_tool_connection_check` where it is needed only.

web/pgadmin/tools/sqleditor/utils/query_tool_connection_check.py
anilsahoo20: Sure

web/pgadmin/tools/sqleditor/static/js/components/dialogs/FilterDialog.jsx
adityatoshniwal: We should definitely avoid need to connect to database in filter dialog. The filter dialog can be changed to not to fetch columns from db and used cached columns info. The columns info can be updated on next execute query. That way, no connection will be needed for filter dialog and connecting server task will be delegated back to execute query.

web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -839,6 +842,9 @@ def start_view_data(trans_id):
         # set fetched row count to 0 as we are executing query again.
         trans_obj.update_fetched_row_cnt(0)
 
+        # Fetch the columns for the table and store it in session
+        trans_obj.fetch_all_columns(conn)
```
adityatoshniwal, line 846: why not use - https://github.com/pgadmin-org/pgadmin4/blob/f75291aec83a6833d392355f26cf602582e0ffe9/web/pgadmin/tools/sqleditor/__init__.py#L1101 instead of fetching the columns yourself. Query execution will give the columns right?

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/__init__.py
anilsahoo20: Done

web/pgadmin/tools/sqleditor/static/js/components/dialogs/FilterDialog.jsx
anilsahoo20: Done

PR #8361, opened 2025-01-15
Ensure double click event is not ignored in the browser tree.

yogeshmahajan-1903:


khushboovashi (commented):
I think, below code would work rather then introducing the new code.
@yogeshmahajan-1903, check the code below, that should work.


  private readonly handleClick = (ev: React.MouseEvent) => {
    const { item, itemType, onClick } = this.props;
    let isSingleClick;
    if (isSingleClick = event.detail == 1){
      setTimeout(() => {
        if(isSingleClick){ 
          if (itemType === ItemType.File || itemType === ItemType.Directory) {
               console.warn("Handle Single Click");
              onClick(ev, item as FileEntry, itemType);
            }
        }
      }, 200);
    }
    else if (event.detail == 2) {
      console.warn("Handle Double Click");
      this.handleDoubleClick();
    }
  };

yogeshmahajan-1903:
> I think, below code would work rather then introducing the new code. @yogeshmahajan-1903, check the code below, that should work.
> 
> private readonly handleClick = (ev: React.MouseEvent) => { const { item, itemType, onClick } = this.props; let isSingleClick; if (isSingleClick = event.detail == 1){ setTimeout(() => { if(isSingleClick){ if (itemType === ItemType.File || itemType === ItemType.Directory) { console.warn("Handle Single Click"); onClick(ev, item as FileEntry, itemType); } } }, 200); } else if (event.detail == 2) { console.warn("Handle Double Click"); this.handleDoubleClick(); } };

Having separate hook will be helpful if we need it in some other places too. I have sampled the code.

khushboovashi:
> > I think, below code would work rather then introducing the new code. @yogeshmahajan-1903, check the code below, that should work.
> > private readonly handleClick = (ev: React.MouseEvent) => { const { item, itemType, onClick } = this.props; let isSingleClick; if (isSingleClick = event.detail == 1){ setTimeout(() => { if(isSingleClick){ if (itemType === ItemType.File || itemType === ItemType.Directory) { console.warn("Handle Single Click"); onClick(ev, item as FileEntry, itemType); } } }, 200); } else if (event.detail == 2) { console.warn("Handle Double Click"); this.handleDoubleClick(); } };
> 
> Having separate hook will be helpful if we need it in some other places too. I have sampled the code.

I don't think we need a separate hook here, as it could be managed without a React state.

adityatoshniwal:
> > Having separate hook will be helpful if we need it in some other places too. I have sampled the code.
> 
> I don't think we need a separate hook here, as it could be managed without a React state.

It is important to cleanup setTimeout in React as it can lead to many adverse unknown effects. Having a hook is good for properly initiating and cleaning up things.

khushboovashi:
> > > Having separate hook will be helpful if we need it in some other places too. I have sampled the code.
> > 
> > 
> > I don't think we need a separate hook here, as it could be managed without a React state.
> 
> It is important to cleanup setTimeout in React as it can lead to many adverse unknown effects. Having a hook is good for properly initiating and cleaning up things.

As I mentioned earlier, if the event itself gives the event details, there is no need to save it in the React state.

adityatoshniwal:
> > > > Having separate hook will be helpful if we need it in some other places too. I have sampled the code.
> > > 
> > > 
> > > I don't think we need a separate hook here, as it could be managed without a React state.
> > 
> > 
> > It is important to cleanup setTimeout in React as it can lead to many adverse unknown effects. Having a hook is good for properly initiating and cleaning up things.
> 
> As I mentioned earlier, if the event itself gives the event details, there is no need to save it in the React state.

I'm talking about hook and setTimeout cleanup. Not talking about states.

khushboovashi:
> > > > > Having separate hook will be helpful if we need it in some other places too. I have sampled the code.
> > > > 
> > > > 
> > > > I don't think we need a separate hook here, as it could be managed without a React state.
> > > 
> > > 
> > > It is important to cleanup setTimeout in React as it can lead to many adverse unknown effects. Having a hook is good for properly initiating and cleaning up things.
> > 
> > 
> > As I mentioned earlier, if the event itself gives the event details, there is no need to save it in the React state.
> 
> I'm talking about hook and setTimeout cleanup. Not talking about states.

I am against Hook in this particular case.

adityatoshniwal:
> > > > > > Having separate hook will be helpful if we need it in some other places too. I have sampled the code.
> > > > > 
> > > > > 
> > > > > I don't think we need a separate hook here, as it could be managed without a React state.
> > > > 
> > > > 
> > > > It is important to cleanup setTimeout in React as it can lead to many adverse unknown effects. Having a hook is good for properly initiating and cleaning up things.
> > > 
> > > 
> > > As I mentioned earlier, if the event itself gives the event details, there is no need to save it in the React state.
> > 
> > 
> > I'm talking about hook and setTimeout cleanup. Not talking about states.
> 
> I am against Hook in this particular case.

As I mentioned earlier, hooks are good for properly initiating and cleaning up things at a single place. And they're re-usable. This is one case where we can introduce it and later use the same hook wherever needed. Duplication of logic/code should be avoided.

Issue #8299, opened 2024-12-27
Disable master password in server mode


johansmitsnl:
**Describe the solution you'd like**

We are using oauth2 for authentication and the idea is that we don't want to have password (besides the sql servers) to access pgadmin. There is `PGADMIN_CONFIG_MASTER_PASSWORD_REQUIRED` but this only works in desktop mode. Can this also support server mode?

**Describe alternatives you've considered**

None

drehelis:
Upgrading from `8.11` -> `8.14` brings back "Set Master Password" popup even thought `PGADMIN_CONFIG_MASTER_PASSWORD_REQUIRED` was in place.

`PGADMIN_CONFIG_AUTHENTICATION_SOURCES="['oauth2']"` is also set for us.

knutia:
Same issue using :
PGADMIN_CONFIG_AUTHENTICATION_SOURCES="['webserver']"
Was working in fine in 8.13
Upgrading from 8.13 -> 8.14 stoped working.

anilsahoo20:
@yogeshmahajan-1903 found some issues while testing,
1. AUTHENTICATION_SOURCES = ['oauth2'], MASTER_PASSWORD_REQUIRED=False, and on Login>Open query tool of a DB>Logout, it gives the below error
```
{
  "success": 0,
  "errormsg": "'NoneType' object is not subscriptable",
  "info": "",
  "result": null,
  "data": null
}
```

2. AUTHENTICATION_SOURCES = ['internal'] or AUTHENTICATION_SOURCES = ['oauth2'], and on Login>Open query tool of a DB>Logout, it gives the below error
```
{
  "success": 0,
  "errormsg": "'utf-8' codec can't decode byte 0x8b in position 0: invalid start byte",
  "info": "",
  "result": null,
  "data": null
}
```

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-01-28/
Package: arm64, pip wheel
Environment: macOs Sonoma 14.7.1

PR #8373, opened 2025-01-20
Fixed the error occurring during logout from pgadmin when using OAuth2 authnecation with query tool open.#8299

yogeshmahajan-1903:


web/regression/feature_tests/pg_datatype_validation_test.py
```diff
@@ -75,7 +75,7 @@ def before(self):
         # For this test case we need to set "Insert bracket pairs?"
         # SQL Editor preference to 'false' to avoid codemirror
         # to add matching closing bracket by it self.
-        self._update_preferences()
+        # self._update_preferences()
```
khushboovashi, line 78: Did you disable this intentionally? If so, what's the reason? If the reason is valid, you need to remove that function from the code, too.

khushboovashi (changes requested):

web/regression/feature_tests/pg_datatype_validation_test.py
yogeshmahajan-1903: I wanted to confirm nothing break though I checked locally with one server. Hence I had commented. Delete commented code and method which is not used. This might have needed in older code mirror version,

Issue #8391, opened 2025-01-22
Enable additonal profile array response oauth2 flow (Github private email addresses)


hauserkristof:
**Current issue / limitation**

As mentioned in #7949 users cannot authenticate if the `/user` endpoint of GitHub is returning `None`. This is not an issue of the user or administrator of pgAdmin.

GitHub has a feature that enables users to [hide their email addresses](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address#about-commit-email-addresses), but it could be queried with [an authenticated request](https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-email-addresses-for-the-authenticated-user) (other providers can solve this).

The main issue is that the `user/emails` endpoint returns an array of objects, instead of a single object.

**"Solutions" that do not work**

I've tried to set the `OAUTH2_USERINFO_ENDPOINT` config value to `user/emails` with no luck

Also tried setting the `OAUTH2_ADDITIONAL_CLAIMS` to various values too, with no luck, email field is still `None`.

<details>
<summary><h3>Debug log:</h3></summary>
<pre>
255.2.3.4 - - [22/Jan/2025:20:33:49 +0000] "GET /static/js/generated/vendor.others.js?ver=81400 HTTP/1.1" 200 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/132.0.0.0 Safari/537.36"
2025-01-22 20:33:49,757: DEBUG	pgadmin:	Authentication initiated via source: oauth2
2025-01-22 20:33:49,758: DEBUG	pgadmin:	Authentication initiated via source: oauth2 is failed.
255.2.3.4 - - [22/Jan/2025:20:33:49 +0000] "POST /authenticate/login HTTP/1.1" 302 639 "https://pgadmin.xxx.yyy.dev/login?next=/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/132.0.0.0 Safari/537.36"
2025-01-22 20:33:50,758: ERROR	pgadmin:	'list' object has no attribute 'keys'
Traceback (most recent call last):
  File "/venv/lib/python3.12/site-packages/flask/app.py", line 917, in full_dispatch_request
[...]

PR #8392, opened 2025-01-22
add support for oauth2 profile arrays, empty email error log

hauserkristof:
Details are covered in #8391 

TLDR; enable handling of JSON array return types on `OAUTH2_USERINFO_ENDPOINT`s, like github's `user/emails` for supporting private email addresses.

### DIFF
--- a/web/pgadmin/authenticate/oauth2.py
+++ b/web/pgadmin/authenticate/oauth2.py
@@ -135,11 +135,38 @@ class OAuth2Authentication(BaseAuthentication):
     def validate(self, form):
         return True, None
 
+    def get_profile_dict(self, profile):
+        """
+        Returns the dictionary from profile whether it's a list or dictionary.
+        Includes additional type checking.
+        """
+        if isinstance(profile, list):
+            return profile[0] if profile else {}
+        elif isinstance(profile, dict):
+            return profile
+        else:
+            return {}
+
     def login(self, form):
         profile = self.get_user_profile()
+        profile_dict = self.get_profile_dict(profile)
+        
+        current_app.logger.debug(f"profile: {profile}")
+        current_app.logger.debug(f"profile_dict: {profile_dict}")
+        
+        if not profile_dict:
+            error_msg = "No profile data found."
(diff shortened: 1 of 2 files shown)


web/pgadmin/authenticate/oauth2.py
```diff
@@ -135,11 +135,38 @@ def get_friendly_name(self):
     def validate(self, form):
         return True, None
 
+    def get_profile_dict(self, profile):
+        """
+        Returns the dictionary from profile whether it's a list or dictionary.
+        Includes additional type checking.
+        """
+        if isinstance(profile, list):
+            return profile[0] if profile else {}
```
khushboovashi, line 144: The code looks good to me. I have a little bit of hesitation here, as the assumption made is that most of the providers will have the first profile as their primary. However, that would be complex to verify as the response will be different for the providers, as you already mentioned.

Please make the appropriate documentation changes in the `oauth2.rst` file so we can proceed with this PR.

khushboovashi (changes requested):

web/pgadmin/authenticate/oauth2.py
```diff
@@ -135,11 +135,39 @@ def get_friendly_name(self):
...
+        
+        if not profile_dict:
+            error_msg = "No profile data found."
+            current_app.logger.exception(error_msg)
+            return False, gettext(error_msg)
+        
         email_key = \
-            [value for value in self.email_keys if value in profile.keys()]
-        email = profile[email_key[0]] if (len(email_key) > 0) else None
+            [value for value in self.email_keys if value in profile_dict.keys()]
```
khushboovashi, line 164: Please fix the PEP-8 issue; you can check the workflow logs for the same; it is failing.

khushboovashi (changes requested):
Please fix the PEP 8 issue. You can check the

web/pgadmin/authenticate/oauth2.py
hauserkristof: sorry, resolved the error and the warnings too

PR #8427, opened 2025-02-07
Fix the sonarqube issues

yogeshmahajan-1903:
1.Do not use Array index in keys
2.Visible, non-interactive elements with click handlers must have at least one keyboard listener. 3.Add replacement fields or use a normal string instead of an f-string. 4.Add a parameter to the parent lambda function

### DIFF
--- a/web/pgadmin/static/js/SchemaView/FieldControl.jsx
+++ b/web/pgadmin/static/js/SchemaView/FieldControl.jsx
@@ -18,7 +18,7 @@ export const FieldControl = ({schemaId, item}) => {
     <Control {...props}>
       {
         children?.map(
-          (child, idx) => <FieldControl key={idx} item={child}/>
+          (child, idx) => <FieldControl key={`${child.controlProps.id}-${idx}`} item={child}/>
         )
       }
     </Control>, [schemaId, Control, props, children]
--- a/web/pgadmin/static/js/components/PgTree/FileTreeItem/DoubleClickHandler.jsx
+++ b/web/pgadmin/static/js/components/PgTree/FileTreeItem/DoubleClickHandler.jsx
@@ -11,11 +11,12 @@ import { useSingleAndDoubleClick } from '../../../custom_hooks';
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import CustomPropTypes from '../../../../js/custom_prop_types';
-
+import { getEnterKeyHandler } from '../../../../js/utils';
+ 
 export default function DoubleClickHandler({onSingleClick, onDoubleClick, children}){
   const onClick = useSingleAndDoubleClick(onSingleClick, onDoubleClick) ;
   return(
-    <div onClick={(e)=>onClick(e)}>
+    <div onClick={(e)=>onClick(e)} onKeyDown = { getEnterKeyHandler(onClick)}>
       {children}
     </div>
   );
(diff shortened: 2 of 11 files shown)


web/pgadmin/static/js/SchemaView/FieldControl.jsx
```diff
@@ -18,7 +18,7 @@ export const FieldControl = ({schemaId, item}) => {
     <Control {...props}>
       {
         children?.map(
-          (child, idx) => <FieldControl key={idx} item={child}/>
+          (child, idx) => <FieldControl key={`${child.controlProps.id}-${idx}`} item={child}/>
```
adityatoshniwal, line 21: No sure if `child.controlProps.id` will always be available.

web/pgadmin/static/js/components/PgTree/FileTreeItem/DoubleClickHandler.jsx
```diff
@@ -15,7 +15,7 @@ import CustomPropTypes from '../../../../js/custom_prop_types';
 export default function DoubleClickHandler({onSingleClick, onDoubleClick, children}){
   const onClick = useSingleAndDoubleClick(onSingleClick, onDoubleClick) ;
   return(
-    <div onClick={(e)=>onClick(e)}>
+    <div onClick={(e)=>onClick(e)} onKeyDown={(e)=>onClick(e)}>
```
adityatoshniwal, line 18: There is no filter on which key in onKeyDown. Generally, enter key means a click.

web/pgadmin/tools/sqleditor/static/js/components/sections/StatusBar.jsx
```diff
@@ -131,6 +131,7 @@ export function StatusBar({eol, handleEndOfLineChange}) {
           <Tooltip title="Select EOL Sequence" enterDelay={2500}>
             <span
               onClick={toggleMenu}
+              onKeyDown={toggleMenu}
```
adityatoshniwal, line 134: Blind onKeyDown is not acceptable.

adityatoshniwal (changes requested):

web/pgadmin/static/js/components/PgTree/FileTreeItem/DoubleClickHandler.jsx
```diff
@@ -15,7 +15,12 @@ import CustomPropTypes from '../../../../js/custom_prop_types';
 export default function DoubleClickHandler({onSingleClick, onDoubleClick, children}){
   const onClick = useSingleAndDoubleClick(onSingleClick, onDoubleClick) ;
   return(
-    <div onClick={(e)=>onClick(e)}>
+    <div onClick={(e)=>onClick(e)} onKeyDown={(e)=>{
```
adityatoshniwal, line 18: how about we create a wrapper like -
```
function getEnterHandler(callback) {
    return (e)=>{
        if(e.code === 'Enter'){
            e.preventDefault();
            callback(e);
         }
    }
}
```
and then `onKeyDown = { getEnterHandler(onClick) }`

adityatoshniwal (changes requested):

web/pgadmin/static/js/components/PgTree/FileTreeItem/DoubleClickHandler.jsx
yogeshmahajan-1903: Done.

PR #8440, opened 2025-02-12
Documentation: Fix wrong parameter description for Password Exec Command

mwil:
The documentation mentions that `%HOST%` can be used as a replacement parameter for the Password Exec Field in the Advanced tab of the Server Dialog, but the parameter is actually called `%HOSTNAME%`:

https://github.com/pgadmin-org/pgadmin4/blob/58bb14253e9c135ccdf871408600b85c2a44641b/web/pgadmin/utils/passexec.py#L42


rtlong:
Excited for this to be released. I just wasted 2 hours to figure this out :-/

Issue #8296, opened 2024-12-24
White screen in the Query Tool or View/Edit data when trying to edit a json type field after Paste


furje:
1. Create this simple table:

CREATE TABLE IF NOT EXISTS testtbl
(
    id smallint NOT NULL,
    value jsonb,
    CONSTRAINT testtbl_pkey PRIMARY KEY (id)
)

2. insert one record: INSERT INTO lvl2db.testtbl (id, value) VALUES (1, '{"a": "b"}');
3. Open table in View/Edit data or in Query Tool put query: select * from testtbl;
4. In data output form select row and click "copy":

![Screenshot_20241224_164922](https://github.com/user-attachments/assets/1b4dfc84-d0cf-4644-8b00-0235dfdca6e6)

5. After this, click "paste":
![Screenshot_20241224_165004](https://github.com/user-attachments/assets/435c9482-2b71-4688-835c-7918a8252180)

6. Make double click to edit field "value" in pasted row, after which the interface crashes (white screen):

![Screenshot_20241224_165228](https://github.com/user-attachments/assets/7f391fdf-59a3-4bb0-90e6-525e8ae60a63)

pgadmin4 version 8.14, commit: 9e58717321a434519875773707df78c020bcde40 2024-12-09, desktop version, browser: Chrome 130.0.6723.137, OS: Ubuntu 22.04.5 LTS, Linux-6.8.0-50-lowlatency-x86_64-with-glibc2.35

anilsahoo20:
Hi @furje 
I am not able to reproduce the issue.
Can you please share the pgAdmin logs and the screen recording of the issue.

prisoner881:
> Hi @furje I am not able to reproduce the issue. Can you please share the pgAdmin logs and the screen recording of the issue.

This may be related to a different bug I noticed, which is not consistently reproduceable.

Sometimes when I run a query on a table, the font randomly changes, or the "Scratch Pad" section background goes white, or the "Total Rows: XXX Query complete YY:YY:YY.YY" footer text color changes to a random shade of grey that occasionally matches the background color, making the output unreadable.

Repeated runs of the same query will eventually cause all to be restored to normal again, but it's completely unpredictable.

furje:
> Hi @furje I am not able to reproduce the issue. Can you please share the pgAdmin logs and the screen recording of the issue.

@anilsahoo20 , sorry for delay!

[...]

PR #8441, opened 2025-02-12
Fixed an issue where pasting JSON data type values into the Result Gr…

anilsahoo20:
…id caused a syntax error by replacing our custom CSV-to-array conversion code with the PapaParse package. #8296

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -122,6 +122,7 @@
     "moment-timezone": "^0.5.34",
     "notificar": "^1.0.1",
     "notistack": "^3.0.1",
+    "papaparse": "^5.5.2",
     "path-fx": "^2.0.0",
     "postcss": "^8.4.31",
     "rc-dock": "^3.2.9",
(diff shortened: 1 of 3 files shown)


web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/CopyData.js
```diff
@@ -8,8 +8,8 @@
 //////////////////////////////////////////////////////////////
 import JSONBigNumber from 'json-bignumber';
 import _ from 'lodash';
+import Papa from 'papaparse';
 import * as clipboard from '../../../../../../static/js/clipboard';
-import { CSVToArray } from '../../../../../../static/js/utils';
```
adityatoshniwal, line 12: lets keep `CSVToArray` as it is and change logic of `CSVToArray` to use papaparse.

adityatoshniwal (changes requested):
We also have multiple test cases on CSVToArray, so change it would test our scenarios as well.

web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/CopyData.js
anilsahoo20: Okay, I will change it

Issue #8435, opened 2025-02-10
Failed to decrypt the saved password after update to v9


yanehi:
**Description**

As soon as I try to start a saved connection to a server after updating to pgAdmin v9, I get the error message:

```
Failed to decrypt the saved password.
Error: 'utf-8' codec can't decode byte 0xab in position 0: invalid start byte
```

When I enter the password, I can log in normally. If I then try to save the password again, the same problem occurs again after logout/login to pgAdmin. We did not have this problem in `v8.14`.

**To Reproduce**

1. Go to 'Servers'
2. Click on 'Saved Server configuration to connect to the server'
3. See the error:

![Image](https://github.com/user-attachments/assets/6ae9cbd0-2c9b-4a6f-98b1-a98d46087a85)

**Expected behavior**

I can connect to the server using access data that has already been saved in pgAdmin.

**Error message**

```
Failed to decrypt the saved password.
Error: 'utf-8' codec can't decode byte 0xab in position 0: invalid start byte
```

**Setup:**
 - OS: Docker Container based on `dpage/pgadmin4`
 - Version: 9.0
 - Mode: Server
 - Package type: Docker Container

We use a database to store the user settings.

yogeshmahajan-1903:
@yanehi 
I am not able to reporduce the issue. Can you please share your docker deployment config? What are the authentications sources being used?

danielabbatt:
I have the same problem. The servers are all there in the list once I enter my master password, but each instance comes up with that error

Failed to decrypt the saved password.
Error: 'utf-8' codec can't decode byte 0xb7 in position 0: invalid start byte

yanehi:
@yogeshmahajan-1903 
We use SSO for the initial login to pgadmin. However, the problem exists when logging in to the DB server when I am already logged in to pgAdmin. After upgrading from v8 to v9, pgAdmin can no longer read or use the password already saved for logging in to the server. It still worked on pgAdmin v8.

We have only set the following environment variables: PGADMIN_DEFAULT_EMAIL, OAUTH2_CLIENT_ID, OAUTH2_CLIENT_SECRET, PGADMIN_DEFAULT_PASSWORD, PGADMIN_CONFIG_CONFIG_DATABASE_URI, PGADMIN_LISTEN_PORT.

yogeshmahajan-1903:
With SSO, I am able to reproduce the issue. Thanks for reporting.

danielabbatt:
[...]

PR #8454, opened 2025-02-13
In server mode with SSO authentication, ensure saved passwords are encrypted with correct encryption key.#8435

yogeshmahajan-1903:


khushboovashi (commented):
The changes you made were already covered in the old code, so I am not getting the fix. Can you please elaborate?

Issue #6503, opened 2023-06-27
Unable to load database node properties, getting connection lost error


nikhil-mohite:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

pgAdmin is unable to load selected database node properties and getting connection loss notification

**To Reproduce**

1. Create a Database
2. Open Query Tool for the newly created database
3. Go to the database collection node and select the 'Properties' tab 
4. Select the newly created database from the list (click on the check box)
5. Click on the 'Delete' button from the top of the table. (Don’t use Delete (Force) option)
6. Now select the same database in the object explorer tree.
7. It shows the loading spinner
8. After a few sec pgAdmin shows dialog as it is unable to drop the database as few connections are open this is expected behavior 
9. Click on the `OK` button and check it is showing one more error popup `Connection to the server has been lost`
<img width="644" alt="Screenshot 2023-06-21 at 3 03 55 PM" src="https://github.com/pgadmin-org/pgadmin4/assets/69510038/c1488113-bd90-4131-954f-abd0e73970f9">

11. Check properties tab spinner is showing and unable to load the selected database node properties

**Expected behavior**

It should show the selected database node details on `Properties` tab

**Desktop (please complete the following information):**
 - OS: macOS (Ventura 13.3)
 - Version: 7.3
 - Mode: Desktop

PR #8478, opened 2025-02-18
Fixed an issue where deleting a database from the properties tab of d…

anilsahoo20:
…atabase collections and immediately opening its properties resulted in a connection lost error. #6503

### DIFF
--- a/web/pgadmin/browser/server_groups/servers/databases/__init__.py
+++ b/web/pgadmin/browser/server_groups/servers/databases/__init__.py
@@ -1011,6 +1011,40 @@ class DatabaseView(PGChildNodeView):
 
         return False, ''
 
+    def _check_db_connections(self, sid, did, database):
+        """
+        Check the no. of connections of a specific database and if it has
+        multiple connections then don't release else release.
+        :param sid: Server Id.
+        :param did: Database Id.
+        :param database: Database Name.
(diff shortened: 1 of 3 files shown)


web/pgadmin/browser/server_groups/servers/databases/__init__.py
```diff
@@ -1011,6 +1011,40 @@ def _release_conn_before_delete(self, sid, did):
...
+        # Categorizing connections on the basis of current database connected
+        db_connections = [
+            conn for conn in connections
+            if
+            connections[conn].db == database and connections[conn].connected()
+        ]
+
+        # Extracting the other connections such as QT, ERD, or PSQL
+        # connections of current database
+        query_tool_connections = [
```
yogeshmahajan-1903, line 1034: Connections could be external from other than pgadmin too. Like psql, JDBC connection etc. You  should check all the conenctions on the database not only from the pgadmin.

yogeshmahajan-1903 (changes requested):
Please check logic to get existing external  connections too

web/pgadmin/browser/server_groups/servers/databases/__init__.py
anilsahoo20: Okay

Issue #3564, opened 2020-05-12
pgadmin doesn't work with pgbouncer database aliases (RM #5511)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/5511***
*Originally created by **Michael Vitale** at 2020-05-12 16:43:38 UTC.*

pgbouncer background info:

Here is my databases section in pgbouncer.ini:
[databases]
mydbwriter =                    port=5432 dbname=mydb auth_user=pgbouncer
mydbreader = host=mystandbyhost port=5432 dbname=mydb auth_user=pgbouncer

pgbouncer resides on the primary host, so not specifying a host means it will connect to the PG instance on the same host as pgbouncer.  By specifying host, it means pgbouncer will redirect to the standby.

Now lets setup pgadmin to connect....
I specify the alias, mydbwriter for the maintenance database.  Add all the other usual stuff, user, password port (port = 6432) which is what pgbouncer is listening on.  PG is listening on 5432.

When I hit the connect button in pgadmin, it opens up the tree view pane on the left as expected listing all the databases it found there including the real database, mydb.

BUT here is the bug:
When I click on mydb, I get an error saying "Error: no such database: mydb".

Suggested fix: have pgadmin always do this query to determine what is the real database we connected to: SELECT current_database();
Then allow the user to click on a database that matches that value.

derhuerst:
I just ran into this problem:

We have databases `gtfs_*`, of which we dynamically (during runtime of both PostgreSQL and pgbouncer) alias one as `gtfs` using pgbouncer. We do not pass through the original databases.

When using pgAdmin 4 to connect to pgbouncer's "virtual" `gtfs` database,
1. pgAdmin connects successfully,
2. but then "discovers" (enumerates) all databases using SQL, which returns the "real"/unaliased database names (which are of course not available through pgbouncer),
3. and *does not* offer the "virtual" `gtfs` database in the tree view.

This effectively means that, **as long as I alias a database using pgbouncer, I cannot connect to it using pgAdmin 4**.

Can you please consider
- offer the configured database in the tree view, as long as pgAdmin can connect to it, regardless of the result of "discovery".
- allowing users to hard-code a single database? – It is already possible to filter the list of automatically "discovered" databases, after all.

---

related Stack Exchange questions, AFAICT none of them providing a solution:

- https://stackoverflow.com/questions/55927647/no-such-database-error-on-pgadmin-when-connecting-through-pgbouncer
[...]

PR #8485, opened 2025-02-21
Allow adding and connecting the server using pgbouncer database aliases.#3564

yogeshmahajan-1903:


web/migrations/versions/320114f59d9c_.py
```diff
@@ -0,0 +1,35 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '320114f59d9c'
+down_revision = '255e2842e4d7'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('server', sa.Column('db_alias', sa.String(length=256)))
```
adityatoshniwal, line 30: when adding new column to `server` always add autoincrement
```
    with op.batch_alter_table(
            "server", table_kwargs={'sqlite_autoincrement': True}) as batch_op:
        batch_op.add_column(sa.Column('tags', sa.JSON(), nullable=True))
```

web/pgadmin/browser/server_groups/servers/__init__.py
```diff
@@ -767,6 +768,7 @@ def update(self, gid, sid):
             'name': gettext('name'),
             'port': gettext('Port'),
             'db': gettext('Maintenance database'),
+            'db_alias': gettext('Database Alias '),
```
adityatoshniwal, line 771: remove leading space.

web/pgadmin/browser/server_groups/servers/static/js/server.ui.js
```diff
@@ -315,8 +316,16 @@ export default class ServerSchema extends BaseUISchema {
...
+        mode: ['properties', 'edit', 'create'], readonly: obj.isConnected,
+        disabled: function(state) {
+          return !(isEmptyString(state.db_alias) && obj.isShared);
+        },
+      },{
+        id: 'db_alias', label: gettext('Database alias'), type: 'text', group: gettext('Connection'),
+        mode: ['properties', 'edit', 'create'], readonly: obj.isConnected,
+        disabled: function(state) {
+          return !isEmptyString(state.db);
+        },
```
adityatoshniwal, line 328: We should add help message here - `This is helpful in pg_bouncer where server can be connected using database alias name`

adityatoshniwal (changes requested):

akshay-joshi:
As per discussion with Dave, this PR is not required anymore, hence closing it.

Issue #4194, opened 2021-03-04
Option to automatically open a downloaded CSV (RM #6290)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/6290***
*Originally created by **Kay Khandpur** at 2021-03-04 17:41:03 UTC.*

Prior to v5.0 a download of query results would open the CSV (browser permitting).

With 5.0 in its own container now, the CSV is downloaded and it takes an extra step to open it.

Would be great to have a configurable option to either 1) download the CSV only or 2) download and open in selected app.

Thanks.

yogeshmahajan-1903:
This will be taken care after moving to electron.

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-03-11/
Package: arm64, pip wheel
Environment: macOs Sonoma 14.7.1

PR #8517, opened 2025-03-04
Added support for automatically opening a file after it is downloaded in desktop mode. #4194

pravesh-sharma:


akshay-joshi (changes requested):
1) Disable the preferences option for File Downloads in server mode.
2) Update the documentation and screenshots.

anthonydb:
@akshay-joshi  Hi, is this functionality going to be enabled by default? If so, I would suggest it be shut off by default. Automatically opening CSV files with Excel, for example, can destroy data. Any US Zip code such as `07031` is automatically converted by Excel to `7031`. Unsuspecting users will not know this. Thank you for considering.

akshay-joshi:
> @akshay-joshi Hi, is this functionality going to be enabled by default? If so, I would suggest it be shut off by default. Automatically opening CSV files with Excel, for example, can destroy data. Any US Zip code such as `07031` is automatically converted by Excel to `7031`. Unsuspecting users will not know this. Thank you for considering.

By Default this setting will be disabled.

akshay-joshi (changes requested):
1) Take the screenshot again with the proper help message. The help message looks old.

Issue #8449, opened 2025-02-12
Disabled icons need tooltips


usernamesareaholdoverfromthe90s:
**Describe the solution you'd like**

When you hover your mouse over a enabled icon, you get a tooltip.

But when you hover your mouse over a disabled icon, you don't get a tooltip.

This makes it difficult to discover functionality.

All icons should always have a tooltip.  But, if you want to make your chunky black tooltips a grey background to underscore that they're disabled, that's fine.  But there should be a tooltip.

See attached video, that shows me not being able to figure out if a couple of icons might (or might not) be useful to me.

https://github.com/user-attachments/assets/26b5e033-0298-4f8d-ab92-029fb83f4cee

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-03-10/
Package: arm64
Environment: macOs Sonoma 14.7.1

PR #8518, opened 2025-03-04
Show tooltips when icons are disabled.#8449

RohitBhati8269:
Detail Info :- #8449

### DIFF
--- a/web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
+++ b/web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
@@ -73,8 +73,10 @@ export default function ObjectExplorerToolbar() {
     <Box display="flex" alignItems="center" gap="2px">
       <PgButtonGroup size="small">
         <ToolbarButton icon={<QueryToolIcon />} menuItem={menus['query_tool']} shortcut={browserPref?.sub_menu_query_tool} />
-        <ToolbarButton icon={<ViewDataIcon />} menuItem={menus['view_all_rows_context']} shortcut={browserPref?.sub_menu_view_data} />
-        <ToolbarButton icon={<RowFilterIcon />} menuItem={menus['view_filtered_rows_context']} />
+        <ToolbarButton icon={<ViewDataIcon />} menuItem={menus['view_all_rows_context'] ?? 
+          {label :'All Rows'}}
+        shortcut={browserPref?.sub_menu_view_data} />
+        <ToolbarButton icon={<RowFilterIcon />} menuItem={menus['view_filtered_rows_context'] ?? { label :'Filtered Rows...'}} />
         <ToolbarButton icon={<SearchOutlinedIcon style={{height: '1.4rem'}} />} menuItem={menus['search_objects']} shortcut={browserPref?.sub_menu_search_objects} />
         {!_.isUndefined(menus['psql']) && <ToolbarButton icon={<TerminalRoundedIcon style={{height: '1.4rem'}}/>} menuItem={menus['psql']} />}
       </PgButtonGroup>
--- a/web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
+++ b/web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
@@ -18,7 +18,7 @@ import _ from 'lodash';
 import PropTypes from 'prop-types';
 import CustomPropTypes from '../custom_prop_types';
 import usePreferences from '../../../preferences/static/js/store';
-
+import gettext from 'sources/gettext';
 
 function ToolbarButton({menuItem, ...props}) {
   return (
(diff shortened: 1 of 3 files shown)


web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
```diff
@@ -42,11 +42,18 @@ export default function ObjectExplorerToolbar() {
...
-      menuItems?.
-      find((m)=>(m.name=='view_data'))?.
-      menu_items;
+    // Find the 'object' menu and get its menuItems & 
+    // Try to find 'view_data' in the object menu and get its menu_items
+    let objectMenu = pgAdmin.Browser.MainMenus.find((m) => (m.name == 'object'))?.menuItems;
+    let viewMenus = objectMenu?.find((m) => (m.name == 'view_data'))?.menu_items;
+
+    // If view_data not found in object menu or it doesn't have menu_items
+    if (_.isUndefined(viewMenus)) {
```
adityatoshniwal, line 51: Since when tools menu has started adding view data menu?

web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
```diff
@@ -42,11 +42,18 @@ export default function ObjectExplorerToolbar() {
...
   const checkMenuState = ()=>{
-    const viewMenus = pgAdmin.Browser.MainMenus.
-      find((m)=>(m.name=='object'))?.
-      menuItems?.
-      find((m)=>(m.name=='view_data'))?.
-      menu_items;
+    // Find the 'object' menu and get its menuItems & 
+    // Try to find 'view_data' in the object menu and get its menu_items
+    let objectMenu = pgAdmin.Browser.MainMenus.find((m) => (m.name == 'object'))?.menuItems;
+    let viewMenus = objectMenu?.find((m) => (m.name == 'view_data'))?.menu_items;
```
adityatoshniwal, line 48: Why change this?

web/regression/javascript/file_manager/FileManager.spec.js
```diff
@@ -162,7 +162,7 @@ describe('FileManger', ()=>{
 
       await user.click(ctrl.container.querySelector('[name="menu-shared-storage"]'));
       await user.click(ctrl.container.querySelector('[data-label="Shared Storage"]'));
-      expect(ctrl.container.querySelector('button[aria-label="Shared Storage"]')).not.toBeNull();
+      expect(ctrl.container.querySelector('button[aria-label="Shared Storage"]')).toBeNull();
```
adityatoshniwal, line 165: This is not the right way. aria-label should be added button for accessibility reasons.

adityatoshniwal (changes requested):

web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
```diff
@@ -73,8 +73,10 @@ export default function ObjectExplorerToolbar() {
     <Box display="flex" alignItems="center" gap="2px">
       <PgButtonGroup size="small">
         <ToolbarButton icon={<QueryToolIcon />} menuItem={menus['query_tool']} shortcut={browserPref?.sub_menu_query_tool} />
-        <ToolbarButton icon={<ViewDataIcon />} menuItem={menus['view_all_rows_context']} shortcut={browserPref?.sub_menu_view_data} />
-        <ToolbarButton icon={<RowFilterIcon />} menuItem={menus['view_filtered_rows_context']} />
+        <ToolbarButton icon={<ViewDataIcon />} menuItem={menus['view_all_rows_context'] ?? 
+          {label :'All Rows'}}
```
adityatoshniwal, line 77: why?

web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
```diff
@@ -73,8 +73,10 @@ export default function ObjectExplorerToolbar() {
...
+        shortcut={browserPref?.sub_menu_view_data} />
+        <ToolbarButton icon={<RowFilterIcon />} menuItem={menus['view_filtered_rows_context'] ?? { label :'Filtered Rows...'}} />
```
adityatoshniwal, line 79: why?

web/regression/javascript/file_manager/FileManager.spec.js
```diff
@@ -153,16 +153,16 @@ describe('FileManger', ()=>{
...
+      const user = userEvent.setup(); 
       await act(async ()=>{
         ctrl = await ctrlMount({});
       });
-
+    
       await user.click(ctrl.container.querySelector('[name="menu-shared-storage"]'));
       await user.click(ctrl.container.querySelector('[data-label="Shared Storage"]'));
-      expect(ctrl.container.querySelector('button[aria-label="Shared Storage"]')).not.toBeNull();
+      expect(ctrl.container.querySelector('button[data-label="Shared Storage"]')).not.toBeNull();
```
adityatoshniwal, line 165: I said aria-label should be there for accessibility reasons but you completely replaced it. This is in not the right way.

web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
RohitBhati8269: I needed to add this condition explicitly because when no object is selected and we hover over icons like the query tool, the adjacent 'All Rows' and 'Filtered Rows' buttons don't show their labels - they only display their shortcuts. This makes it difficult for users to understand the button's purpose. By adding this explicit condition, we ensure that the button labels will be properly displayed alongside their shortcuts when hovering, improving the toolbar's usability.

web/regression/javascript/file_manager/FileManager.spec.js
RohitBhati8269: I'll keep the element selection working as intended by using both attributes together. I'll add both aria-label and data-label so we maintain accessibility while ensuring our tests can properly select the elements. Thank you for highlighting this important accessibility concern.

web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
RohitBhati8269: same as below.

web/regression/javascript/file_manager/FileManager.spec.js
adityatoshniwal: Why add data-label if aria-label is already there? Testing is already been taken care with aria-label.

web/pgadmin/static/js/helpers/ObjectExplorerToolbar.jsx
adityatoshniwal: In that case, you should wrap the text in gettext.

web/regression/javascript/file_manager/FileManager.spec.js
RohitBhati8269: With aria-label, the querySelector did not get the element, which caused test cases to fail. That's why I need to change it to use data-label instead.

adityatoshniwal: Which means aria-label is not added to the button. A button which doesn't have any label should have aria-label for accessibility.

adityatoshniwal: Please read -https://github.com/mui/material-ui/issues/33182. It will give you more insight.

PR #8519, opened 2025-03-05
Add option to load/replace servers.json every boot

LauJosefsen:
Hi.

This PR adds the env `PGADMIN_REPLACE_SERVERS_ON_STARTUP` do the Docker image. 

If this env is set to true, the servers.json will be loaded on every container boot, with the `--replace` option.

The motivation for this change is to be able to define servers as part of the deployments / infrastructure as code instead of  manually manipulating the pgadmin database state. 

This should not introduce any breaking change, and will only differ in execution if `PGADMIN_REPLACE_SERVERS_ON_STARTUP` is set to true.

I have tested the image locally, with both `PGADMIN_REPLACE_SERVERS_ON_STARTUP` undefined and with it set to true. I have also tested both cases on an initialized database and on a non-initialized database.


firecow:
Uh, nice :D 

I could use this.

I'm using a different IoC db row seeding solution at the moment.

yogeshmahajan-1903 (changes requested):
Please update documentation for the same [here](https://www.pgadmin.org/docs/pgadmin4/latest/container_deployment.html).

LauJosefsen:
> Please update documentation for the same [here](https://www.pgadmin.org/docs/pgadmin4/latest/container_deployment.html).

I tried adding some documentation below the PGADMIN_SERVER_JSON_FILE definition, as it seems related. I dont know if it should be moved?

yogeshmahajan-1903 (approved):

Issue #8473, opened 2025-02-18
Different icon should be used to avoid confusion


ArqamKarim:
Hi Team, I have a simple suggestion, when I am backing up anything in PG Admin Desktop version, I see this cross mark icon, this usually implies that a process has been failed, instead stop icon should be used (a red square to stop the process, not a 'failed' icon.

Here's the screenshot of the same.

![Image](https://github.com/user-attachments/assets/2151472f-e52f-40fb-9af4-193699d6701d)

anil-unni:
yes

pravesh-sharma:
Issue fixed, verified on the latest snapshot build.

PR #8521, opened 2025-03-05
Replace cancel icon with stop icon in backup process.#8473

RohitBhati8269:
Detail Info :- #8473


adityatoshniwal (changes requested):
processes screenshot doesn't look right. please check once.

Issue #8355, opened 2025-01-13
No garbage collection for session files


cbr46:
**Describe the bug**

We run pgadmin4 as a container in a kubernetes cluster. We monitor the availability of pgadmin through regular checks on the main website. After about one month the directory /var/lib/pgadmin/sessions lists over 400 mb of session files. These should be cleaned up.

**To Reproduce**

1. Monitor pgadmin.localhost
2. Monitor disk use of volume /var/lib/pgadmin/

**Expected behavior**

Regular cleanup or garbage collection in the /var/lib/pgadmin/sessions directory

**Error message**

No error.

**Screenshots**

No screenshots.

**Desktop (please complete the following information):**
 - OS: Image dpage/pgadmin4:8.14
 - Version: 8.14
 - Mode: Server
 - Browser (if running in server mode): -
 - Package type: Container

**Additional context**

-

adityatoshniwal:
Hi @cbr46,
The cleanup API is called every 5 minutes which takes care of cleaning up session files which are older than `CHECK_SESSION_FILES_INTERVAL` (default to 24hrs)
https://github.com/pgadmin-org/pgadmin4/blob/36df6d5c8af70498971ee5b080f7fe1572c8290d/web/pgadmin/misc/__init__.py#L173

adityatoshniwal:
> It looks like that this function is not called. Should there be a log entry when this function is called?

Can you please check pgAdmin browser network if `cleanup` API is called in interval of 5 mins?

adityatoshniwal:
> It looks like the cleanup function is only called when an active session runs in a browser. As soon as the session ends, the cleanup is not triggered. Since we monitor the availability of this service, those short connections create a session file that is not cleaned up automatically.

@cbr46 What is the endpoint are you using for monitoring? pgAdmin provides `/misc/ping` for health check. It doesn't create any session file.

adityatoshniwal:
@cbr46 There won't be any session files if there no active users logged in :)
Please feel free to re-open if issue persists.

adityatoshniwal:
> [@adityatoshniwal](https://github.com/adityatoshniwal) The session files are still created when you open https://pgadmin.local and don't sign in. Thank you!

Yeah, but people must have installed pgAdmin to use it 😄  and not just open login screen.

adityatoshniwal:
Reopening for testing.

yogeshmahajan-1903:
This is working fine.

[...]

Issue #1999, opened 2018-08-06
No Session cleanup (RM #3550)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/3550***
*Originally created by **Martin Zeug** at 2018-08-06 11:03:50 UTC.*

We use pgadmin4 3.1 in docker (https://hub.docker.com/r/dpage/pgadmin4/) and got a sessionfolder auf 14GB.
After deleting, using pgadmin for 10 Minutes gives 1500 new session files.
We use pgadmin 4 behind a haproxy with SSL.
is there any cleanup mechanism?

Greetings,
Martin

dpage:
Issue closed on Redmine.

PR #8526, opened 2025-03-07
Delete expired session files when there is no traffic

cbandy:
Most HTTP requests create a file in the sessions directory. Open browsers call `/misc/cleanup` every five minutes, but session files can accumulate after browsers are closed.

This wakes periodically to delete expired sessions according to the `CHECK_SESSION_FILES_INTERVAL` setting.

Fixes: pgadmin-org/pgadmin4#1999
Fixes: pgadmin-org/pgadmin4#8355

### DIFF
--- a/web/pgadmin/misc/__init__.py
+++ b/web/pgadmin/misc/__init__.py
@@ -25,6 +25,7 @@ from pgadmin.utils.ajax import precondition_required, make_json_response, \
 from pgadmin.utils.heartbeat import log_server_heartbeat, \
     get_server_heartbeat, stop_server_heartbeat
 import config
+import threading
 import time
 import json
 import os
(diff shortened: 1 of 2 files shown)


adityatoshniwal:
Hi @cbandy,
There shouldn't be any http request when the browser is closed then how are session files accumulated?
The health check URL `/misc/ping` doesn't generate any session file.

cbandy:
> There shouldn't be any http request when the browser is closed then how are session files accumulated?

We run in server mode, so a request can come from anywhere on our intranet. We trust most things not to be malicious there, but there's no guarantee.

Here's what my idle (no recent user activity) pgAdmin sessions directory looks like before and after one OWASP scan:

```console
$ du -sh /var/lib/pgadmin/sessions
44.0K   /var/lib/pgadmin/sessions

$ du -sh /var/lib/pgadmin/sessions
19.4M   /var/lib/pgadmin/sessions
```

Running a more intensive scan or scanning daily would quickly fill this disk.

adityatoshniwal:
In that case, I think `cleanup_session_files` can be removed from `/misc/cleanup` as it will be taken care by the new code.

cbandy:
> In that case, I think `cleanup_session_files` can be removed from `/misc/cleanup` as it will be taken care by the new code.

Done! I also moved the timer into a `try..finally` so it will recur if there is trouble with the session directory temporarily.

Issue #392, opened 2016-06-27
Font choices (RM #1411)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/1411***
*Originally created by **Melvin Davidson** at 2016-06-27 19:58:44 UTC.*

The previous version (PgAdmin3) had the ability to specify the font for the Query Tool and the Database Designer.
This is currently not available in Beta 2 of PgAdmin4 and is essential because the current fonts as they appear 
in Windows 10 are very small.

pravesh-sharma:
Found the following behaviour when testing:

If we provide a font which is not installed then it is not falling back to default `Source Code Pro` font.

pravesh-sharma:
Verified on latest source code and is working fine.

PR #8553, opened 2025-03-14
Ability to configure font family in SQL editors

MikeChristensen:
Decided to take a stab at #392 since I've been wanting to use [Nerd Fonts](https://www.nerdfonts.com/) in the SQL editor as I do with my other IDEs.  Should be a fairly straight forward change.

## New Font Family Option in Editor Preferences:

image

## SQL Editor:

image

## Default Font

If the Font family setting is blank (or the settings were upgraded from previous version of pgAdmin), it will default to `Source Code Pro`, which is the current default (and only) font:

image

## Localization

Since there's some new strings (Such as `"Font family"` in the UI, these would need to be localized into the various languages pgAdmin supports.  From what I can tell, running `make messages` would use pybabel to pull out all the strings into the various translation files.  When I ran this locally, it found all sorts of new stuff and wrong line numbers of existing stuff.  This leads me to believe that there's probably some sort of "localization pass" before new versions are released to the public.  However, if there's anything I should do with this PR, let me know.

## Ligatures

For some reason, I was not able to get ligatures to display in Codemirror even if the font supports it.  Chome will render them correctly by default using a font that supports them, so I came to the conclusion Codemirror does something that breaks them.  Using the developer tools and searching the Codemirror source code and issue tracker didn't shed any light on the reason.  I figure most people don't want or need coding ligatures in SQL so it's probably not a big deal.  I wouldn't mind figuring out how to get them working, and also provide the user a way to disable them using the `font-variant-ligatures: none;` CSS property.  I might take another stab at this when I have some time, but as I said it's probably not a huge priority.


web/pgadmin/tools/sqleditor/utils/query_tool_preferences.py
```diff
@@ -390,6 +390,13 @@ def register_query_tool_preferences(self):
         )
     )
 
+    self.sql_font_family = self.preference.register(
+        'Editor', 'sql_font_family',
+        gettext("Font family"), 'text', 'Source Code Pro',
+        category_label=PREF_LABEL_DISPLAY,
+        help_str=gettext('The font family to use for the SQL text boxes and editors.')
```
adityatoshniwal, line 397: help str could be  `Specify the font family to be used for all SQL editors. The specified font should already be installed on your system. If the font is not found, the editor will fall back to the default font, Source Code Pro.`

adityatoshniwal (changes requested):
Please update the documentation along with screenshots.

web/pgadmin/tools/sqleditor/utils/query_tool_preferences.py
akshay-joshi: Also fix the PEP8(Python Style) issue.

akshay-joshi (changes requested):

web/pgadmin/tools/sqleditor/utils/query_tool_preferences.py
MikeChristensen: Will do!  It'll be a couple days, I recently switched over to NixOS and I'm still getting everything setup again..  I'll get back to this ASAP.

adityatoshniwal:
> Please update the documentation along with screenshots.

@MikeChristensen Please update docs as well. Thanks for the PR.

web/pgadmin/tools/sqleditor/utils/query_tool_preferences.py
MikeChristensen: Looks like the Python Style issue was just around a line being too long, so I fixed this to look like the other lines.  I've updated it to use your suggested help str as well.

MikeChristensen:
> > Please update the documentation along with screenshots.
> 
> @MikeChristensen Please update docs as well. Thanks for the PR.

Okay, I'm pretty sure I did that right.  Lemme know if any more changes are needed.  Thanks for taking the time to review, this is my first PR.

Issue #4503, opened 2021-08-13
sql to be executed upon connection (RM #6681)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/6681***
*Originally created by **Michael Shapiro** at 2021-08-13 13:43:05 UTC.*

I'd like to request that PgAdmin support having SQL executed after connecting to a Postgresql database.

An example of the sql I would use this for is to have client_min_messages set:

<pre>
set client_min_messages='DEBUG';
</pre>

PR #8571, opened 2025-03-18
Added support for post-connection SQL execution, which will be run automatically on each connection made to any database of the server. #4503

akshay-joshi:


web/pgadmin/utils/driver/psycopg3/connection.py
```diff
@@ -600,9 +600,19 @@ def _initialize(self, conn_id, **kwargs):
 
         self._set_server_type_and_password(kwargs, manager)
 
+        # Execute post connection SQL if provided in the server dialog
+        errmsg = None
```
khushboovashi, line 604: Can we create a separate function for this functionality like other functions, for example, _set_server_type_and_password? I know It is hardly 4 to 5 lines of code, but for better readability, we can write this function logic in a new function.

khushboovashi (changes requested):

Issue #8574, opened 2025-03-18
Open User management in tab instead of dialog


adityatoshniwal:
1. Open user management in a closable tab.
2. Each user can be edited individually in a separate pop up dialog.
3. New user can be created in a pop up dialog.
4. Add checkboxes to users table to delete multiple users.
5. UI/UX improvement.

pravesh-sharma:
Working fine. Tested on latest snapshot build.

PR #8578, opened 2025-03-20
Open user management in a tab instead of dialog for better UI/UX. #8574

adityatoshniwal:


web/pgadmin/tools/user_management/static/js/Users.jsx
```diff
@@ -0,0 +1,293 @@
...
+        updateList();
+      } catch (error) {
+        pgAdmin.Browser.notifier.error(parseApiError(error));
+      }
+      setLoading('');
+    };
+
+    pgAdmin.Browser.notifier.confirm(gettext('Delete User'), gettext('Are you sure you want to delete the user %s?', row.original.username),
+      async () => {
+        setLoading('Deleting user...');
```
akshay-joshi, line 96: The string should be wrapped in gettext. Please check.

akshay-joshi (changes requested):
GUI Review Comments:
1) The email field should be disabled for the currently logged-in user, as the email should not be updated.
2) Active and Locked should be toggle control instead of the 'Yes/No' text in the cell, as for consistency with other places.
3) New Password and Confirm Password should be disabled for other than 'internal' mode.
4) If the username is not allowed to modify then make it disabled, this behavior is present in the old logic as well, but we need to fix.
5) Do we need a refresh button as creating/deleting the users is instant?
6) Change ownership dialog showing "undefined shared server owned by this user" it should show the count of the shared server.

adityatoshniwal:
> GUI Review Comments:
> 
> 1. The email field should be disabled for the currently logged-in user, as the email should not be updated.
> 2. Active and Locked should be toggle control instead of the 'Yes/No' text in the cell, as for consistency with other places.
> 3. New Password and Confirm Password should be disabled for other than 'internal' mode.
> 4. If the username is not allowed to modify then make it disabled, this behavior is present in the old logic as well, but we need to fix.
> 5. Do we need a refresh button as creating/deleting the users is instant?

There can be multiple admins which many change users. As it is a tab which can remain open for a long time, its good to have a refresh button.

> 6. Change ownership dialog showing "undefined shared server owned by this user" it should show the count of the shared server.

All other issues are fixed.

Issue #8572, opened 2025-03-18
CTRL+A on value bug


xWildFire:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

When you click on any column value, and then want to select the whole text by pressing ctrl+a its selecting all rows.

**To Reproduce**

1. Go to any database table
2. Click on any column value
3. Press CTRL+A

**Expected behavior**

Select only column value, when the popup is open.

**Screenshots**

![Image](https://github.com/user-attachments/assets/3a9ab2f5-6947-449e-bc00-7b7c183fc1bb)

![Image](https://github.com/user-attachments/assets/b368c218-a3bb-4c75-a728-ac23f3ec88e3)

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-04-14/
Package: arm64
Environment: macOs Sonoma 14.7.1

PR #8615, opened 2025-03-31
Prevent Ctrl/cmd+A from selecting all rows when editing a cell. #8572

RohitBhati8269:
Detail info :- #8572


adityatoshniwal (changes requested):
This doesn't work for integer/in place edit cells

web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
```diff
@@ -311,6 +315,10 @@ export function NumberEditor({row, column, onRowChange, onClose}) {
     return false;
   };
   const onKeyDown = (e)=>{
+    // Handle "Ctrl + A" or "Cmd + A".
+    if ((e.ctrlKey || e.metaKey) && e.keyCode === 65) {
```
adityatoshniwal, line 319: We cannot keep on adding this on each place. There has to be some generic fix.

adityatoshniwal (changes requested):

PR #8624, opened 2025-04-01
Some readme improvements

MikeChristensen:
A few readme updates for those setting up pgAdmin for the first time.


yogeshmahajan-1903:
As this readme is intended for setting up pgadmin dev enviroment, adding how to setup postgres will not be helpful.

PR #8626, opened 2025-04-02
Configurable ligature support

MikeChristensen:
Now that fonts are configurable in the SQL editor, why not ignite one of the most controversial debates in the programming community since tabs vs. spaces.  That's right, it's coding ligatures in the code editor.

I'll admit I'm a big fan of ligatures and enable them in any IDE that supports them (terminal emulators like Warp and Ghostty, IDEs like VSCode and Rider, and text editors such as Sublime - in my world they all have ligatures).  However, I'll admit they _might_ be a bit annoying in SQL.  Then it occurred to me, pgAdmin is not just about SQL.  It's also used for languages supported by Postgres functions written in pgSQL, PERL, Python as well as all the crazy extensions people have created such as Javascript (V8).  There's even [pl/Rust](https://plrust.io) because of course there is.

This PR enables ligature support in the SQL editor, but the feature is disabled by default so ligature enthusiasts such as myself will have to go looking for it.

No biggie if this doesn't make it into 9.2, and I'm open to any comments or feedback.

## Enabled Ligatures

image

image

## Disabled Ligatures

image

image

* Note: Yes, the Python function makes no sense I was just adding in some ligatures.


web/pgadmin/static/js/Theme/overrides/codemirror.override.js
```diff
@@ -14,6 +14,7 @@ export default function cmOverride(theme) {
       height: '100%',
       color: editor.fg,
       backgroundColor: editor.bg,
+      letterSpacing: '0px',
```
adityatoshniwal, line 17: is this needed?

adityatoshniwal (changes requested):
Nice work!

web/pgadmin/static/js/Theme/overrides/codemirror.override.js
MikeChristensen, line 17: I believe something (Perhaps Babel?) is setting the `letter-spacing` on the BODY too high, which causes ligatures not to render.  This took me [forever to figure out](https://stackoverflow.com/questions/79504725/ligatures-does-not-show-up-in-codemirror-editor-using-font-that-supports-them/79549706#79549706), but this sets the letter spacing to normal (which is zero) within the code editor.

MikeChristensen: Yea, without it the `letter-spacing` gets set too high and ligatures won't render.  This took me [a while to figure out](https://stackoverflow.com/questions/79504725/ligatures-does-not-show-up-in-codemirror-editor-using-font-that-supports-them/79549706#79549706)

adityatoshniwal (approved):

Issue #2256, opened 2019-01-12
Support type constructors for PostGIS spatial types (RM #3879)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/3879***
*Originally created by **Robert Hewlett** at 2019-01-12 03:04:44 UTC.*

When using the UI to add a spatial column (geometry | geography) one cannot set the specific type and SRID.
For character varying one can set a length e.g. character varying(30)
For numeric one can set a scale and precision e.g. numeric(17,10)
However, for geometry one can only say geometry()
Instead it would be much better if one could, for example, say
geometry(polygon, 3005)
geometry(point, 26910)
BTW LOVE the geometry viewer!

anilsahoo20:
@pravesh-sharma  please check
1. ERD tool> create table> geometry/geography data type does not loads options
2. ERD tool> drag n drop a already created table table> geometry/geography data type does not loads value
3. Edit dialog>change geometry type or srid value of geometry/geography column>save is enabled and msql is empty
4. Edit dialog of geometry/geography column not able to change geometry type and srid value
5. Schema diff with same table name and diff DDL, geometry and geography ’s geometry type and srid are not coming in the difference section, it must come like numeric data type length and precision- generate script also shows same as difference section
<img width="1673" alt="Image" src="https://github.com/user-attachments/assets/21c301f1-23c3-48cd-8294-ebc27bc033e4" />

<img width="1150" alt="Image" src="https://github.com/user-attachments/assets/e81a32a3-a40f-4506-b3aa-34b836de7c35" />

<img width="1146" alt="Image" src="https://github.com/user-attachments/assets/9a27daad-9e38-4b65-b36c-0adffafbbdd1" />

<img width="904" alt="Image" src="https://github.com/user-attachments/assets/4f4435ec-2145-492b-8654-12a2446d5bd9" />
<img width="663" alt="Image" src="https://github.com/user-attachments/assets/28aee201-8650-40f0-a0b7-28bc55b70c80" />

rhewy:
Hi,

Just tried it out!

Thank you!

💖

PR #8656, opened 2025-04-14
Added support for type constructors for PostGIS spatial types. #2256

pravesh-sharma:


web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/utils.py
```diff
@@ -474,3 +475,12 @@ def parse_column_variables(col_variables):
             k, v = spcoption.split('=')
             spcoptions.append({'name': k, 'value': v})
     return spcoptions
+
+
+def fetch_geometry_details(data):
+    if 'typname' in data and (data['typname'] in ('geometry', 'geography')):
+        match_obj = re.search(r'([a-zA-Z]+),(\d+)', data['cltype'])
```
adityatoshniwal, line 482: looks like this logic is repeating in columns as well. We can have something like parseArgs which will return an array of args.

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/tests/utils.py
```diff
@@ -593,3 +594,75 @@ def get_table_id(server, db_name, table_name):
...
+    try:
+        connection = utils.get_db_connection(self.db_name,
+                                             self.server['username'],
+                                             self.server['db_password'],
+                                             self.server['host'],
+                                             self.server['port'],
+                                             self.server['sslmode'])
+        pg_cursor = connection.cursor()
+        # Check if postgis extension is available to install
+        pg_cursor.execute('''SELECT COUNT(*) FROM pg_available_extensions WHERE
```
adityatoshniwal, line 613: No need to check if extension exists. The SQL is already doing that "IF NOT EXISTS"

web/pgadmin/browser/server_groups/servers/databases/schemas/utils.py
```diff
@@ -192,6 +192,25 @@ def get_types(self, conn, condition, add_serials=False, schema_oid=''):
 
         return True, res
 
+    def get_geometry_types(self, conn):
+        types = []
+        sql = "SELECT * FROM pg_extension WHERE extname = 'postgis';"
```
adityatoshniwal, line 197: You've checked for count everywhere but here it is diff. Probably create a function to check extension exists and pass the conn object and extname as args and re-use it.

adityatoshniwal (changes requested):
I didn't see changes in git workflows to install PostGIS extension. Otherwise, your test cases would never run.

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/tests/utils.py
pravesh-sharma: The reason for this check is if the extension is not available, then running create extension will raise an exception. "IF NOT EXISTS" checks whether the extension is already created or not.

web/pgadmin/browser/server_groups/servers/databases/schemas/utils.py
pravesh-sharma: Okay.

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/utils.py
pravesh-sharma: Regular expression in this case is different. However, I can check if we can have a common function.

pravesh-sharma:
> I didn't see changes in git workflows to install PostGIS extension. Otherwise, your test cases would never run.

Done

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/utils.py
```diff
@@ -477,10 +480,19 @@ def parse_column_variables(col_variables):
...
+    """
+    This function will fetch length and precision details
+    from fulltype.
+
+    :param fulltype: Full type.
+    :param data: Data.
+    """
+
+    match_obj = None
+    regex = r'\((\d*)([a-zA-Z]*),?(\d*)\)'
```
adityatoshniwal, line 493: why add 3rd group?

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/utils.py
```diff
@@ -477,10 +480,19 @@ def parse_column_variables(col_variables):
...
+    if data['typname'] in ('geometry', 'geography'):
```
adityatoshniwal, line 494: I don't understand why this is added. It kills the purpose of separating out the params parsing part.

web/pgadmin/utils/__init__.py
```diff
@@ -984,3 +984,9 @@ def get_safe_post_logout_redirect():
         if url.startswith(item):
             return url
     return url_for('security.login')
+
+
+def check_extension_exists(conn, extension_name):
```
adityatoshniwal, line 989: If the function name is check_extension_exists then it should result should be either true or false.

adityatoshniwal (changes requested):

adityatoshniwal (approved):

Issue #8608, opened 2025-03-28
Data change in result grid of query tool returns Saved successfully after server disconnection but returns old data on re-execution


anilsahoo20:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Data change in result grid of query tool returns Saved successfully after server disconnection but returns old data on re-execution of same query.

**To Reproduce**

1. Open QT or View Edit Data for a table(select query)
2. disconnect the server
3. change the data from data output panel
4. save tha data
5. showing saved successfully
6. re-execute query
7. showing old output
**Expected behavior**

Must show updated data.

**Screenshots**

1st row, id=3 col value changed here for reference.

<img width="564" alt="Image" src="https://github.com/user-attachments/assets/5cc20a82-d8c9-41d9-817a-eaa8e6eb2ed0" />
<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/61f2b246-93a6-47c5-8f0c-eb5c737090f0" />
<img width="953" alt="Image" src="https://github.com/user-attachments/assets/4c3156be-c0a0-400f-bb65-c049dab7c31f" />

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-05-14/
Package: arm64
Environment: macOs Sonoma 14.7.1

PR #8682, opened 2025-04-22
Handle query execution on View/Edit Data when server is disconnected with server reconnect option. #8608

RohitBhati8269:
Detail Info:-  #8608

### DIFF
--- a/web/pgadmin/tools/sqleditor/__init__.py
+++ b/web/pgadmin/tools/sqleditor/__init__.py
@@ -826,6 +826,9 @@ def start_view_data(trans_id):
     if not status and error_msg and type(error_msg) is Response:
         return error_msg
 
+    # Check if connect is passed in the request.
+    connect = 'connect' in request.args and request.args['connect'] == '1'
+
     # get the default connection as current connection which is attached to
     # trans id holds the cursor which has query result so we cannot use that
     # connection to execute another query otherwise we'll lose query result.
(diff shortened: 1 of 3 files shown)


khushboovashi (commented):
@RohitBhati8269, can you verify that these changes work with the transaction status set to On from preferences?

RohitBhati8269:
> @RohitBhati8269, can you verify that these changes work with the transaction status set to On from preferences?

Yes checked working fine.

web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -825,6 +825,9 @@ def start_view_data(trans_id):
     if not status and error_msg and type(error_msg) is Response:
         return error_msg
 
+    # Check if connect is passed in the request.
```
khushboovashi, line 828: The code should work without considering the connect request argument because it's been handled in the check_transaction_status function.

Please check other instances of this type of code which you have introduced.

web/pgadmin/tools/sqleditor/__init__.py
```diff
@@ -841,19 +844,21 @@ def start_view_data(trans_id):
...
     if not default_conn.connected():
-        # This will check if view/edit data tool connection is lost or not,
-        # if lost then it will reconnect
-        status, error_msg, conn, trans_obj, session_obj, response = \
-            query_tool_connection_check(trans_id)
-        # This is required for asking user to enter password
-        # when password is not saved for the server
-        if response is not None:
-            return response
+        if connect:
```
khushboovashi, line 847: I think this condition is not required, the original code will work.

khushboovashi (changes requested):

web/pgadmin/tools/sqleditor/__init__.py
RohitBhati8269: I looked into it and found it's dependent on Anil’s recent changes. We’ve synced up — I’ll update my PR once his final changes are in.

Issue #8149, opened 2024-11-19
Reconnect option for server or database


ohelit:
Sometimes in pgadmin in Docker when you left the application running and there is no activity for some time, the server connection drops (we use VPN connection and sometimes the VPN disconnect). Here is an example, I open a connection to a database and work into the database for some time. Then, I left the browser with the pgadmin tab open for 1 hour or so and then when I try to operate with the database, I receive the message "Connection to the server has been lost.". Trying to refresh I receive the message "no password supplied" and the message is Ok. But I need to disconnect to the server and again connect.

![image](https://github.com/user-attachments/assets/0c63b66e-a060-4de9-9cc2-7e61266f4d7d)

![image](https://github.com/user-attachments/assets/40cad40a-f8ac-4d40-b938-ef0a846815e6)

![image](https://github.com/user-attachments/assets/9d706eda-2c61-424d-8c4e-40626b3486ea)

If there is a chance to have an option from left click button, where you can reconnect to the server and provide the password again, will be nice to do it in one click, or if there is no connection for that kind of case (no password supplied), the pgadmin shows the password popup in order to provide again the password.

Thanks!

ohelit:
Hi

These logs are from the container:
![image](https://github.com/user-attachments/assets/a76a6592-b90f-44ad-9f4d-dfa9eeb20af9)

`192- - [20/Nov/2024:03:54:21 -0500] "GET /sqleditor/status/1222271 HTTP/1.1" 500 95 "/sqleditor/panel/1222271?is_query_tool=true&sgid=3&sid=116&did=489075&database_name=ar" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36 Edg/131.0.0.0"

192 - - [20/Nov/2024:03:54:21 -0500] "GET /browser/database/children/3/116/489075 HTTP/1.1" 500 171 "/browser/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36 Edg/131.0.0.0"`

Here is the information about the server:
![image](https://github.com/user-attachments/assets/1f9af137-493c-44fb-aa31-e84ec6643c98)

Thanks

anilsahoo20:
[...]

PR #8778, opened 2025-05-20
Added refresh callback for server to ensure accurate UI state and icon updates on connection status changes. #8149

RohitBhati8269:
Detail Infor:- #8149

### DIFF
--- a/web/pgadmin/browser/server_groups/servers/static/js/server.js
+++ b/web/pgadmin/browser/server_groups/servers/static/js/server.js
@@ -331,6 +331,30 @@ define('pgadmin.node.server', [
 
           return false;
         },
+        /* Refresh the server */
+        refresh: function(args, i) {
+          let obj = this,
+            tree = pgBrowser.tree,
+            item = i || tree.selected(),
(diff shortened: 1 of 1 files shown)


web/pgadmin/browser/server_groups/servers/static/js/server.js
```diff
@@ -331,6 +331,30 @@ define('pgadmin.node.server', [
...
           return false;
         },
+        /* Refresh the server */
+        refresh: function(args, i) {
+          let obj = this,
+            tree = pgBrowser.tree,
+            item = i || tree.selected(),
+            data = item && tree.itemData(item);
+          // Fetch connection status.
+          fetch_connection_status(obj, data, tree, item);
```
adityatoshniwal, line 341: We had discussed not to call this function on refresh and use the node endpoint instead. The fix will be generic and will apply to all nodes.

adityatoshniwal (changes requested):

adityatoshniwal:
Closing in favor of https://github.com/pgadmin-org/pgadmin4/pull/8852

Issue #1926, opened 2018-07-02
Disable user password change (RM #3459)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/3459***
*Originally created by **Henrik Uggla** at 2018-07-02 07:22:44 UTC.*

Users tend to pick short and/or stupid login passwords. I would prefer if the Pgadmin Server administrators were the only ones with the power to change user login passwords.

Please make it possible to disable the Pgadmin password change dialog for ordinary users.

yogeshmahajan-1903:
Related to #1632

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-06-10/
Package: pip wheel
Environment: macOs Sonoma 14.7.1

PR #8805, opened 2025-05-29
Disable user password change for non-admin user. #1926

RohitBhati8269:
Details info:- #1926


pravesh-sharma:
Shouldn't this be configurable? What if someone wants the change password option available for non-admin users?

RohitBhati8269:
> Shouldn't this be configurable? What if someone wants the change password option available for non-admin users?

Yes, it could be, but we need to discuss it with the team.

pravesh-sharma:
> > Shouldn't this be configurable? What if someone wants the change password option available for non-admin users?
> 
> Yes, it could be, but we need to discuss it with the team.

I think in the issue also it is mentioned to make it possible to disable and not disable it permanently so I think we should make it configurable.

web/pgadmin/tools/user_management/PgAdminPermissions.py
```diff
@@ -37,6 +38,7 @@ class AllPermissionCategories:
     object_explorer = 'Object Explorer'
     tools = 'Tools'
     storage_manager = 'Storage Manager'
+    miscellaneous = 'Miscellaneous'
```
adityatoshniwal: This reminds me, missed the gettext wrapping. Please add to all.

web/migrations/versions/c62bcc14c3d6_.py
```diff
@@ -0,0 +1,43 @@
...
+    meta.reflect(op.get_bind(), only=('role',))
+    role_table = sa.Table('role', meta)
+
+    perm = role_table.c.permissions
+    op.execute(role_table.update().where(
+        (role_table.c.name != 'Administrator') &
+        (
+            (perm == None) |
+            (perm == '') |
+            (~perm.ilike('%change_password%'))
```
adityatoshniwal: what is this for? What is it doing?

RohitBhati8269: My bad, I missed removing that. I added it for my local testing.

web/pgadmin/tools/user_management/PgAdminPermissions.py
RohitBhati8269: Done

Issue #8691, opened 2025-04-24
Multi-Line Block Copy In Query Tool Only Copies First Line.


KeyboardG:
When performing multi-line edits in the query editor using Alt+Mouse selection, you can select text across multi-lines however when you perform a copy, it only copies from the first line of text. When you then multi-line paste, it repeats the first value in each line.

**To Reproduce**

1. Go to Query Tool
2. Select several rows of text with multi-line selection (Alt+mouse), for example dates across rows, Cntrl+C to copy, arrow the cursor over and press Cntrl-V to paste. 
3. See the error that the rows of text pasted are just the first line.

Alternatively, multi-select rows of text, Cntrl-C, switch to a different text editor and paste and see that only the first line of text was copied.

**Expected behavior**

If block selecting text and copying, the paste should also be block pasted.

**Desktop (please complete the following information):**
Version 9.2
Application Mode Desktop
Commit 73fefb860369ce660eb4b04d304d63d048935b83 2025-04-02
Python Version 3.13.2
Current User pgadmin4@pgadmin.org
Electron Version 35.1.2
Browser Chrome 134.0.6998.178
Operating System Windows-11-10.0.26100-SP0

KeyboardG:
Here is an example text that I tested with. I block selected with Alt+Mouse the date out part of the partition table names, then paste that else where and its only from the first line.

```
alter table public."SampleTableName" attach partition public."SampleTableName_20250309" for values from ('2025-03-09') to ('2025-03-10');
alter table public."SampleTableName" attach partition public."SampleTableName_20250310" for values from ('2025-03-10') to ('2025-03-11');
alter table public."SampleTableName" attach partition public."SampleTableName_20250311" for values from ('2025-03-11') to ('2025-03-12');
alter table public."SampleTableName" attach partition public."SampleTableName_20250312" for values from ('2025-03-12') to ('2025-03-13');
alter table public."SampleTableName" attach partition public."SampleTableName_20250313" for values from ('2025-03-13') to ('2025-03-14');
alter table public."SampleTableName" attach partition public."SampleTableName_20250314" for values from ('2025-03-14') to ('2025-03-15');
alter table public."SampleTableName" attach partition public."SampleTableName_20250315" for values from ('2025-03-15') to ('2025-03-16');
alter table public."SampleTableName" attach partition public."SampleTableName_20250316" for values from ('2025-03-16') to ('2025-03-17');
```

pravesh-sharma:
[...]

PR #8806, opened 2025-05-29
Fixed an issue in the query tool where, when multiple cursors were used to copy, only the first line was being copied. #8691

pravesh-sharma:


web/pgadmin/static/js/components/ReactCodeMirror/components/Editor.jsx
```diff
@@ -158,9 +158,8 @@ const defaultExtensions = [
   autoCompleteCompartment.of([]),
   EditorView.clipboardOutputFilter.of((text, state)=>{
     const lineSep = state.facet(eol);
-    // Fetch the primary selection from the editor's current state.
-    const selection = state.selection.main;
-    return state.doc.sliceString(selection.from, selection.to, lineSep);
+    // Fetch the selection range from the editor's current state.
+    return state.selection.ranges.map((range)=>state.sliceDoc(range.from, range.to)).join(lineSep) ?? '';
```
adityatoshniwal, line 162: did you check what sliceString does? We don't want to have any performance troubles.

adityatoshniwal (changes requested):

Issue #7466, opened 2024-05-10
%PATH%


Dmole:
**Describe the bug**

On Windows Backup and Restore exit with no error message.

**To Reproduce**

1. After using the installer linked from https://www.postgresql.org/download/windows/
2. Run pgAdmin 4
3. Create a database
4. Attempt to Backup from or Restore to the database
5. See it error with no message or logs

**Expected behavior**

Backup / Restore should just work.
If they fail, pgAdmin 4 should log why.

**Error message**

Process failed
No longs available
Exit code 3221225781

**Desktop (please complete the following information):**
 - OS: Windows 10 (server 2022)
 - pgAdmin 4 Version: 7.8

**Additional context**

If the shown command is pasted into cmd.exe the following error is shown:

The code execution cannot proceed because libzstd.dll was not found. 

echo  %errorlevel%
-1073741515

Adding the PostgreSQL bin folder to the windows %PATH% fixes the issue.

- EnterpriseDB should set the %PATH% : https://github.com/EnterpriseDB/edb-installers/issues/123
- PGAdmin4 should show error messages

khushboovashi:
Hi,

Please uninstall pgAdmin and re-install it from https://www.pgadmin.org/download/ and check.

dpage:
That seems like a workaround to the problem at hand. We should still fix the underlying issue, as other builds of pg_dump etc. may have similar dependencies.

Dmole:
@khushboovashi pgadmin.org links to enterprisedb.com for Windows, are you implying there is a fresh build with a fix?

khushboovashi:
> > Please uninstall pgAdmin and re-install it from https://www.pgadmin.org/download/ and check.
> 
> @khushboovashi (like postgresql.org) pgadmin.org links to enterprisedb.com for Windows, are you implying there is a fresh build with a fix?

pgadmin.org doesn't link with enterprisedb.com installers. But as a workaround, suggesting you install the pgAdmin installer, as we do not have this issue.

Dmole:
> pgadmin.org doesn't link with enterprisedb.com

curl -s "https://www.pgadmin.org/download/pgagent-windows/" | grep -c enterprisedb.com
1

But I do see now that's the wrong large windows button ( 2x the height and contrast as the <h2> )  

Regardless 

> we do not have this issue.

I'll test re-installing, breaking the path and testing if errors are shown.

dpage:
[...]

PR #8823, opened 2025-06-02
Fixed an issue where utilities such as pg_dump and pg_restore failed to log error messages when required dependency files were missing. #7466

akshay-joshi:


web/pgadmin/misc/bgprocess/processes.py
```diff
@@ -608,6 +630,10 @@ def status(self, out=0, err=0):
                 'process_state': self.process_state
             }
 
+        # Set error message based on exit code if stderr is empty.
+        if err_completed and len(stderr) == 0:
```
adityatoshniwal: There is no harm in putting the error message even if there is something in stderr. It may be possible that existing stderr doesn't have the error details. I should simply check `err_completed and self.ecode != 0`

web/pgadmin/misc/bgprocess/processes.py
```diff
@@ -608,6 +630,10 @@ def status(self, out=0, err=0):
...
+            set_error_msg(self.cmd, self.ecode, stderr)
```
adityatoshniwal: I would suggest to make it as get_error_msg and then append  the returned message to stderr here.

web/pgadmin/utils/constants.py
```diff
@@ -115,7 +115,13 @@
     ]
 }
 
-UTILITIES_ARRAY = ['pg_dump', 'pg_dumpall', 'pg_restore', 'psql']
+UTILITIES_ARRAY = ['pg_dumpall', 'pg_dump', 'pg_restore', 'psql']
+
+BG_PROCESS_ERROR_MSGS = {
+    3221225781: gettext('Unable to find a dll needed by the utility. Ensuring '
```
adityatoshniwal: Ensuring should be Ensure

adityatoshniwal (changes requested):

web/pgadmin/misc/bgprocess/processes.py
akshay-joshi: I did that first, but somehow it ended up stuck in an infinite loop.

web/pgadmin/utils/constants.py
akshay-joshi: It's correct as I take it from ChatGPT and Grammerly also suggested the same

web/pgadmin/misc/bgprocess/processes.py
akshay-joshi: Done

web/pgadmin/utils/constants.py
akshay-joshi: Done. Changed to Ensure.

PR #8824, opened 2025-06-03
Fix #7173: Add flag to allow access to system Python packages on recent Linux distributions (-> PEP-668)

### LINKED ISSUE #7173: Python virtual environment as installed from Debian package lacks `dbus-python`
After installing version 8.2 from packages on Debian Bookworm, pgAdmin 4 does not start, showing a "The pgAdmin 4 server could not be contacted" message and the last lines of the log are telling me that `python-dbus` was missing.

1. Install pgAdmin 4 from Debian package
2. Start it
3. See the error

**Expected behavior**

pgAdmin 4 should start OOTB

**Error message**

> The pgAdmin 4 server could not be contacted

 - OS: Debian Bookworm
 - Version: 8.2
 - Mode: Desktop
 - Package type: DEB

**Additional context**

Program starts fine after installing `dbus-python` into the installed Python virtual environment under `/usr/pgadmin4/venv`.

Please fix the Debian package to ship a complete Python virtual environment containing all needed dependencies.

dhs-rec:
This fixes #7173.

### DIFF
--- a/pkg/linux/build-functions.sh
+++ b/pkg/linux/build-functions.sh
@@ -63,7 +63,7 @@ _create_python_virtualenv() {
     cd "usr/${APP_NAME}" || exit
 
     # Create the blank venv
-    "${SYSTEM_PYTHON_PATH}" -m venv venv
+    "${SYSTEM_PYTHON_PATH}" -m venv --system-site-packages venv
     # shellcheck disable=SC1091
     . venv/bin/activate
 
--- a/tools/setup-python-env.sh
+++ b/tools/setup-python-env.sh
@@ -7,7 +7,7 @@ then
 fi
 
 echo Creating a blank environment...
-python3 -m venv venv || { echo 'ERROR: Failed to create the blank virtual env.' ; exit 1; }
+python3 -m venv --system-site-packages venv || { echo 'ERROR: Failed to create the blank virtual env.' ; exit 1; }
 
 echo Activating the virtual environment...
 . venv/bin/activate || { echo 'ERROR: Failed to activate the virtual env.' ; exit 1; }


khushboovashi:
Hi @dhs-rec,

The dbus-python is the indirect dependency of keyring (Ref:https://pypi.org/project/keyring/). The author of the keyring package mentioned that "_On Linux, the KWallet backend relies on [dbus-python](https://gitlab.freedesktop.org/dbus/dbus-python), which does not always install correctly when using pip (compilation is needed). For best results, install dbus-python as a system package._"

So, we are already installing python3-dbus as a pgAdmin dependency. Please check while installing pgAdmin whether this dependency is getting installed on your system or not.

dhs-rec:
Yes, `python3-dbus` IS installed as a system package. But that doesn't matter (as I already wrote in the ticket), because the `pgadmin4` Debian package ships a complete Python virtual environment, which isn't allowed to access system Python packages. Hence, `dbus-python` needs to be installed into the Python virtual environment that is contained in the `pgadmin4` Debian package. See `/usr/share/doc/python3.11/README.venv` and [PEP-668](https://peps.python.org/pep-0668/) for the details.

dhs-rec:
As an alternative solution, from above `README`:
> If needed, the isolated environment can also have access to system Python
modules, with the "--system-site-packages" flag.

This flag must be provided when creating the Python virtual environment.

khushboovashi:
@dhs-rec, do you have any specific installation? We test on Debian Bookworm and have never faced this issue. Also, as you can see, GitHub actions are failing with your PR.

dhs-rec:
No, plain Debian Bookworm, as stated in the ticket.

I'll look into the failing actions...

dhs-rec:
Hmm, can't see any tests on Debian Bookworm. I only see Ubuntu 22.04 (when it comes to Linux). But PEP-668 was only enabled as of Ubuntu 2**3**.04 (and Debian Bookworm). No wonder the problem didn't show up if you don't test on recent versions.

Anyway, I do see that in the Dockerfile, the Python virtual environment is indeed already created with `--system-site-packages`. I think I'll change the PR to include the flag into the Linux package creation scripts, too, and revert the change in `requirements.txt`, since installation of `dbus-python` requires a C compiler and some header files.

Issue #1947, opened 2018-07-09
Disable add new server, change or drop existing server connexion unless member of a specific pgadmin4 role (RM #3484)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/3484***
*Originally created by **Marc Rechté** at 2018-07-09 14:29:57 UTC.*

For server installation mode, it would be useful to disable the creation or modification of server connexions to selected users.

The pgadmin4 administrator would prepare the user sever connexion while creating his/her account.

This feature would be especially useful in a hosted shared pgadmin4 instance, which could be located in a privileged zone where one does not want the end-user to connect wherever he/she wants from that zone.

Thanks

yogeshmahajan-1903:
Relates to #1632

pravesh-sharma:
Working as expected. Tested on latest snapshot build.

PR #8831, opened 2025-06-04
Added role-based restriction for editing server connections. #1947

RohitBhati8269:
Detail Info:- #1947

### DIFF
--- a/web/pgadmin/browser/static/js/node.js
+++ b/web/pgadmin/browser/static/js/node.js
@@ -141,7 +141,7 @@ define('pgadmin.browser.node', [
             function() {
               return !!(self.canEdit(...arguments));
             } : (!!self.canEdit),
-          permission: self.type == 'server' ? 'object_register_server' : undefined,
+          permission: self.type == 'server' ? AllPermissionTypes.OBJECT_REGISTER_SERVER : undefined,
         }]);
       }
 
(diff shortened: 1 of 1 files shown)


web/pgadmin/tools/user_management/PgAdminPermissions.py
```diff
@@ -48,7 +48,7 @@ def __init__(self):
         self.add_permission(
             AllPermissionCategories.object_explorer,
             AllPermissionTypes.object_register_server,
-            gettext("Register/remove server")
+            gettext("Register/Edit/remove server")
```
adityatoshniwal: Probably rename it to `Manage server`

adityatoshniwal (changes requested):

web/pgadmin/browser/static/js/node.js
```diff
@@ -141,6 +141,7 @@ define('pgadmin.browser.node', [
             function() {
               return !!(self.canEdit(...arguments));
             } : (!!self.canEdit),
+          permission: self.type == 'server' ? 'object_register_server' : undefined,
```
adityatoshniwal: we have define enum, don't hardcode.

adityatoshniwal (changes requested):

Issue #6743, opened 2023-09-08
Open preferences in a tab instead of dialog


adityatoshniwal:
**Describe the solution you'd like**

Currently, the preferences dialog opens in dialog. A user has to frequently scroll around or resize the dialog to manage preferences. I suggest we can open the preferences in a tab just like VS-code does so that it will be easier to use.
In future if we add search then we'll have more space to implement.

anilsahoo20:
Tested and verified on latest snapshot build and source code.

Issue #2864, opened 2019-08-09
Add search box in preferenes (RM #4587)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/4587***
*Originally created by **Aditya Toshniwal** at 2019-08-09 10:39:53 UTC.*

Hi Team,

It would be very convenient to have a search box in preferences. Will save time.

Thanks.

anilsahoo20:
Tested and verified on latest snapshot build and source code.

PR #8832, opened 2025-06-05
1) Open the preferences in a tab instead of dialog for better UX. #6743

adityatoshniwal:
2) Add a search box to allow searching in the preferences tab. #2864


docs/en_US/images/preferences_misc_user_interface.png
akshay-joshi, line 1: Please update the images with "Save the application state" parameter.

web/pgadmin/static/js/PgTreeView/index.jsx
```diff
@@ -3,37 +3,64 @@ import { styled } from '@mui/material/styles';
...
 import { Tree } from 'react-arborist';
-import AutoSizer from 'react-virtualized-auto-sizer';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import ChevronRightIcon from '@mui/icons-material/ChevronRight';
 import PropTypes from 'prop-types';
 import IndeterminateCheckBoxIcon from '@mui/icons-material/IndeterminateCheckBox';
 import EmptyPanelMessage from '../components/EmptyPanelMessage';
 import CheckBoxIcon from '@mui/icons-material/CheckBox';
-
+import useResizeObserver from 'use-resize-observer';
```
akshay-joshi, line 21: Add a copyright header to this file.

docs/en_US/images/preferences_misc_user_interface.png
adityatoshniwal: This will be taken care in full preferences docs review.

adityatoshniwal:
> **GUI Related Comments**:
> 
> 1. When a workspace other than the default is selected, clicking the Settings button or selecting Preferences from the menu opens the Preferences tab in the default workspace. However, the active workspace does not switch accordingly.

Fixed.

> 2. Open the Preferences tab, make some changes, and click Reset All Preferences. Then select Save & Reload Later. While the changes are correctly applied in the backend, they are not reflected in the currently open tab. A manual refresh or closing and reopening of the Preferences tab is needed to see the updates.

Fixed. 

> 3. In the Preferences tab, go to Browser → Display, change any one toggle setting, and click Save. The setting is saved successfully. However, after this, attempting to toggle any controls again (especially multiple times) will not work; the controls become unresponsive.

Fixed.

> 4. Select the Browser node. You will see navigation links on the right side. Search for a string like "Layout", and click on any navigation link, it will fail, which is expected. Can we do something?

I have changed logic to select the first item automatically if filter changes.

> 
> **Source Code Related Comments**:
> 
> 1. Please update the missing preferences in the "preferences.rst" file, as you have already updated the screenshots. A few preferences are missing from the documentation (not related to your PR, but it's good to have)

As discussed, it will be taken care in a separate ticket.

> 2. Fix 5 old SonarQube code smells in Button.jsx if possible.

Fixed 3/5. Cognitive complexity didn't fix.

Issue #2659, opened 2019-06-10
Ability to customize Query Tool -> Edit keyboard shortcuts (RM #4339)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/4339***
*Originally created by **Alan Barrington-Hughes** at 2019-06-10 19:01:51 UTC.*

Add the ability to customize the keyboard shortcut values attributed to the Query Tool -> Edit operations.

I have a browser plugin that overrides the inline comment shortcut, I believe it would be beneficial to customize these values as one can for other Query Tools operations (Preferences -> Query Tool -> Keyboard shortcuts).

anilsahoo20:
@pravesh-sharma, found some issues
1. Toggle comment is working with old default shortcut cmd+/ along with new ones.
2. Auto-complete not working with default shortcut or other new shortcuts.

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-06-22/
Package: arm64
Environment: macOs Sonoma 14.7.1

PR #8838, opened 2025-06-09
Added support to customize query tool edit menu keyboard shortcuts. #2659

pravesh-sharma:


yogeshmahajan-1903 (changes requested):
1.Go to line shortcuts is missing on UI.
2.Cases for the all char keys should be either lower or upper. Is it case sensitive?

web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
```diff
@@ -541,18 +524,14 @@ export default function Query({onTextSelect, setQtStatePartial}) {
...
-                altKey: e.altKey,
-                shiftKey: e.shiftKey,
-                ctrlKey: e.ctrlKey,
-                metaKey: e.metaKey,
-              }));
-            }
+            queryToolCtx.mainContainerRef?.current?.dispatchEvent(new KeyboardEvent('keydown', {
+              which: e.which,
+              keyCode: e.keyCode,
+              altKey: e.altKey,
```
yogeshmahajan-1903, line 492: Remove the space.

pravesh-sharma: There is no space in the actual code, not sure why GitHub diff is showing space.

pravesh-sharma:
> 1.Go to line shortcuts is missing on UI. 2.Cases for the all char keys should be either lower or upper. Is it case sensitive?

Fixed

web/pgadmin/static/js/components/ReactCodeMirror/index.jsx
```diff
@@ -59,30 +63,45 @@ CopyButton.propTypes = {
...
     preventDefault: true,
     stopPropagation: true,
+  }, {
+    key: toCodeMirrorKey(preferences.comment), run: () => {
+      editor.current?.execCommand('toggleComment');
+    },
+    preventDefault: true,
+    stopPropagation: true,
+  },{
+    key: toCodeMirrorKey(preferences.format_sql), run: () => {
```
adityatoshniwal: this should not come here. Format SQL is not codemirror internal.
Or - you can add new command for formatSql just like toggleComment or others - because format here will work only is query tool and not outside.

web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
```diff
@@ -417,12 +350,64 @@ export default function Query({onTextSelect, setQtStatePartial}) {
       }
     };
 
+    const unregisterEditorExecCmd = eventBus.registerListener(QUERY_TOOL_EVENTS.EDITOR_EXEC_CMD, (cmd='')=>{
+      if(cmd == 'gotoLineCol') {
```
adityatoshniwal: Will gotoLineCol work outside query tool? If not we should add a cmd for this as well.

adityatoshniwal (changes requested):

web/pgadmin/static/js/components/ReactCodeMirror/index.jsx
pravesh-sharma: If we press the shortcut for format sql outside it will not cause any error or issues that's why I kept here. I will add a separate command

web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
pravesh-sharma: It works outside query tool. If editor is not editable like we have in SQL tab then it will not have any effect.

web/pgadmin/tools/sqleditor/static/js/components/sections/Query.jsx
```diff
@@ -318,57 +316,41 @@ export default function Query({onTextSelect, setQtStatePartial}) {
...
-          keyCode: shortcut.key.key_code,
-          metaKey: shortcut.ctrl_is_meta,
-          ctrlKey: shortcut.control,
-          shiftKey: shortcut.shift,
-          altKey: shortcut.alt,
+      let key = {}, gotolinecol = queryToolCtx.preferences.sqleditor.goto_line_col,
+        formatSql = queryToolCtx.preferences.sqleditor.format_sql;
+      switch(cmd) {
+      case 'gotoLineCol':
+        key = {
```
adityatoshniwal: the key creation logic is repeated. I think it can also be common.

yogeshmahajan-1903 (approved):

Issue #8675, opened 2025-04-19
Search Objects window opens a different object than the selected result


PhantomWatson:
# To reproduce
1. Search for a column name in a database
2. Go down the list of results, double-clicking or hitting Enter to open the object

# Expected result
- The object referenced in the Search Objects window is opened 
- A result always opens the same object

# Actual result
- Sometimes (once every ten or so attempts) what's opened is an object that's different from the search result, but in the same table as the search result
- Attempting to open the same result a second time, immediately after the first time, opens the correct object

yogeshmahajan-1903:
@PhantomWatson 
Can you please share screen recording for the same?

PhantomWatson:
![Image](https://github.com/user-attachments/assets/8d900e40-6354-45fb-b5fc-a5bc7e0782f9)

RohitBhati8269:
Tested and verified on [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-07-15/).
Package: dmg
Environment: macOS Sequoia 15.3.1

PR #8851, opened 2025-06-16
Fixed an issue in the Search Objects tool where selecting a node woul…

anilsahoo20:
…d sometimes result in a different node being selected. #8675

### DIFF
--- a/web/pgadmin/browser/static/js/browser.js
+++ b/web/pgadmin/browser/static/js/browser.js
@@ -197,6 +197,7 @@ define('pgadmin.browser', [
       obj.Events.on('pgadmin:browser:tree:update', obj.onUpdateTreeNode.bind(obj));
       obj.Events.on('pgadmin:browser:tree:refresh', obj.onRefreshTreeNodeReact.bind(obj));
       obj.Events.on('pgadmin-browser:tree:loadfail', obj.onLoadFailNode.bind(obj));
+      obj.Events.on('pgadmin:browser:tree:disable_select', obj.disableTreeSelection.bind(obj));
       obj.bind_beforeunload();
 
       /* User UI activity */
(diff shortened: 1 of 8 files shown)


web/pgadmin/browser/static/js/browser.js
```diff
@@ -1650,6 +1651,10 @@ define('pgadmin.browser', [
 
     },
 
+    disableTreeSelection: function(_disable) {
+      pgBrowser.disable_tree_select = _disable;
```
adityatoshniwal: I think disable selection value should be internal to tree object which can internally check when calling .select. That way we will not need to add check at all the places.
Other than that, I don't think this is a good solution. You're disable select on search objects - kind of a hack and not solution. You're not fixing the root cause.

adityatoshniwal (changes requested):

web/pgadmin/browser/static/js/browser.js
anilsahoo20: Done

web/pgadmin/static/js/helpers/Layout/index.jsx
```diff
@@ -192,6 +193,11 @@ export class LayoutDocker {
         })],
       }, null, 'float');
     }
+    if (panelData?.id == 'id-search-objects') {
+      // Prevent previous tree state restore and opened tree event being called as it is clashing with
+      // the tree.select() in tree.findNodeWithToggle() of Search objects module.
+      pgAdmin.Browser.tree.suppressAddOpenEvents = true;
```
adityatoshniwal: we discussed to set this only while searching - why are you doing it on open?

web/pgadmin/browser/static/js/browser.js
```diff
@@ -1653,6 +1653,32 @@ define('pgadmin.browser', [
       }
 
     },
+
+    editor_shortcut_keys: {
```
adityatoshniwal: this should be removed.

web/pgadmin/static/js/helpers/Layout/index.jsx
```diff
@@ -342,6 +348,11 @@ function DialogClose({panelData}) {
     <Box display="flex" alignItems="center">
       <PgIconButton title={gettext('Close')} icon={<CloseIcon />} size="xs" noBorder onClick={()=>{
         layoutDocker.close(panelData.activeId);
+        if (panelData?.activeId == 'id-search-objects') {
+          // Enable the previous tree state restore and opened tree event after closing of Search Objects dialog as it 
+          // is clashing with the tree.select() in tree.findNodeWithToggle() of Search objects module.
+          pgAdmin.Browser.tree.suppressAddOpenEvents = false;
```
adityatoshniwal: same as above.

adityatoshniwal (changes requested):

web/pgadmin/static/js/helpers/Layout/index.jsx
anilsahoo20: I actually checked, It's getting true properly but getting false and then all the other events like added and opened gets called. so it's causing the issue. so made it true while opening the search dialog and false when closing the dialog.

adityatoshniwal: There is a very good chance to break the existing tree flow with this change. Also, a user may manually expand a tree when the dialog is open - how are you planning to handle that?

anilsahoo20: I had changed the logic.

web/pgadmin/static/js/helpers/Layout/index.jsx
anilsahoo20: Done

web/pgadmin/browser/static/js/browser.js
anilsahoo20: Done

PR #8852, opened 2025-06-16
Fixed an issue where pgAdmin failed to update the server connection s…

adityatoshniwal:
…tatus when the server was disconnected in the background and a refresh was performed on that server. #8149


web/pgadmin/browser/static/js/browser.js
```diff
@@ -1239,173 +1238,39 @@ define('pgadmin.browser', [
...
+
+      const itemNodeData = nodeItem && this.tree.itemData(nodeItem);
+      let nodeObj = this.Nodes[itemNodeData?._type];
+
+      // If the node is a collection node, we can directly refresh it.
+      if(!nodeObj?.collection_node) {
+        // If the node is not a collection node, we need to fetch its data
+        // from the server and update the tree node.
+        try {
+          const url = nodeObj.generate_url(nodeItem, 'nodes', itemNodeData, true);
```
khushboovashi, line 1241: The backend call should be handled via tree_nodes.ts, and to achieve that, we may need to change the aspen tree refresh call itself.  As per our current logic, only children get refreshed on refresh.

adityatoshniwal: I think then there is inconsistency at a lot of places. Similar calls are done everywhere in browser.js for tree related stuff. I referred the existing code.

khushboovashi: If that is the case, we should eliminate that code, too, if possible.

adityatoshniwal: I cannot do that in this PR. Removing them will require complete tree revamp and I'm facing difficulty to understand that code.

khushboovashi: That change doesn't require in this PR, just this PR related change would be sufficient.

adityatoshniwal: The current changes are PR related only.

akshay-joshi: Aditya, please merge this PR We will do a revamp later.

PR #8897, opened 2025-06-26
adding the abillity to config to all users mutiple postgresql server

greasadh:
I was trying to configure default database for sandbox, and I see nothing to do it easly in config. 

I propose you a function that do the job 

The goal is to have a docker compose and in it you configure multiple database service, and in your pgadmin config add it. 

The next developper just make a docker compose up -d and have all the database without necessary needed to know them password and they are ready to use. 

The other goal is to not focus on config and adding db server and not be afraid to teardown the entire sandbox and rebuild it without lossing time.

### DIFF
--- a/web/config.py
+++ b/web/config.py
@@ -319,6 +319,33 @@ MAX_SESSION_IDLE_TIME = 60
 CONFIG_DATABASE_URI = ''
 
 ##########################################################################
+# add default server to all users 
+# exemple:
+# postgresql_servers = [
+#     {
+#         "name": "Serveur A",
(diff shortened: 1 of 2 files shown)


adityatoshniwal:
Hi @greasadh,
The changes you shared cannot be merged as it is against the overall design. What you can do is add a command to setup.py to add servers using command line. You'll also have more flexibility using cli.
Thanks.

adityatoshniwal:
I'm closing the PR as no response. Feel free to reopen.

Issue #8420, opened 2025-02-03
Installer not continuously signed


stephannn:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

At work we use Microsoft AppLocker and most apps in the user context are allowed by certificate. The pgadmin4-8.14-x64.exe admin installer is signed, but during the setup a tmp file is triggered which is no signed. Running the pgadmin4-8.14-x64.exe results in these logs:

```
pgadmin4-8.14-x64.exe was allowed to run.
%OSDRIVE%\USERS\MyUser\APPDATA\LOCAL\TEMP\2\IS-5LIUQ.TMP\PGADMIN4-8.14-X64.TMP was prevented from running.
```

Are more detailed log can be found in the details tab of the event:

```
+ System 
  - Provider 

   [ Name]  Microsoft-Windows-AppLocker 
   [ Guid]  {cbda4dbf-8d5d-4f69-9578-be14aa540d22} 
 
   EventID 8004 
 
   Version 0 
 
   Level 2 
 
   Task 0 
 
   Opcode 0 
 
   Keywords 0x8000000000000000 
 
  - TimeCreated 

   [ SystemTime]  2025-02-03T13:05:41.1525350Z 
 
   EventRecordID 3634 
 
   Correlation 
 
  - Execution 

   [ ProcessID]  7688 
   [ ThreadID]  8052 
 
   Channel Microsoft-Windows-AppLocker/EXE and DLL 
 
   Computer MyHost.Contoso.com
 
  - Security 

   [ UserID]  S-1-5-21-**** 
 

- UserData 

  - RuleAndFileData 

   PolicyNameLength 3 
 
   PolicyName EXE 
 
   RuleId {00000000-0000-0000-0000-000000000000} 
 
   RuleNameLength 1 
 
   RuleName - 
 
   RuleSddlLength 1 
 
   RuleSddl - 
 
   TargetUser S-1-5-21-****
 
   TargetProcessId 11972 
 
   FilePathLength 78 
 
   FilePath %OSDRIVE%\USERS\MyUser\APPDATA\LOCAL\TEMP\2\IS-5LIUQ.TMP\PGADMIN4-8.14-X64.TMP 
 
   FileHashLength 32 
 
   FileHash 6EFEC37D80B39E5C3A74CD9A3364AB17C7E49B7C3F84EF8FA17CB421884FC3AD 
 
   FqbnLength 1 
 
   Fqbn - 
 
[...]

PR #8933, opened 2025-07-08
Fixed an issue where windows installer is not signing tmp files gener…

adityatoshniwal:
…ated by InnoSetup. #8420

### DIFF
--- a/Make.bat
+++ b/Make.bat
@@ -375,7 +375,7 @@ REM Main build sequence Ends
 
 :SIGN_INSTALLER
     ECHO Attempting to sign the installer...
-    CALL "%PGADMIN_SIGNTOOL_DIR%\signtool.exe" sign /fd certHash /tr http://timestamp.digicert.com /td SHA256 "%DISTROOT%\%INSTALLERNAME%"
+    CALL "%PGADMIN_INNOTOOL_DIR%\ISCC.exe" "%WD%\pkg\win32\installer.iss"  "/Ssigntool=%PGADMIN_SIGNTOOL_DIR%\signtool.exe sign /fd certHash /tr http://timestamp.digicert.com /td SHA256 "%DISTROOT%\%INSTALLERNAME%" $f"
     IF %ERRORLEVEL% NEQ 0 (
         ECHO.
         ECHO ************************************************************
--- a/pkg/win32/installer.iss.in
+++ b/pkg/win32/installer.iss.in
@@ -36,6 +36,7 @@ ArchitecturesInstallIn64BitMode={#MyAppArchitecturesMode}
 AllowNoIcons=yes
 WizardImageFile=sidebar.bmp
 MinVersion={#MinimumWindowsVer}
+SignedUninstaller=yes
 
 [Languages]
 Name: "english"; MessagesFile: "compiler:Default.isl"


stephannn:
Sorry, I forgot to mention, when this signing change works, this part will be obsolete
```
:SIGN_INSTALLER
    ECHO Attempting to sign the installer...
    CALL "%PGADMIN_SIGNTOOL_DIR%\signtool.exe" sign /fd certHash /tr http://timestamp.digicert.com /td SHA256 "%DISTROOT%\%INSTALLERNAME%"
    IF %ERRORLEVEL% NEQ 0 (
        ECHO.
        ECHO ************************************************************
        ECHO * Failed to sign the installer
        ECHO ************************************************************
        PAUSE
    )

    EXIT /B 0
```

adityatoshniwal:
@stephannn We got the below logs on our build machine. Note that I didn't add `SignTool=signtool` in .iss file

```
Creating new signed uninstaller file: <[http://jenkins.buildfarm.pgadmin.org:8080/job/pgadmin4-windows-x64-qa/ws/pkg\win32\Output\uninst-6.2.2-c89243dec7.e32](http://jenkins.buildfarm.pgadmin.org:8080/job/pgadmin4-windows-x64-qa/ws/pkg%5Cwin32%5COutput%5Cuninst-6.2.2-c89243dec7.e32)>
Error in <[http://jenkins.buildfarm.pgadmin.org:8080/job/pgadmin4-windows-x64-qa/ws/pkg\win32\installer.iss](http://jenkins.buildfarm.pgadmin.org:8080/job/pgadmin4-windows-x64-qa/ws/pkg%5Cwin32%5Cinstaller.iss)>: Signed uninstaller mode is enabled. Using an external code-signing tool, please attach your digital signature to the following executable file:
```

stephannn:
Hi @adityatoshniwal ,
the issue probably is that only `SignedUninstaller=yes `was added, but this has a dependency to `SignTool=signtool`. SignTool signs the installer, but SignedUninstaller also the uninstall.exe in case the user would like to remove the program.

adityatoshniwal:
> Hi @adityatoshniwal , the issue probably is that only `SignedUninstaller=yes `was added, but this has a dependency to `SignTool=signtool`. SignTool signs the installer, but SignedUninstaller also the uninstall.exe in case the user would like to remove the program.

So you're suggesting adding `SignTool=signtool` should make it work?

stephannn:
yes, but signtool is just a variable name, but it has to be consistent with the command one. You can also name it SignPGAdmin.

Btw. here the output if I only add SignedUninstaller:
image

adityatoshniwal:
> yes, but signtool is just a variable name, but it has to be consistent with the command one. You can also name it SignPGAdmin.
> 
> Btw. here the output if I only add SignedUninstaller: image

But how can set the variable value? Sure We can set `SignTool=SignPGAdmin`, but SignPGAdmin should also be set right.

stephannn:
SignTool=**_signtool_** has just be a matching between the command line /S/**_signtool_**=%PGADMIN_SIGN

if you name both signtool names, which are in bold and Italic, for example SignPGAdmin it will also work. If one is different you will get:
image

I hope it is understandable.

PS.
I had to add another / after /S because I don't know the escape character

adityatoshniwal:
@stephannn  I've added SignTool=signtool as is and pushed the changes.

PR #8940, opened 2025-07-10
Enable pkce workflow for oauth2 authentication

albundy83:
A simple way to enable PKCE secure workflow.

I also imagine to totally overwrite `client_kwargs` parameters by passing a `OAUTH2_CLIENT_KWARGS` parameter.
Or even just pass a `OAUTH2_PKCE` that will enable a new `client_kwargs` with `code_challenge_method` and `response_type` correctly set.

### DIFF
--- a/web/pgadmin/browser/tests/test_oauth2_with_mocking.py
+++ b/web/pgadmin/browser/tests/test_oauth2_with_mocking.py
@@ -219,7 +219,8 @@ class Oauth2LoginMockTestCase(BaseTestGenerator):
         the default client_kwargs is correctly included.
         """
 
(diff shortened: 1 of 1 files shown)


khushboovashi (changes requested):
The code looks good to me. Please add test cases for the same. 
Reference: pgadmin/browser/tests/test_oauth2_with_mocking.py

Have you been able to test the code manually?

albundy83:
Yes, I have deployed it in Kubernetes using pgadmin release 9.5 and I have overridden `oauth2.py` using a configMap, it works perfectly :)
I will try to add tests, I will use link from [here](https://github.com/oauthlib/oauthlib/pull/743/files) as inspiration.

Issue #1923, opened 2018-06-29
Add tooltips/shortcut for keyboard shortcuts (RM #3456)


dpage:
***Issue migrated from Redmine: https://redmine.postgresql.org/issues/3456***
*Originally created by **Khushboo Vashi** at 2018-06-29 10:34:58 UTC.*

Put tooltip on buttons
Put shortcut on menu items

pravesh-sharma:
Working fine test on latest candidate build.

PR #8973, opened 2025-07-18
Show shortcuts in Electron native menu items.GH-1923

RohitBhati8269:


web/pgadmin/browser/static/js/MainMenuFactory.js
```diff
@@ -34,10 +49,12 @@ export default class MainMenuFactory {
           MainMenuFactory.electronCallbacks[smName] = sm.callback;
           return {
             ...sm.serialize(),
+            accelerator: MainMenuFactory.convertShortcutToAccelerator(sm.shortcut),
```
adityatoshniwal, line 52: accelerator will actually register a shortcut. It is not just for labelling. Did you test triggering the shortcut?
Although, it will trigger double shortcut.

adityatoshniwal (changes requested):

web/pgadmin/browser/static/js/MainMenuFactory.js
RohitBhati8269: I checked the electron side by console when hit the shortcut, it’s not being called twice.

adityatoshniwal: I don't know what you checked but it is opening 3 query tools for query tool shortcut.


https://github.com/user-attachments/assets/987c6882-f371-46db-8cb6-f03199f16dac

adityatoshniwal:
Closing in favor of https://github.com/pgadmin-org/pgadmin4/commit/bec47845bec9e2325a7fec88430fcd1d9573b4a1

Issue #8650, opened 2025-04-09
"Dashboard/State/Sessions" display should be resizable


mboeringa:
**Describe the solution you'd like**

I often make use of "Dashboard/State/Sessions/Details" display to check and possibly cancel sessions based on the SQL running by inspecting the details and SQL in the window.

I am not sure if this actually changed in a recent version, but I seem to remember that it was possible to resize the "Dashboard/State/Sessions" display to properly show large numbers of sessions and their details. Details are accessible through the expand button left of the session, however will often not fully display in the small fixed maximum vertical size of the "Sessions" window above "Locks". This forces the continuous need to scroll down to see the full details of the session, which is cumbersome. It would be desirable to be able to resize the "Sessions" section to a larger size, and/or have the grid showing the sessions automatically move up the selected session to the top of the "Sessions" window once a user clicks the "Expand" button to show the "Details", to allow the full details to be displayed in the limited fixed vertical space without scrolling.

In some situations, I have even seen it become confusing, because e.g. only the 'SQL' line of the "Details" was missing, suggesting there was no more information in the "Details" section, as there was a clean display cut exactly in-between the "Last state changed at" and "SQL" section of the "Details".

![Image](https://github.com/user-attachments/assets/19df45d7-8a72-48bd-aee6-ee32e897e35b)

**Describe alternatives you've considered**

See text above, make it resizable or automatically scroll expanded session details to top of fixed window.

**Additional context**

- Windows 10 - 22H2
- pgAdmin 4 v9.2

pravesh-sharma:
Issue fixed. Tested on latest snapshot build.

mboeringa:
@pravesh-sharma ,

Does this change possibly also change the display of the actual SQL statement from an inconvenient one-line textline to a larger multi-line textbox, preferably even with formatting? This would be really desirable. The current one-line display means that I need to copy out large SQL and paste them in another application or script to properly display them, as I cannot evaluate them properly otherwise.

PR #8990, opened 2025-07-25
Make Section container component vertically resizable.#8650

RohitBhati8269:
Detail info:- #8650


adityatoshniwal (changes requested):
image

Issue #9067, opened 2025-08-13
Disabling 'save application state' from preferences does not disable the saving application state feature.


yogeshmahajan-1903:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Disabling 'save application state' from preferences does not disable the saving application state feature.

**To Reproduce**

1. Go to preferences > User Interface > Turn off 'Save application state' radio button.
2. Open query tool, erd tool etc.
3. Quit/reopen pgadmin
4. QT and ERD tool tabs are restored.
Another scenario
6. Turn on 'Save application state' > Open different tools > Turn off 'Save application state' > Restart pgadmin > tools are restored.

**Expected behavior**

As 'Save application state' behaviour is turned off, tools should not be restored.

**Error message**

If possible, include the full, unedited text of any error messages displayed.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10, OSX monterey]
 - pgAdmin version: [e.g. 6.15]
 - Mode: [e.g Desktop or Server]
 - Browser (if running in server mode): [e.g. chrome, safari]
 - Package type: [e.g. RPM, DEB, Python, Container, etc.]

**Additional context**

pravesh-sharma:
Issue fixed. Tested on latest snapshot build.

PR #9068, opened 2025-08-13
Ensure that disabling "Save Application State" in Preferences prevents tool data from being saved and stops it from being restored on application restart.#9067

yogeshmahajan-1903:


web/pgadmin/static/js/helpers/Layout/index.jsx
```diff
@@ -470,10 +479,8 @@ export default function Layout({groups, noContextGroups, getLayoutInstance, layo
   const saveTab = (tab) => {
   // 'tab' here is the full TabData object, potentially with 'title', 'content', etc.
   // We only want to save the 'id' and any custom properties needed by loadTab.
-    const savedTab = {
-      id: tab.id,
-    };
-    if (tab.metaData && !BROWSER_PANELS.DEBUGGER_TOOL.includes(tab.id.split('_')[0])) {
+    const saveAppState = prefStore?.getPreferencesForModule('misc')?.save_app_state;
```
adityatoshniwal: At this moment, we can safely use `saveAppStateRef.current`

web/pgadmin/static/js/helpers/Layout/index.jsx
```diff
@@ -485,9 +492,13 @@ export default function Layout({groups, noContextGroups, getLayoutInstance, layo
         },
         restore: true,
       };
-      savedTab.metaData = updatedMetaData;
+      return {
```
adityatoshniwal: I didn't get why was this changed. Previous code looks fine to me.

adityatoshniwal (changes requested):

web/pgadmin/static/js/helpers/Layout/index.jsx
yogeshmahajan-1903: Done.

web/pgadmin/static/js/helpers/Layout/index.jsx
yogeshmahajan-1903: Done.

Issue #9095, opened 2025-08-22
Foreign key problem when update to 9.7


ish1mura:
Hi!

There is an error after upgrade from 9.6 to 9.7

Reproducing:

1) apt update
2) apt upgrade
3) Upgrade pgadmin4-server/web from 9.6 to 9.7
4) Error 500 in browser
5) Error from apache:
```
[Fri Aug 22 09:24:32.365870 2025] [wsgi:error] [pid 1743453:tid 134750647850688] [remote 10.33.7.131:48948]     return meth(
[Fri Aug 22 09:24:32.365872 2025] [wsgi:error] [pid 1743453:tid 134750647850688] [remote 10.33.7.131:48948]            ^^^^^
[Fri Aug 22 09:24:32.365877 2025] [wsgi:error] [pid 1743453:tid 134750647850688] [remote 10.33.7.131:48948]   File "/usr/pgadmin4/venv/lib/python3.12/site-packages/sqlalchemy/sql/elements.py", line 526, in _execute_on_connection
[Fri Aug 22 09:24:32.365880 2025] [wsgi:error] [pid 1743453:tid 134750647850688] [remote 10.33.7.131:48948]     return connection._execute_clauseelement(
[Fri Aug 22 09:24:32.365882 2025] [wsgi:error] [pid 1743453:tid 134750647850688] [remote 10.33.7.131:48948]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[Fri Aug 22 09:24:32.365887 2025] [wsgi:error] [pid 1743453:tid 134750647850688] [remote 10.33.7.131:48948]   File "/usr/pgadmin4/venv/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1641, in _execute_clauseelement
[Fri Aug 22 09:24:32.365889 2025] [wsgi:error] [pid 1743453:tid 134750647850688] [remote 10.33.7.131:48948]     ret = self._execute_context(
[Fri Aug 22 09:24:32.365891 2025] [wsgi:error] [pid 1743453:tid 134750647850688] [remote 10.33.7.131:48948]           ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #9112, opened 2025-08-27
Fixed an issue where pgAdmin config db migration was failing on application upgrade. #9095

pravesh-sharma:


web/migrations/versions/e6ed5dac37c2_.py
```diff
@@ -85,14 +86,12 @@ def upgrade():
                 pref_map[pref.id] = new_pref.id
 
     for key, val in pref_map.items():
-        record_to_update = session.query(UserPreference).filter_by(
-            pid=key).first()
-        if record_to_update:
-            record_to_update.pid = val
+        stmt = update(UserPreference).where(UserPreference.pid == key).\
```
adityatoshniwal: we've not used this update anywhere. We've used `sa.MetaData()`, keep it same.

adityatoshniwal (changes requested):

PR #9162, opened 2025-09-13
Fix Font Awesome icons not showing for oauth2 integration

charleszlu:
Specified Font Awesome icons would show a cross unless the class attribute```fab``` is changed to ```fa```.

Before:
image

After:
image


charleszlu:
Just realised fa alone does not cover brands.

PR #9166, opened 2025-09-15
1) Added 'failover' parameter support in CREATE and ALTER SUBSCRIPTIO…

anilsahoo20:
…N for PostgreSQL v17+.

2) Added 'two_phase' parameter support for ALTER SUBSCRIPTION for PostgreSQL v18+. 3) Updated versioned_template_loader.py to prioritize v18+ templates.

3) Updated versioned_template_loader.py to prioritize v18+ templates.

4) Updated the default value of the streaming parameter in CREATE SUBSCRIPTION to 'parallel' in PG v18 (previously false).


web/regression/re_sql/tests/test_resql.py
```diff
@@ -95,6 +95,35 @@ def setUp(self):
         # Added line break after scenario name
         print("")
 
+        # Create replication slot if it does not exist for the
```
yogeshmahajan-1903, line 98: This can be created using precondition_sql, Check tests for extensions.
Something like - 
```
DO $$ BEGIN IF NOT EXISTS (...) THEN PERFORM pg_create_logical_replication_slot(...); END IF; END $$;
```
And use      
"post_scenario_sql": "DROP ROLE IF EXISTS test_default_priv_user;"

akshay-joshi: @yogeshmahajan-1903 ,

Please create another PR for the suggested fix. I am merging this PR.

Issue #9157, opened 2025-09-10
Shortcuts do not work as expected in Windows operating systems with multiple keyboards.


ugurlu2001:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Problem:
------
Shortcuts do not work as expected in Windows operating systems with multiple keyboards.

How Long Has the Problem Been Presented?
-------------------------
This issue has been present since the latest PgAdmin update (Version: 9.8). This issue did not exist before.

Description:
--------
I use multiple keyboard configurations on the Windows operating system (Windows 11 Pro).

In my scenario, my keyboard configurations are as follows:

Q - Turkish Keyboard
F - Turkish Keyboard
U - Custom F-Keyboard
(Turkish Compatible; Custom configuration without Turkish characters, F-Keyboard Based)

When using the "Q" keyboard layout, all shortcuts work as expected.

For example: "Ctrl + A", "Ctrl + C", "Ctrl + F"

However; I'm having trouble switching between keyboard configurations using the "Win + Space" key combination:

When I switch to the "F - Keyboard" layout:

"Ctrl + A" > Doesn't work. It only selects the line where the cursor is located. ( [Q-Keyboard] A = U [F-Keyboard] )
"Ctrl + F" > Doesn't work. No events are triggered ( [Q-Keyboard] Q = F [F-Keyboard] )
"Ctrl + C" > working (OK)
"Ctrl + V" > working (OK)

I haven't checked all the other shortcuts.

As far as I've noticed, after the latest update, the keyboard shortcuts aren't working as intended!

I hope you can help me resolve this issue, which started in the latest version.

Thanks...

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - pgAdmin version: 9.8 (latest)
 - Mode: Desktop
 - Browser chrome - latest]
 - Package type: [e.g. RPM, DEB, Python, Container, etc.]

ugurlu2001:
"Ctrl + C" > works (OK)
"Ctrl + V" > works (OK)

Sory for typing mistake!

RohitBhati8269:
Working fine. Tested on the latest snapshot build.

ugurlu2001:
I upraded pgadmin version 9.9 today and keyboard shortcuts working very well when I switched keyboard layouts.

Thank you very much to all pgadmin team:)

PR #9171, opened 2025-09-16
Fixed an issue where shortcuts are not working as expected on multipl…

adityatoshniwal:
…e keyboard layouts. #9157

Update default shortcut for 'Replace' in editor to avoid using the Option key.


pravesh-sharma:
I have tested the PR and it is working as expected.

pravesh-sharma (approved):

PR #9255, opened 2025-10-15
[pgadmin4/pgadmin4-helm] Typo indentation fix for extraEnvVars field, serverDefinitions preserved tpl. README.md typo fix, increased resources

KorenP1:
There has been a typo in the indentation of extraEnvVars field
14 -> 12

Now preserving tpl integers and booleans in the serverDefinitions

Typo fix in the README.md, persistence.enabled supposed  to be true

Increased default resources

Please repush after merging.
Thanks. Koren.

PR #9283, opened 2025-10-22
Disable persistence by default

egvimo:
Many helm charts leave PVC disabled by default. In addition, the readme states that the default value for persistence is `false`. Therefore, it should also be `false` in the values.

### DIFF
--- a/pkg/helm/values.yaml
+++ b/pkg/helm/values.yaml
@@ -78,7 +78,7 @@ resources:
     memory: 512Mi
 
 persistence:
-  enabled: true
+  enabled: false
   size: 1Gi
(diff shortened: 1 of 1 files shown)


korenperetz:
Hi, I think that the wanted behavior is that it is enabled by default.
The README.md persistence.enabled is just a mistake.
You can see that alot of helm chart does enable persistence by default, for example all of bitnami charts and much more.

I have created a PR few weeks ago to fix a few things and I have added the typo fix. Thanks for informing.

https://github.com/pgadmin-org/pgadmin4/pull/9255

egvimo:
Closing in favor of #9255

Issue #8968, opened 2025-07-17
Option to show dtype in the Columns lists


deanm0000:
**Describe the solution you'd like**

I'd like it if we could see the data types next to the column names like this:

<img width="398" height="210" alt="Image" src="https://github.com/user-attachments/assets/46c836a0-4603-4061-8363-53965b51d0df" />

**Describe alternatives you've considered**

Generate the create table script to see the datatype or query the information_schema.

TorsionTools:
Yes, and also the Length when required would be great. (VarChar) (12)

<img width="325" height="194" alt="Image" src="https://github.com/user-attachments/assets/46ce6be8-aa00-40df-be84-3532d08eac30" />

RohitBhati8269:
Working fine, verified and tested on the latest [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2025-12-01/).

PR #9329, opened 2025-11-07
Added datatype display beside column name in browser tree. #8968

anilsahoo20:


web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/columns/sql/default/nodes.sql
```diff
@@ -1,4 +1,5 @@
 SELECT DISTINCT att.attname as name, att.attnum as OID, pg_catalog.format_type(ty.oid,NULL) AS datatype,
+pg_catalog.format_type(ty.oid,att.atttypmod) AS displaytypname,
```
adityatoshniwal, line 2: We should not duplicate logic in front end and back end. Lets keep is consistent with other places.

web/pgadmin/browser/register_browser_preferences.py
```diff
@@ -33,6 +33,15 @@ def register_browser_preferences(self):
         )
     )
 
+    self.show_column_datatype = self.preference.register(
+        'display', 'show_column_datatype',
+        gettext("Show column data type?"), 'boolean', False,
+        category_label=PREF_LABEL_DISPLAY,
+        help_str=gettext(
+            'If turned on, then the data type of columns will be displayed.'
```
adityatoshniwal: `If turned on, then the data type of columns will be displayed along side the column name.`

web/pgadmin/static/js/tree/tree_nodes.ts
```diff
@@ -60,6 +60,13 @@ export class ManageTreeNodes {
   public addNode = (_parent: string, _path: string, _data: []) => new Promise((res) => {
     _data.type = _data.inode ? FileType.Directory : FileType.File;
     _data._label = _data.label;
+
+    const nodeClass = pgAdmin.Browser.Nodes[_data._type];
+
+    if (nodeClass && typeof nodeClass.getNodeLabel === 'function') {
```
adityatoshniwal: I don't think we need to check `getNodeLabel`. Plus you've already provided the default value.

web/pgadmin/browser/static/js/node.js
```diff
@@ -99,6 +99,9 @@ define('pgadmin.browser.node', [
       // during the copying process of any node.
       return d;
     },
+    getNodeLabel: function(data) {
+      return data._label;
```
adityatoshniwal: it should be data.label

web/pgadmin/static/js/tree/tree_nodes.ts
```diff
@@ -60,6 +60,13 @@ export class ManageTreeNodes {
   public addNode = (_parent: string, _path: string, _data: []) => new Promise((res) => {
     _data.type = _data.inode ? FileType.Directory : FileType.File;
     _data._label = _data.label;
```
adityatoshniwal, line 62: directly call `pgAdmin.Browser.Nodes[_data._type].getNodeLabel`.

adityatoshniwal (changes requested):

web/pgadmin/static/js/tree/tree_nodes.ts
anilsahoo20: Done

web/pgadmin/browser/static/js/node.js
anilsahoo20: Done

web/pgadmin/static/js/tree/tree_nodes.ts
anilsahoo20: Done

web/pgadmin/browser/register_browser_preferences.py
anilsahoo20: Done and updated the image in docs.

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/columns/sql/default/nodes.sql
anilsahoo20: Done

PR #9348, opened 2025-11-11
User- Application name

PfunzoMulaudzy:
These changes will now use the current user's username instead of the application name for PostgreSQL application names and connection string prefixes, providing better identification of who is making the database connections.


adityatoshniwal:
Hi @PfunzoMulaudzy,

The APP_NAME is used to identify the app origin of connection, in this case pgAdmin.
This PR cannot be merged.

Thanks.

PfunzoMulaudzy:
What will be the best way to capture the machine name username, this will be useful for auditing purpose.

PR #9377, opened 2025-11-20
Fixed `.git` copy destination inside Dockerfile

Guiorgy:
In the Dockerfile in the `app-builder` stage the `.git` directory is copied:

```dockerfile
COPY .git .git
```

And then later in the same stage it is supposedly removed:

```dockerfile
# Build the JS vendor code in the app-builder, and then remove the vendor source.
RUN export CPPFLAGS="-DPNG_ARM_NEON_OPT=0" && \
    npm install -g corepack && \
    ... # skipped for brevity
    rm -rf node_modules \
           ... # skipped for brevity, attention to below VVV
           /pgadmin4/.git
```

I say _supposedly_, because the `.git` directory is copied into the filesystem root directory (`/`), whereas, the rest of the project files are inside the `pgadmin4` directory, and the later command expects the `.git` directory to be inside it. Clearly a mistake, it was supposed to be copied into `/pgadmin4/.git`.

Using `docker build --target app-builder --tag "pgadmin4:app-builder-stage" .` to only build and export this stage, and then `dive pgadmin4:app-builder-stage` to explore the layers, I confirmed that:

Screanshot of the output of dive

With that being said, is the `.git` directory even needed during the build process? If not, it would be better to completely omit the copy to save on having to copy more than `366 MB` (the size of `.git` as of writing) into the Docker image each time, which takes some time.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the deployment/build process so repository files are copied and cleaned consistently, resolving build-time file-handling issues and improving reliability across environments.


Guiorgy:
The only possible benefit to copying the `.git` directory in the build process, that I can see, is to get and store the git revision/hash, but even that has much better ways of handling, such as passing it from the host as a build argument. do we **_really_** need to be copying the whole `.git` directory into the image? It wasn't even being correctly copied until now anyway :D

PS. Gotta love flaky tests :P

Guiorgy:
@akshay-joshi @khushboovashi as I asked previously, is the .git directory really needed for the following npm and yarn commands? I didn't get the answer, and am not a web dev, so I wouldn't know.

Ideally, we'd avoid copying the .git directory all together, which only keeps on growing, slowing the image build.

Issue #9036, opened 2025-08-10
Discrepancy in how default values are passed to procedures in the debugger


diraneyya:
# Description

Described here https://github.com/EnterpriseDB/pldebugger/issues/33 and copied here as well (for convenience)

## Procedure

```sql 
create or replace procedure issue_pgadmin(
	in param text default 'silly')
language 'plpgsql' as $$
declare
	var text;
begin
	select format('Param''s literal value is "%L" and string value is "%s"', param, param) into var;
	raise notice '%', var;
end $$;
```

### Called from `psql`

Called, using `call issue_pgadmin();` gives the following:

```
NOTICE:  Param's literal value is "'silly'" and string value is "silly"
CALL
```

### Calling using the debugger

Leave the default value checkbox marked as seen below:

<img width="724" height="416" alt="Image" src="https://github.com/user-attachments/assets/bf59d534-18a8-46d5-9dd5-6da80cd7a5cb" />

In this case we get this output:

```
NOTICE:  Param's literal value is "'''silly''::text'" and string value is "'silly'::text"
CALL
```

Namely, the entire thing `'silly'::text` is passed as a string!

This is easily rectified by unchecking the default value checkbox and entering the default string value as just 'silly':

<img width="728" height="422" alt="Image" src="https://github.com/user-attachments/assets/6319087a-10d9-4859-a27a-95a5be9b7c89" />

In which case we get the expected output:

```
NOTICE:  Param's literal value is "'silly'" and string value is "silly"
CALL
```

RohitBhati8269:
Tested and verified. Working fine, no issue found.

PR #9397, opened 2025-11-25
Fix an issue on type casting of default string function/procedure arguments in debugger tool. #9036

anilsahoo20:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parameter validation in the debugger to prevent invalid argument configurations and provide clearer error messages when parameters are incorrectly positioned after default values.

### DIFF
--- a/web/pgadmin/tools/debugger/__init__.py
+++ b/web/pgadmin/tools/debugger/__init__.py
@@ -797,6 +797,7 @@ def filter_params_by_default(params_list, default_value_str):
     default_values = []
     if default_value_str:
         # Split at comma, strip spaces, and add a space after index 0
+        # as the arguments coming from frontend has space
         raw_defaults = str(default_value_str).split(',')
         default_values = [
             v.strip() if i == 0 else ' ' + v.strip()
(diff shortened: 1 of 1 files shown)


web/pgadmin/tools/debugger/__init__.py
```diff
@@ -785,6 +785,40 @@ def get_search_path(conn):
...
+    :param default_value_str:
+    :return:
+    """
+    default_values = []
+    if default_value_str:
+        # Split at comma, strip spaces, and add a space after index 0
+        # as the arguments coming from frontend has space
+        raw_defaults = str(default_value_str).split(',')
+        default_values = [
+            v.strip() if i == 0 else ' ' + v.strip()
```
khushboovashi, line 803: Can't we fix this issue at UI rather than stripping at the backend?

anilsahoo20: The query for getting the function info is already returning the list of default values with space and I did not find any way to remove by modifying the query. so handled that in the backend.

Issue #9368, opened 2025-11-18
Run Plain SQL restore with "\restrict" option.


akshay-joshi:

anilsahoo20:
Tested and verified on candidate build: https://developer.pgadmin.org/builds/2025-12-08-1/
Package: arm64, python
Environment: macOs Sonoma 15.7.1

akshay-joshi:
The pgAdmin project thanks the following individuals for reporting this issue:

1. Xi Rui <Octane-vs@outlook.com>
2. Ngọc Chung Kim <kimngocchung.k2a@gmail.com>
3. Dominik Penner <dominik.penner@mandconsulting.ca>
4. Louis Nyffenegger <louis@pentesterlab.com>
5. Meenakshi SL <msreekumar@endor.ai>

We have already assigned a CVE for this vulnerability, which can be found here:

https://nvd.nist.gov/vuln/detail/CVE-2025-13780

PR #9426, opened 2025-12-02
Plain SQL restore runs with '\restrict' option to prevent harmful psql meta-commands. #9368

akshay-joshi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Plain SQL restore now runs with an added backend restriction to block embedded meta-commands, improving security when restoring untrusted plain-text dumps.

* **Documentation**
  * Restore dialog documentation updated to explain the security behavior and how plain SQL restores are handled.

### DIFF
--- a/docs/en_US/restore_dialog.rst
+++ b/docs/en_US/restore_dialog.rst
@@ -29,6 +29,10 @@ restore process:
      copy of the backed-up object.
    * Select *Plain* to restore a plain SQL backup. When selecting this option
      all the other options will not be applicable.
+     **Note:** The plain SQL restore process is executed in the backend using
+     the psql command with the \restrict option. The purpose of \restrict is to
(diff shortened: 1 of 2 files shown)


docs/en_US/restore_dialog.rst
```diff
@@ -29,6 +29,11 @@ restore process:
      copy of the backed-up object.
    * Select *Plain* to restore a plain SQL backup. When selecting this option
      all the other options will not be applicable.
+     **Note:** The plain SQL restore process is executed in the backend using
```
dpage, line 32: **Note:** The plain SQL restore process is executed in the backend using
     the psql command with the \restrict option. The purpose of \restrict is to
     enhance security by preventing dangerous commands embedded in a
     plain text dump file from being executed on a PostgreSQL server.

PR #9485, opened 2025-12-25
Fix: User Management email column closure bug

mktwix:
The email column in User Management was using accessorKey which gets converted to accessorFn in PgTable.jsx. This caused a closure bug where all columns captured the same accessorKey value, resulting in all users displaying the logged-in user's email address.

Fix by explicitly using accessorFn to avoid the closure issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing email addresses in the Users list to display correctly when email data is unavailable.

### DIFF
--- a/web/pgadmin/tools/user_management/static/js/Users.jsx
+++ b/web/pgadmin/tools/user_management/static/js/Users.jsx
@@ -208,7 +208,7 @@ export default function Users({roles}) {
     },
     {
       header: gettext('Email'),
-      accessorKey: 'email',
+      accessorFn: (row) => row.email ?? '',
       enableSorting: true,
       enableResizing: true,
       size: 200,


khushboovashi:
Hi @mktwix, can you please provide a screenshot of the mentioned bug in this PR?

adityatoshniwal:
@mktwix In addition to what Khushboo asked, if the issue is closure in PgTable then I think PgTable should be fixed.

mktwix:
@khushboovashi @adityatoshniwal 
The issue is that when a user logs into the web app, the user menu displays the main admin user's email instead of the logged-in user’s email.
Here I mean
Screenshot 2026-01-30 115942

khushboovashi:
@mktwix, I haven't come across this scenario. The user menu displays only the logged-in user's email.

adityatoshniwal:
@mktwix Your fix is unrelated to the issue you're reporting.

PR #9497, opened 2025-12-30
More Dockerfile optimizations

Guiorgy:
3 things changed by this PR (separated into 3 commits):

1. Instead of copying the `.git` directory and then deleting it, we should use [RUN --mount=type=bind](https://docs.docker.com/reference/dockerfile/#run---mounttypebind) to temporarily mount the directory from the host in read-only mode visible to the commands in that `RUN` instruction. This avoids the needless copy and removal of `.git`, which will only grow in size as the repo ages (currently around 352 MB).
2. Instead of deleting the build cache directories after building, use [RUN --mount=type=tmpfs](https://docs.docker.com/reference/dockerfile/#run---mounttypetmpfs) to mount the build directories as tmpfs, which will be faster (since RAM is used) and automatically released at the end of the `RUN` instruction.
3. Split a `RUN` instruction to avoid redownlaoding and installing `apk` packages when the `requirements.txt` files is modified. Now only `pip` packages are redownloaded (assuming the `apk` stage is still cached) improving build time during development.

## Release Notes

* **Chores**
  * Optimized Docker build process and dependency installation workflow.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,7 +64,6 @@ RUN --mount=type=bind,source=.git,target=/pgadmin4/.git \
 FROM python:3-alpine AS env-builder
 
 # Install dependencies
-COPY requirements.txt /
 RUN apk add --no-cache \
         make && \
     apk add --no-cache --virtual build-deps \
@@ -77,8 +76,9 @@ RUN apk add --no-cache \
         cargo \
         zlib-dev \
         libjpeg-turbo-dev \
-        libpng-dev && \
-    python3 -m venv --system-site-packages --without-pip /venv && \
+        libpng-dev
+COPY requirements.txt /
+RUN python3 -m venv --system-site-packages --without-pip /venv && \
     /venv/bin/python3 -m pip install --no-cache-dir -r requirements.txt && \
     apk del --no-cache build-deps
 


Dockerfile
```diff
@@ -40,23 +39,23 @@ COPY web /pgadmin4/web
 WORKDIR /pgadmin4/web
 
 # Build the JS vendor code in the app-builder, and then remove the vendor source.
-RUN export CPPFLAGS="-DPNG_ARM_NEON_OPT=0" && \
+RUN --mount=type=bind,source=.git,target=/pgadmin4/.git \
```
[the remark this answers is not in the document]
Guiorgy: Please read the docs that I linked to:

> A bind mount is read-only by default.
> Option:    rw,readwrite	Allow writes on the mount. Written data will be discarded.

It **is** read-only.

Edit: also, there's no `readonly` option to begin with.

Dockerfile
```diff
@@ -40,23 +39,23 @@ COPY web /pgadmin4/web
...
+    rm -rf yarn.lock \
            package.json \
            .[^.]* \
            babel.cfg \
            webpack.* \
            jest.config.js \
-           babel.* \
-           ./pgadmin/static/js/generated/.cache \
-           /pgadmin4/.git
+           babel.*
```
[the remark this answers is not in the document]
Guiorgy: I can not find any mention of BuildKit being a requirement for the `--mount` option in the Dockerfile Reference. Can you point out the exact location in any documentation where this is specified?

And even if it were required, BuildKit has been the default since Docker Engine 23, which was released 2023-02-01, almost 3 years ago now, not to mention that with the release of 29, all versions prior to 25 are now EoL:

> Docker versions older than v25 are now end of life, and as such, we have increased the Minimum API version to [1.44 (Moby v25)](https://github.com/moby/moby/pull/51186).

There should not be anybody using the legacy EoL build system!

As for the requirement to have the `git` directory, this was the case even before the PR:

```
 => ERROR [app-builder 3/6] COPY .git /pgadmin4/.git                                               0.0s
------
 > [app-builder 3/6] COPY .git /pgadmin4/.git:
------
Dockerfile:34
--------------------
  32 |         zlib-dev
  33 |
  34 | >>> COPY .git /pgadmin4/.git
  35 |     # Create the /pgadmin4 directory and copy the source into it. Explicitly
  36 |     # remove the node_modules directory as we'll recreate a clean version, as well
--------------------
ERROR: failed to build: failed to solve: failed to compute cache key: failed to calculate checksum of ref xtsbduvkl3uzrzzvr7gwme7v1::n4ygw29hpyfw4nn37ldc01h2f: "/.git": not found
```

While I agree that it would be better to document this requirement, this PR is only meant to make some optimizations, so I'll leave that decision to the maintainers.

Guiorgy:
@coderabbitai Just a note, in the list of things you learnt you have:

```
Learnt from: Guiorgy
Repo: pgadmin-org/pgadmin4 PR: 9497
File: Dockerfile:42-42
Timestamp: 2025-12-30T23:46:57.826Z
Learning: In Dockerfiles using BuildKit, `RUN --mount=type=bind` mounts are read-only by default. The `readonly` or `ro` flag is not needed. This differs from runtime bind mounts (`docker run --mount` or `-v`), which are read-write by default and require the `ro` or `readonly` flag to be read-only.
```

The `RUN --mount=type=bind` instruction/option doesn't even have a flag named `readonly` or `ro`. It does have `rw` and `readwrite` flags, since it's read-only by default already. Quote from the docs:

> rw,readwrite	Allow writes on the mount. Written data will be discarded.

So the part "The `readonly` or `ro` flag is not needed" is wrong, since they don't exist, at least, there's no mention of such a thing in the reference.

Dockerfile
```diff
@@ -31,7 +31,6 @@ RUN apk add --no-cache \
     yarn \
     zlib-dev
 
-COPY .git /pgadmin4/.git
```
adityatoshniwal, line 34: the only problem with this is - we get the latest commit id and put in a file when bundling. The commit info is shown in the about dialog.
We could restrict the size of .git while creating the builds using `git clone --depth 10 <repourl>`

Guiorgy: As the description of the PR, and the commit message explain, the RUN command has:

```Dockerfile
RUN --mount=type=bind,source=.git,target=/pgadmin4/.git
```

Which mounts the `.git` directory in a read-only mode temporarily for that RUN instruction. I confirmed that the `commit_hash` file was correctly created and  the copy-deletion was avoided.

I can build an image and copy the output and the contents of the `commit_hash` file here if it would help?

adityatoshniwal: Great!

Guiorgy:
@adityatoshniwal Just an FYI, CodeRabbitAI pointed out correctly that currently building a docker image has a hard requirement for the existence of the `.git` directory, without it the build fails. This was true before this change and also now. If you ditribute tarballs without the `.git` directory anywhere, just know that the users of those can't build the docker image. Usually, metadata like git commit hash, build date and etc. are passed as build arguments, but that would require many changes, mounting the `.git` directory was the least disruptive and efficient solution I could find. CodeRabbitAI suggested better documenting the hard requirement for the `.git` directory in the docker build README, which is something worth considering.

Issue #9500, opened 2026-01-02
SSL mode uses localized value "允許" instead of "allow" in Traditional Chinese UI, causing connection failure


CatTimothy:
**Describe the bug**

When using pgAdmin in Traditional Chinese (zh_TW), selecting SSL Mode as “允許” causes pgAdmin to send the literal localized string to PostgreSQL instead of the expected value allow. PostgreSQL does not recognize this value and rejects the connection.

**To Reproduce**

1. Set pgAdmin UI language to Traditional Chinese  
2. Create or edit a server connection  
3. Set **SSL Mode** to “允許”  
4. Attempt to connect 

**Expected behavior**

Localized UI labels should map to valid PostgreSQL SSL mode values (e.g. `allow`) internally.

**Error message**

connection is bad: invalid sslmode value: "允許"

**Screenshots**

<img width="1196" height="536" alt="Image" src="https://github.com/user-attachments/assets/39991840-cb37-475d-9d08-c3c6ba4616dc" />

**Desktop**
 - OS: Linux
 - pgAdmin 4: 9.11
 - Mode: Desktop
 - Browser: None
 - Package type: DEB

RohitBhati8269:
Tested and verified on snapshot [build](https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2026-01-20/)
Package: dmg
Environment: macOS Tahoe 26.2

PR #9517, opened 2026-01-07
Fixed an issue where connection parameters were using localized values instead of literal values, causing connection failures. #9500

pravesh-sharma:
## Summary by CodeRabbit

* **Refactor**
  * Updated server connection parameter configuration handling to streamline internal implementation while maintaining all existing functionality and user interface behavior.


web/pgadmin/browser/server_groups/servers/static/js/variable.ui.js
```diff
@@ -104,7 +104,7 @@ export default class VariableSchema extends BaseUISchema {
       return {
         cell: 'select',
         options: (variable.enumvals || []).map((val)=>(typeof(val)=='string' ? {
-          label: val,
+          label: gettext(val),
```
adityatoshniwal: gettext doesn't work on variables. remove this change.

PR #9577, opened 2026-01-30
Add files via upload

sbraaa:
update translations


akshay-joshi:
@sbraaa,

Translation builds are failing and also please correct the commit message to "Updated Italian translation for v9.12"

sbraaa:
> @sbraaa,
> 
> Translation builds are failing and also please correct the commit message to "Updated Italian translation for v9.12"

Hi @akshay-joshi ,
for the commit message you're right! I'll do.
But I cannot understand which error causes broke the build :  my files are fully validated. Looking at the logs it seems the problem is related to CN translation.
Could you check please?

Issue #8988, opened 2025-07-25
Restore the internal settings of individual tab tools like QT, Schema diff, ERD


anilsahoo20:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

When we use different tools like QT, ERD, Schema diff, we are not restoring internal settings of QT like Open in a new tab, different execute options, different explain settings and for ERD tool like SQL options, Cardinality notations and Schema diff like Compare options, filter options of each separate tab.

anilsahoo20:
Tools are not getting closed on Save application state=false.

Steps:

1. Save application state=true
2. open some tools like QT,ERD,Schema diff, PSQL
3. Change Save application state to false.
4. Quit the app
5. Reopen the app
6. Still shows the tools tabs

pravesh-sharma:
I have implemented fix for ERD tool and schema diff. Created a new issue for the query tool. #9603

anilsahoo20:
@pravesh-sharma 
Schema diff internal settings for filter options when no option selected, is not getting restored properly.

anilsahoo20:
Tested and verified on snapshot build: https://www.postgresql.org/ftp/pgadmin/pgadmin4/snapshots/2026-02-27/
Package: arm64
Environment: macOs Sonoma 15.7.4

PR #9600, opened 2026-02-06
Fixed an issue where internal settings for the separate tools tab were not being restored during application relaunch. #8988

pravesh-sharma:
## Summary by CodeRabbit

* **New Features**
  * ERD toolbar preferences (notation, cardinality display, SQL formatting) are now saved, restored, and applied automatically; toolbar reflects saved choices.

* **Improvements**
  * Schema Diff compare and filter options are persisted and restored; filter selections and compare params now sync across sessions for a more consistent workflow.

### DIFF
--- a/web/pgadmin/tools/erd/static/js/erd_tool/components/MainToolBar.jsx
+++ b/web/pgadmin/tools/erd/static/js/erd_tool/components/MainToolBar.jsx
@@ -146,7 +146,7 @@ export function MainToolBar({preferences, eventBus, fillColor, textColor, notati
       }
       notationRef.current = notation;
     }
-  }, [preferences, toolbarPrefs]);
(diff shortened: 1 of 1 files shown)


web/pgadmin/tools/erd/static/js/erd_tool/components/MainToolBar.jsx
```diff
@@ -72,6 +72,7 @@ export function MainToolBar({preferences, eventBus, fillColor, textColor, notati
   const notationMenuRef = React.useRef(null);
   const isDirtyRef = React.useRef(null);
   const [checkedMenuItems, setCheckedMenuItems] = React.useState({});
+  const notationRef = React.useRef(notation);
```
adityatoshniwal: why do we need a notationRef when you directly use `notation`

PR #9633, opened 2026-02-13
docker arbitrary user support via `PUID` and `PGID`

kon-foo:
This PR adds support for `PUID` and `PGID` environment variables as popularised by [linuxserver.io ](https://docs.linuxserver.io/misc/non-root/)images. This allows users to control the UID and GID of the process running gunicorn inside the container.

## Why

Currently, when running pgAdmin via Docker with bind-mounted volumes, the mounted directories must be owned by the hardcoded 5050 user on the host and inside the container. This creates two issues:

1. Permission Headaches:
   Users are forced to manually chown directories on the host to 5050 before starting the container. This is cumbersome, especially if running docker as a non-root users and potentially lose access to the mounted dir.

2. Security Risks:
   The hardcoded 5050 UID might already be in use on the host system. Running the container as this UID could unintentionally grant pgAdmin access to files or resources owned by that UID or vise versa.

## Solution

1. Remove the hardcoded 5050 user, so the container starts then enrypoint.sh script as `UID` 0 by default.
2. In the script, default to `PUID` 5050 and `GUID` 0 to keep current behaviour when no env vars are set.
3. If the script runs as `"$(id -u)" = "0" `, aka no `--user` flag was used to run the image:
   - ensure a group with `PGID` exists and reassign the existing pgadmin user to `PUID` and `PGID`.
   - `chown` required files and dirs to said user.
   - use [su-exec](https://github.com/ncopa/su-exec) to drop privileges and run gunicorn as `$PUID:$PGID`

The check for id=0 makes sure nothing changes for users, that use the `--user` flag or `user: ` in compose to run the image.

## Compatability

Without PUID/PGID set, the behaviour is identical to the current image.

### Tests

For each test, I cheked:

```shell
# The id of the pgadmin user
docker exec pgadmin-test id pgadmin
# The the gunicorn process runs as that user
docker exec pgadmin-test ps aux | grep gunicorn | grep -v grep
# The permissions of a mounted volume
ls -ln /tmp/pgadmin-test-data/
# Access to the web UI
curl -s -o /dev/null -w "%{http_code}" http://localhost:8080/login
```

| #   | Scenario               | Key results                                 |
| --- | ---------------------- | --------------------------------------------- |
| 1   | Default (no PUID/PGID) | ✅ Still starts as UID 5050, GID 0            |
| 2   | PUID=1000 PGID=1000    | ✅ UID 1000, GID 1000, vmodeolume owned correctly |
| 3   | PUID only              | ✅ UID custom, GID defaults to 0              |

Further tests conducted:

| # | Scenario | Key results |
|--|--|--|
| 4 | TLS | ✅ TLS still works. Won't work if /certs is mounted as a read-only filesystem, but that is expected. |
| 5 | Server JSON import | ✅ Imported as root during initialization, no change of permissions needed. |

## Open Questions:

- Are there additional paths that need to be `chown`ed? Currently `/run/pgadmin`, `/var/lib/pgadmin`, `/certs` and `pgadmin4/config_distro.py` are covered.
- Are there any integration tests to be run?

I have been using this for a couple of weeks now, but I am only using a very small subset of PgAdmin functionality.

## Summary by CodeRabbit

* **New Features**
  * PUID/PGID environment support and root-aware startup enabling the server to run as a target user.
  * Safer, conditional ownership adjustments at startup and improved OpenShift compatibility for correct file permissions.

* **Chores**
  * Runtime image includes additional system utilities to expand runtime capabilities.
  * Default container runtime user changed to root.

### DIFF
--- a/pkg/docker/entrypoint.sh
+++ b/pkg/docker/entrypoint.sh
@@ -9,7 +9,8 @@ if [ "$(id -u)" = "0" ]; then
     fi
 
     # Reassign the pgadmin user to the desired UID/GID
-    usermod -o -u "$PUID" -g "$PGID" pgadmin 2>/dev/null || true
+    usermod -o -u "$PUID" -g "$PGID" pgadmin 2>&1 || \
+        echo "WARNING: usermod failed for UID=$PUID GID=$PGID"
 
     # Compose su-exec command
     SU_EXEC="su-exec $PUID:$PGID"


khushboovashi:
@kon-foo, please fix the issues reported by CodeRabbit.

kon-foo:
@khushboovashi I added a `safe_chown` function to check if a dir/file exists and if it already has the correct ownership before chowning it. The ownership check of course has the limitations that it only checks the top-level paths, but checking recursively would kinda negate the performance benefit.
We're now also chowning `$CONFIG_DISTRO_FILE_PATH` to account for `$PGADMIN_CUSTOM_CONFIG_DISTRO_FILE`.

kon-foo:
Regarding the "Nitpick" comment by @coderabbitai 
**Consider validating that PUID and PGID are numeric.** 
Non-numeric PUIDs or PGIDs lead to very clear error messages already, before failing to start:
```
addgroup: invalid number 'test'
pgAdmin will run as UID=1000, GID=test
...
chown: unknown user/group 1000:test
su-exec: getgrnam(asdasd): Invalid argument
```
**Consider logging when usermod fails to aid debugging**
The `-o` flag makes it even work with duplicate IDs. However I agree that supressing an error silently is not good so now we issue an additional warning when usermod fails. 
```
WARNING: usermod failed for UID=test GID=dummy
```

PR #9726, opened 2026-03-10
Add SQL formatting and query plan analysis module via explain-postgresql.com API

MGorkov:
This PR introduces a new module that integrates with the Explain PostgreSQL API (https://explain-postgresql.com). It provides users with enhanced SQL query formatting and in-depth query plan analysis directly within pgAdmin4.

Key Features:

- External Plan Analysis: Send query plans to the external API for a more detailed analysis.
- SQL Formatting: Utilize the API's capabilities to automatically format and beautify SQL queries.

## Summary by CodeRabbit

* **New Features**
  * Explain PostgreSQL integration: new Explain panel for analyzing/displaying query plans from query results, option to send plans to Explain PostgreSQL API, and control over plan privacy.
  * Remote SQL formatting: option to format SQL via Explain PostgreSQL API with graceful fallback to local formatter and clear success/error feedback.

* **Localization**
  * Updated translations across many languages to surface Explain PostgreSQL labels, messages, help text, and errors.

### DIFF
--- a/docs/en_US/release_notes_9_14.rst
+++ b/docs/en_US/release_notes_9_14.rst
@@ -14,7 +14,7 @@ Supported Database Servers
 
 Bundled PostgreSQL Utilities
 ****************************
-**psql**, **pg_dump**, **pg_dumpall**, **pg_restore**: 18.0
+**psql**, **pg_dump**, **pg_dumpall**, **pg_restore**: 18.2
 
 
 New features
@@ -49,3 +49,4 @@ Bug fixes
   | `Issue #9740 <https://github.com/pgadmin-org/pgadmin4/issues/9740>`_ -  Fixed an issue where the AI Assistant input textbox sometimes swallows the first character of input.
   | `Issue #9758 <https://github.com/pgadmin-org/pgadmin4/issues/9758>`_ -  Clarify where the LLM API key files should be.
   | `Issue #9789 <https://github.com/pgadmin-org/pgadmin4/issues/9789>`_ -  Fixed an issue where the Query tool kept prompting for a password when using a shared server.
+  | `Issue #9795 <https://github.com/pgadmin-org/pgadmin4/issues/9795>`_ -  Support /v1/responses for OpenAI models.
--- a/pkg/helm/Chart.yaml
+++ b/pkg/helm/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: pgadmin4-helm
-version: 9.13.0
-appVersion: 9.13.0
+version: 9.14.0
+appVersion: 9.14.0
 description: pgAdmin is the most popular and feature rich Open Source administration and development platform for PostgreSQL, the most advanced Open Source database in the world.
 home: https://pgadmin.org
 sources: ["https://github.com/pgadmin-org/pgadmin4/tree/master/helm"]
--- a/web/version.py
+++ b/web/version.py
@@ -18,7 +18,7 @@
 
 # Application version number components
 APP_RELEASE = 9
-APP_REVISION = 13
+APP_REVISION = 14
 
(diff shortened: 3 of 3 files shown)


dpage:
[With my pgAdmin hat on, and NOT my PGCA one] Can you confirm that explain-postgresql.com has appropriate licensing for use of the domain name including a community trademark, or considers it fair use (and if so, under what part of the trademark policy)? As licensees ourselves, we have to be careful about using any non-compliant services.

We also need to consider whether we actually want to add support for 3rd party services such as this - we've only used services from large PostgreSQL vendors in the past, and this one is quite different in that it is effectively sending schema/query data to a 3rd party, unlike the others which call APIs upon specific user request to create a service (e.g. an RDS instance)

On the patch itself, I note:

- The "Explain PostgreSQL" tab opens on every EXPLAIN query regardless of the preference setting — no check for explain_postgresql preference before opening the tab and making the external API call                                                            
- Query plans and SQL are sent to a third-party service by default with no consent — both explain_postgresql and explain_postgresql_format preferences default to True                                                                                           
- The new tab auto-focuses, stealing attention from the built-in Explain tab — the true flag in openTab forces focus away from the existing Explain panel on every query                                                                                         
- iframe with allow-scripts + allow-same-origin sandbox effectively negates the sandbox — a malicious/compromised endpoint could access pgAdmin's session                                                                                                        
- Module name "ep" is cryptic — other tools use descriptive names like sqleditor, schema_diff
- 10-second hang with no UI feedback when formatting API is unreachable — user sees nothing while waiting before fallback to local formatter                                                                                                                     
- send_post_request parse parameter is dead code — never called with parse=True
- `<p>Loading...</p>` instead of pgAdmin's standard Loader component for the explain panel loading state
- Trailing whitespace / inconsistent formatting in __init__.py
- No tests for the new module
- There are no documentation or screenshot updates

MGorkov:
@dpage Thank you for the thorough review and valuable feedback. I've addressed all the points raised:

**1. Domain name and licensing**
The service has been migrated from _explain-postgresql.com_ to _explain.tensor.ru_. A request for permission to use the "PostgreSQL" trademark in the original domain name has been submitted to the relevant parties.

**2. Data sent to third-party service**
The feature is now **disabled by default**. Documentation now includes a clear warning that enabling this feature will send query plans and SQL data to an external service.

**3. Tab opens on every EXPLAIN regardless of preference**
Fixed. The "Explain PostgreSQL" tab now only opens when the feature is enabled in Preferences (default: disabled). No external API calls or tab creation occur otherwise.

**4. Default values for preferences**
Both _explain_postgresql_ and _explain_postgresql_format_ now default to False.

**5. Tab auto-focus**
Since the tab only opens when the feature is explicitly enabled, focus behavior no longer interferes with the built-in Explain tab.

**6. iframe sandbox security**
The sandbox attribute has been adjusted to use only _allow-scripts_, removing _allow-same-origin_ to prevent potential session access by a compromised endpoint.

**7. Module name**
Renamed from _ep_ to _expl_pgsql_ to follow naming conventions (e.g., sqleditor, schema_diff).

**8. 10-second hang with no feedback**
Added a _Loader_ component that appears after 500ms when the API is unreachable, providing clear UI feedback during the fallback to the local formatter.

**9. send_post_request dead code**
The unused parse parameter has been removed.

**10. Loading state placeholder**
Replaced _\<p>Loading...\</p>_ with pgAdmin's standard Loader component.

**11. Trailing whitespace / formatting**
Fixed. Linter now passes without issues.

**12. Tests**
Added tests for the new module.

**13. Documentation**
Added documentation covering usage, configuration, and the warning about external data transmission.

This pull request is now closed. I will open a new PR with all the corrections shortly.

Thank you again for your time and detailed feedback.

Best regards

PR #9727, opened 2026-03-10
Fixed an issue where AI features are visible in the UI even when LLM_…

### LINKED ISSUE #9709: Activity state of AI
Not everyboy is happy about the introducion of AI functionallity and is therefore disabling AI.
It would be nice to see in the UI if the AI is enabled/disabled by the config*.py setting ('LLM_ENABLED')

It would be nice to have a note in the 'AI Assistent' of the Query Tool Tab if the AI is disabled.

Hiding that the 'AI Assistent' pane if AI is deactivated is of course the better solution.

dpage:
…ENABLED is set to False. Fixes #9709

## Summary by CodeRabbit

* **Bug Fixes**
  * AI features (AI Assistant tab, AI Reports menus, AI Preferences) now correctly hide when AI is disabled at the system level.

* **Improvements**
  * Safer Query Tool handling to prevent runtime errors during tab focus/navigation.
  * Added a browser-level configuration flag so AI menus and UI elements are shown only when system-enabled.

### DIFF
--- a/docs/en_US/ai_tools.rst
+++ b/docs/en_US/ai_tools.rst
@@ -48,15 +48,18 @@ button and select *AI*).
 Select your preferred LLM provider from the dropdown:
 
 **Anthropic**
-  Use Claude models from Anthropic. Requires an Anthropic API key.
+  Use Claude models from Anthropic, or any Anthropic-compatible API provider.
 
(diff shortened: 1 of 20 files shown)


dpage:
The test failure here is the flakey test fixed in #9723

Issue #9672, opened 2026-02-26
Generated always columns are not properly handled in View/Edit Data


pms967:
**Generated always columns are not properly handled in View/Edit Data**

Editing data in tables which includes one (or more) generated always column can be a nightmare.

If by accident you change the value of one such column(s), there seem to be no way to reset its value to "[default]". As a consequence, you cannot save your (other, wanted) changes since obviously that will result in an error:

> cannot insert a non-DEFAULT value into column "...."
Column "...." is a generated column.

As a consequence, you can not e.g. add row(s) to such a table by duplicating existing ones (copy/paste a row) and and then change the values in some of the (non generated) columns.

A similar issue exists when Importing Data from CSV files, though that perhaps can be handled by selecting which columns to import (but I am afraid that can work only if the generated column(s) are the very last one(s), after all the "normal" ones).

**To Reproduce**

Trivial:
1. Create a table including PK, some normal columns and at least one generated column (GENERATED ALWAYS AS ...);
2. Open it in the View/Edit Data tool; 
3. Select a row;
4. Click on the "Copy" icon;
5. Click on the "Paste" icon;
6. Change some of the non-generated columns (pkey and whatever);
7. try to save ==> Error 
8. would like to change the generate column to "[default]" to avoid that error: can't fine any way to do so.

**Expected behavior**

* Generated columns should be handled properly: attempts to directly change their value to anything but "DEFAULT" should be disabled, and any value pasted into them should be ignored (otherwise copy/paste of a whole row would be impossible).
 
* Queries generated by View/Edit Data must properly handle generated always columns in insert / update queries (always set them to DEFAULT, regardless of the corresponding cell content in the GUI). 

* Right-clicking on any "cell" should give the possibility to reset its value to [default] (and [null] as well).

Other possible (unrelated) improvement:

* cell selections should obey the standard modifiers: Shift to select a range of "cells", Ctrl to add a column to the existing selection, ecc.

**Desktop (please complete the following information):**
 - OS: Debian GNU/Linux 13 (trixie)
 - DE : KDE Plasma
 - pgAdmin version: 9.12
 - Mode: Desktop
 - Package type: DEB (from ftp.postgresql.org repository)

PR #9761, opened 2026-03-17
Fixed View/Edit Data not handling generated columns properly. #9672

RohitBhati8269:
## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL 12+ generated-column awareness in column metadata and the SQL editor, including display of generated/sequence info.

* **Bug Fixes**
  * Excluded GENERATED columns from INSERT/UPDATE payloads to avoid errors.
  * Ensured post-save refresh recalculates generated-column values after both INSERT and UPDATE by returning and consuming updated rows.

### DIFF
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/columns/sql/12_plus/nodes.sql
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/columns/sql/12_plus/nodes.sql
@@ -1,0 +1,40 @@
+SELECT DISTINCT att.attname as name, att.attnum as OID, pg_catalog.format_type(ty.oid,NULL) AS datatype,
+pg_catalog.format_type(ty.oid,att.atttypmod) AS displaytypname,
+att.attnotnull as not_null,
+CASE WHEN att.atthasdef OR att.attidentity != '' OR ty.typdefault IS NOT NULL THEN True
+ELSE False END as has_default_val, des.description, seq.seqtypid,
+{# Detect generated columns to exclude from INSERT/UPDATE in View/Edit Data #}
+CASE WHEN att.attgenerated = 's' THEN true ELSE false END as is_generated
+FROM pg_catalog.pg_attribute att
+    JOIN pg_catalog.pg_type ty ON ty.oid=atttypid
+    JOIN pg_catalog.pg_namespace tn ON tn.oid=ty.typnamespace
+    JOIN pg_catalog.pg_class cl ON cl.oid=att.attrelid
+    JOIN pg_catalog.pg_namespace na ON na.oid=cl.relnamespace
+    LEFT OUTER JOIN pg_catalog.pg_type et ON et.oid=ty.typelem
+    LEFT OUTER JOIN pg_catalog.pg_attrdef def ON adrelid=att.attrelid AND adnum=att.attnum
+    LEFT OUTER JOIN (pg_catalog.pg_depend JOIN pg_catalog.pg_class cs ON classid='pg_class'::regclass AND objid=cs.oid AND cs.relkind='S') ON refobjid=att.attrelid AND refobjsubid=att.attnum
+    LEFT OUTER JOIN pg_catalog.pg_namespace ns ON ns.oid=cs.relnamespace
+    LEFT OUTER JOIN pg_catalog.pg_index pi ON pi.indrelid=att.attrelid AND indisprimary
+    LEFT OUTER JOIN pg_catalog.pg_description des ON (des.objoid=att.attrelid AND des.objsubid=att.attnum AND des.classoid='pg_class'::regclass)
+    LEFT OUTER JOIN pg_catalog.pg_sequence seq ON cs.oid=seq.seqrelid
+WHERE
+
+{% if tid %}
+    att.attrelid = {{ tid|qtLiteral(conn) }}::oid
+{% endif %}
+{% if table_name and table_nspname %}
+    cl.relname= {{table_name |qtLiteral(conn)}} and na.nspname={{table_nspname|qtLiteral(conn)}}
+{% endif %}
+{% if clid %}
+    AND att.attnum = {{ clid|qtLiteral(conn) }}
+{% endif %}
+{### To show system objects ###}
+{% if not show_sys_objects and not has_oids %}
+    AND att.attnum > 0
+{% endif %}
+{### To show oids in view data ###}
+{% if has_oids %}
+    AND (att.attnum > 0 OR (att.attname = 'oid' AND att.attnum < 0))
+{% endif %}
+    AND att.attisdropped IS FALSE
+ORDER BY att.attnum
--- a/web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
@@ -1309,8 +1309,10 @@ export function ResultSet() {
       }
 
       pageDataOutOfSync.current = true;
-      if(_.size(dataChangeStore.added)) {
-        // Update the rows in a grid after addition
+      // Update the rows in a grid after addition/update.
+      // row_added contains refetched row data with recalculated
+      // generated column values (for both INSERT and UPDATE).
+      if(_.size(dataChangeStore.added) || _.size(dataChangeStore.updated)) {
         respData.data.query_results.forEach((qr)=>{
           if(!_.isNull(qr.row_added)) {
             let rowClientPK = Object.keys(qr.row_added)[0];
--- a/web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/update.sql
+++ b/web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/update.sql
@@ -4,4 +4,7 @@ UPDATE {{ conn|qtIdent(nsp_name, object_name) | replace("%", "%%") }} SET
 {% if not loop.first %}, {% endif %}{{ conn|qtIdent(col) | replace("%", "%%") }} = %({{ pgadmin_alias[col] }})s{% if type_cast_required[col] %}::{{ data_type[col] }}{% endif %}{% endfor %}
  WHERE
 {% for pk in primary_keys %}
-{% if not loop.first %} AND {% endif %}{{ conn|qtIdent(pk) | replace("%", "%%") }} = {{ primary_keys[pk]|qtLiteral(conn) }}{% endfor %};
+{% if not loop.first %} AND {% endif %}{{ conn|qtIdent(pk) | replace("%", "%%") }} = {{ primary_keys[pk]|qtLiteral(conn) }}{% endfor %}
+{# Return primary keys to refetch row with recalculated generated column values #}
+{% if pk_names and not has_oids %} RETURNING {{pk_names | replace("%", "%%")}}{% endif %}
+{% if has_oids %} RETURNING oid{% endif %};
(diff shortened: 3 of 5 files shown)


web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/update.sql
```diff
@@ -4,4 +4,7 @@ UPDATE {{ conn|qtIdent(nsp_name, object_name) | replace("%", "%%") }} SET
 {% if not loop.first %}, {% endif %}{{ conn|qtIdent(col) | replace("%", "%%") }} = %({{ pgadmin_alias[col] }})s{% if type_cast_required[col] %}::{{ data_type[col] }}{% endif %}{% endfor %}
  WHERE
 {% for pk in primary_keys %}
-{% if not loop.first %} AND {% endif %}{{ conn|qtIdent(pk) | replace("%", "%%") }} = {{ primary_keys[pk]|qtLiteral(conn) }}{% endfor %};
+{% if not loop.first %} AND {% endif %}{{ conn|qtIdent(pk) | replace("%", "%%") }} = {{ primary_keys[pk]|qtLiteral(conn) }}{% endfor %}
+{# Return primary keys to refetch row with recalculated generated column values #}
+{% if pk_names and not has_oids %} RETURNING {{pk_names | replace("%", "%%")}}{% endif %}
```
adityatoshniwal: why not return the complete row instead of just oid? That way - no need to reselect.

web/pgadmin/tools/sqleditor/utils/save_changed_data.py
```diff
@@ -118,6 +118,12 @@ def save_changed_data(changed_data, columns_info, conn, command_obj,
                 if command_obj.has_oids():
                     data.pop('oid', None)
 
+                # Remove generated columns (GENERATED ALWAYS AS) as they
+                # cannot be inserted - PostgreSQL auto-computes their values.
+                for col_name, col_info in columns_info.items():
```
adityatoshniwal: so this will run for each row? why not simply skip from update sql generation - it is already looping through the columns. Can avoid this loop inside loop.

web/pgadmin/tools/sqleditor/utils/save_changed_data.py
```diff
@@ -163,14 +169,38 @@ def save_changed_data(changed_data, columns_info, conn, command_obj,
...
+            pk_names, primary_keys = command_obj.get_primary_keys()
+
             for each_row in changed_data[of_type]:
                 data = changed_data[of_type][each_row]['data']
+                row_primary_keys = changed_data[of_type][each_row][
+                    'primary_keys']
+
+                # Remove generated columns (GENERATED ALWAYS AS) as they
+                # cannot be updated - PostgreSQL auto-computes their values.
+                for col_name, col_info in columns_info.items():
```
adityatoshniwal: same as above.

web/pgadmin/tools/sqleditor/utils/save_changed_data.py
```diff
@@ -180,12 +210,35 @@ def save_changed_data(changed_data, columns_info, conn, command_obj,
...
                     conn=conn
                 )
-                list_of_sql[of_type].append({'sql': sql,
-                                             'data': data,
-                                             'row_id':
-                                                 data.get(client_primary_key)})
+
+                # For tables with generated columns, add select_sql to
+                # refetch updated row.
+                if has_generated_cols:
```
adityatoshniwal, line 223: like I mentioned above - check if we can use the returning clause to return all instead of selecting the row again.

adityatoshniwal (changes requested):

web/pgadmin/tools/sqleditor/utils/save_changed_data.py
```diff
@@ -175,14 +178,33 @@ def save_changed_data(changed_data, columns_info, conn, command_obj,
...
             for each_row in changed_data[of_type]:
                 data = changed_data[of_type][each_row]['data']
+                row_primary_keys = changed_data[of_type][each_row][
+                    'primary_keys']
+
+                # Remove generated columns (GENERATED ALWAYS AS) as they
+                # cannot be updated - PostgreSQL auto-computes their values.
+                data = {k: v for k, v in data.items()
+                        if not columns_info.get(k, {}).get('is_generated',
+                                                           False)}
```
dpage, line 198: **Must fix: this can empty `data`, which renders invalid SQL.**

Nothing guards against the filtered `data` becoming empty, and `update.sql` has no `{% if data_to_be_saved %}` around the `SET` list, so if the user edits *only* the generated cell (which the grid still permits) the rendered statement is:

```sql
UPDATE "public"."t" SET  WHERE "id" = 1 RETURNING *;
```

Confirmed against PostgreSQL 18.4:

```
ERROR:  syntax error at or near "WHERE"
LINE 1: UPDATE t9761 SET  WHERE id = 1 RETURNING *;
```

This is the reporter's own "if by accident you change the value of one such column" scenario, and before this change it at least produced a self-explanatory `column "g" can only be updated to DEFAULT`, so as it stands it trades a clear server error for a malformed statement generated by pgAdmin. Skipping the row when there is nothing left to set (`if not data: continue`) would cover it.

web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/columns/sql/12_plus/nodes.sql
```diff
@@ -0,0 +1,40 @@
+SELECT DISTINCT att.attname as name, att.attnum as OID, pg_catalog.format_type(ty.oid,NULL) AS datatype,
+pg_catalog.format_type(ty.oid,att.atttypmod) AS displaytypname,
+att.attnotnull as not_null,
+CASE WHEN att.atthasdef OR att.attidentity != '' OR ty.typdefault IS NOT NULL THEN True
+ELSE False END as has_default_val, des.description, seq.seqtypid,
+{# Detect generated columns to exclude from INSERT/UPDATE in View/Edit Data #}
```
dpage, line 6: **Must fix: this misses virtual generated columns on PostgreSQL 18.**

PostgreSQL 18 added virtual generated columns, which carry `attgenerated = 'v'` and reject non-`DEFAULT` values exactly as stored ones do, so on a supported server the bug this PR fixes is still reachable. Verified locally:

```
 attname | attgenerated
 g       | v
ERROR:  cannot insert a non-DEFAULT value into column "g"
DETAIL:  Column "g" is a generated column.
```

`CASE WHEN att.attgenerated <> '' THEN true ELSE false END` covers both kinds and will not need revisiting if another is ever added. (`columns/sql/default/properties.sql` has the same `in ('s')` gap, but that is separate work and not yours to fix here.)

web/pgadmin/tools/sqleditor/templates/sqleditor/sql/default/update.sql
```diff
@@ -4,4 +4,6 @@ UPDATE {{ conn|qtIdent(nsp_name, object_name) | replace("%", "%%") }} SET
 {% if not loop.first %}, {% endif %}{{ conn|qtIdent(col) | replace("%", "%%") }} = %({{ pgadmin_alias[col] }})s{% if type_cast_required[col] %}::{{ data_type[col] }}{% endif %}{% endfor %}
  WHERE
 {% for pk in primary_keys %}
-{% if not loop.first %} AND {% endif %}{{ conn|qtIdent(pk) | replace("%", "%%") }} = {{ primary_keys[pk]|qtLiteral(conn) }}{% endfor %};
+{% if not loop.first %} AND {% endif %}{{ conn|qtIdent(pk) | replace("%", "%%") }} = {{ primary_keys[pk]|qtLiteral(conn) }}{% endfor %}
+{# Return complete row to get recalculated generated column values #}
+{% if return_all_columns %} RETURNING *{% endif %};
```
dpage, line 9: `RETURNING *` reads every column of the base table rather than the columns the result set actually has, which has two side effects worth avoiding.

In a Query Tool editable result set with aliased columns (`SELECT a AS foo ...`) the returned keys are the base-table names, so the merge in `ResultSet.jsx` will not refresh the aliased column and instead spreads unrelated base-column keys into the row object. Separately, a role holding `UPDATE` but only column-level `SELECT` on the displayed columns will now get a permission-denied error for a statement that previously succeeded, because `RETURNING *` has to read everything.

Returning just the generated columns is enough to refresh their values and sidesteps both.

web/pgadmin/tools/sqleditor/utils/save_changed_data.py
```diff
@@ -192,12 +214,26 @@ def save_changed_data(changed_data, columns_info, conn, command_obj,
...
+                        'client_row': each_row,
+                        'returning_all': True,
+                        'row_id': data.get(client_primary_key)
+                    })
+                else:
+                    list_of_sql[of_type].append({
+                        'sql': sql,
+                        'data': data,
+                        'row_id': data.get(client_primary_key)
+                    })
```
dpage, line 236: Minor: the two branches differ only by two keys, so a single dict literal plus a conditional `update()` (or simply always setting `client_row` and passing `returning_all=has_generated_cols`) would read more easily. The comment above the `render_template()` call at line 206 also describes this branch rather than the call it sits above.

web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
```diff
@@ -1309,8 +1309,10 @@ export function ResultSet() {
       }
 
       pageDataOutOfSync.current = true;
-      if(_.size(dataChangeStore.added)) {
-        // Update the rows in a grid after addition
+      // Update the rows in a grid after addition/update.
+      // row_added contains refetched row data with recalculated
+      // generated column values (for both INSERT and UPDATE).
+      if(_.size(dataChangeStore.added) || _.size(dataChangeStore.updated)) {
```
dpage, line 1315: This is right, but note it only has an effect when the table has generated columns, since `row_added` is only populated for updates in that case. Worth a word in the comment so the next reader does not go hunting for the missing refresh on ordinary updates.

dpage (commented):
Thanks for picking this up, and for reworking it around Aditya's suggestion to use `RETURNING` rather than a second `SELECT`: that is much the better shape, and the INSERT path does now fix the copy/paste-a-row case from the issue. I have reviewed it against current `master` (it merges cleanly, incidentally, so CodeRabbit's merge-conflict note is stale) and checked the behaviour against a local PostgreSQL 18.4, and there are a couple of things that still need attention before this can go in, because in both cases the original symptom is still reachable.

The two blocking ones are inline below: an `UPDATE` whose `SET` list becomes empty renders invalid SQL, and virtual generated columns on PostgreSQL 18 are not detected at all. Beyond those, the part of the issue that asks for direct editing of generated columns to be *disabled* is not implemented yet, so a value typed into such a cell is now silently discarded rather than refused, and there are no tests.

**Generated columns should be read-only in the grid.** `get_columns_types()` mutates the `col` dicts in place, so `is_generated` already reaches the client in `colinfo`; that means `getFinalColumn()` in `ResultSet.jsx` (around line 628) only needs `'can_edit': (c.name == 'oid' || c.is_generated) ? false : isEditable`. That gives the user the feedback the issue asks for, and it also tidies the copy/paste-a-row flow, because `triggerSaveData()` already strips `can_edit === false` keys from the new-row payload client-side. It would additionally make the empty-`SET` case below unreachable through the UI, although I would still guard it in the backend.

**Tests.** `web/pgadmin/tools/sqleditor/utils/tests/test_save_changed_data.py` already has the scenario scaffolding this needs, and three cases would pin the behaviour down nicely: inserting into a table with a generated column, updating a non-generated column on such a table and asserting the recalculated value comes back, and updating only the generated column.

**On the new template.** Duplicating `default/nodes.sql` into a `12_plus` bucket to add one `CASE` is the established pattern, so I am not going to argue with it, but do note the consequence: since supported servers are PostgreSQL 14 to 18, `default/nodes.sql` becomes unreachable for every version we support, and a future catalog fix applied there will silently not take effect. Either put the `CASE` straight into `default/nodes.sql` (`attgenerated` has existed since 12, and we do not hard-block older servers, so only a genuinely pre-12 connection would notice), or keep the bucket and add a comment to both files noting that they must be kept in step.

One smaller thing whilst you are in `ResultSet.jsx`: the merge at line 1320 takes `rowIdx` from `findIndex` without guarding `-1`, in which case `slice(0, -1)` drops the last row and appends a spread of `undefined`. That is pre-existing on the insert path, but this change routes updates through it too, so an early `if (rowIdx < 0) return prevRows;` is worth adding.

Finally, for the record, the one red CI job (`run-python-tests-pg (windows-latest, 17)`) is infrastructure rather than anything you have done: `psycopg-c` fails to link with `LNK1181: cannot open input file 'libpq.lib'`. `pycodestyle` passes on both Python files.

PR #9769, opened 2026-03-21
Add toast_tuple_target to Materialized View properties dialog (#9626)

### LINKED ISSUE #9626: In the properties window of the materialized view parameter `toast_tuple_target` is not present
**Describe the bug**

After I put `toast_tuple_target ` on the materialized view, I don't see `toast_tuple_target ` in the properties window.

Image

```
ALTER TABLE test_mv SET (toast_tuple_target = 128);
```

**To Reproduce**

1. Run SQL commands
```
CREATE TABLE test AS
	SELECT generate_series(1,1000) AS id, REPEAT('a', 8000) AS descr;

CREATE MATERIALIZED VIEW test_mv AS (
	SELECT * FROM test WHERE id > 5
);

ALTER TABLE test_mv SET (toast_tuple_target = 128);

REFRESH MATERIALIZED VIEW test_mv;
```
2. Go to `Materialized Views` -> `test_mv` -> `Right click` -> `Properties` -> `Definition` tab
3.  I don't see the field `toast_tuple_target` in the definition tab
4. I cannot modify `toast_tuple_target` in the definition tab

**Expected behavior**

I should see the `toast_tuple_target` field in the definition tab, and I should modify it.

**Screenshots**

Image

 - OS: Windows 11 x64
 - pgAdmin 9.12
 - Mode: Desktop
 - Browser Google Chrome
 - Package type: pgadmin4-9.12-x64.exe
 - Postgresql 18.1

**Additional context**

psql / pg_dump  shows everything fine

```
#psql test=# \d+ test_mv
                                   Materialized view "public.test_mv"
 Column |  Type   | Collation | Nullable | Default | Storage  | Compression | Stats target | Description
--------+---------+-----------+----------+---------+----------+-------------+--------------+-------------
 id     | integer |           |          |         | plain    |             |              |
 descr  | text    |           |          |         | extended |             |              |
View definition:
 SELECT id,
    descr
   FROM test
  WHERE id > 5;
Access method: heap
Options: toast_tuple_target=128


#pg_dump -Upostgres -t test_mv --schema-only test

CREATE MATERIALIZED VIEW public.test_mv
WITH (toast_tuple_target='128') AS
 SELECT id,
    descr
   FROM public.test
  WHERE (id > 5)
  WITH NO DATA;

```

pavanmanishd:
Closes #9626

## Problem

The `toast_tuple_target` storage parameter (PostgreSQL 11+) is supported by materialized views via `ALTER MATERIALIZED VIEW ... SET(TOAST_TUPLE_TARGET = N)`, but the pgAdmin Materialized View properties dialog has no way to view or change it.

## Changes

- **properties.sql** — extract `toast_tuple_target` from `c.reloptions` (same regex pattern as `fillfactor`)
- **create.sql** — include `TOAST_TUPLE_TARGET` in the `WITH` clause when set; extend the condition that decides whether a `WITH` block is needed
- **update.sql** — `SET` / `RESET` `TOAST_TUPLE_TARGET` when the value changes (mirrors the `FILLFACTOR` block exactly)
- **mview.ui.js** — integer input field (min 128, max 8160) in the Definition group, `min_version: 110000`

### DIFF
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/views/static/js/mview.ui.js
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/views/static/js/mview.ui.js
@@ -106,6 +106,11 @@ export default class MViewSchema extends BaseUISchema {
         id: 'fillfactor', label: gettext('Fill factor'),
         group: gettext('Definition'), mode: ['edit', 'create'],
         noEmpty: false, type: 'int', controlProps: {min: 10, max: 100}
+      },{
+        id: 'toast_tuple_target', label: gettext('Toast tuple target'),
+        group: gettext('Definition'), mode: ['edit', 'create'],
+        noEmpty: false, type: 'int', controlProps: {min: 128, max: 8160},
+        min_version: 110000,
       },
       {
         id: 'dependsonextensions',
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/views/templates/mviews/pg/default/sql/create.sql
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/views/templates/mviews/pg/default/sql/create.sql
@@ -7,11 +7,13 @@
 {% endif %}
 {% if data.name and data.schema and data.definition %}
 CREATE MATERIALIZED VIEW{% if add_not_exists_clause %} IF NOT EXISTS{% endif %} {{ conn|qtIdent(data.schema, data.name) }}
-{% if(data.fillfactor or data.autovacuum_enabled in ('t', 'f') or data.toast_autovacuum_enabled in ('t', 'f') or data['vacuum_data']|length > 0) %}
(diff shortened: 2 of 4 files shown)


dpage:
Merged to `master` as c720a33b4 (squashed, with you as author). Thanks @pavanmanishd!

One adjustment during merge: the original change only touched the `mviews/pg/default/` SQL bucket, which serves **PostgreSQL < 12** — so it had no effect on any supported server. Supported versions (13–18) resolve to the `13_plus`/`15_plus` buckets (17_plus falls back to 15_plus), which carry their own standalone `create`/`properties`/`update.sql`. The same toast_tuple_target changes are now applied to the `12_plus`, `13_plus` and `15_plus` buckets as well.

Verification: template output is **byte-identical** to the previous SQL when `toast_tuple_target` is unset (no change to existing behaviour), and the generated CREATE/ALTER SET/ALTER RESET/recreate forms were confirmed to execute on a live PostgreSQL 18 server with the expected `reloptions`. A 9.16 release note was added.

PR #9781, opened 2026-03-23
Fix a couple of related issues in the Query Tool layout.

dpage:
- Hide the AI Assistant tab if AI is disabled or uncofnigured.
- Ensure the AI Assistant tab is not the first one shown (that should be the Query tab - saved layouts may need a reset).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented errors when no saved layout exists and preserved the default layout.
  * Restored prior tab focus after adding missing panels to avoid unexpected focus shifts.
  * Adjusted Query Tool tab ordering to improve navigation flow and consistency.

* **Improvements**
  * LLM availability is now checked dynamically at request time for more accurate feature exposure.

### DIFF
--- a/web/pgadmin/static/js/helpers/Layout/index.jsx
+++ b/web/pgadmin/static/js/helpers/Layout/index.jsx
@@ -220,6 +220,10 @@ export class LayoutDocker {
   }
 
   loadLayout(savedLayout) {
+    if (!savedLayout) {
+      // No saved layout - DockLayout already initialized with defaultLayout
+      return;
+    }
     try {
       this.layoutObj.loadLayout(JSON.parse(savedLayout));
       this.addMissingDefaultPanels();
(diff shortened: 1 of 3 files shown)


web/pgadmin/static/js/helpers/Layout/index.jsx
```diff
@@ -220,12 +220,15 @@ export class LayoutDocker {
...
+    if (!savedLayout) {
+      // No saved layout - DockLayout already initialized with defaultLayout
+      return;
+    }
     try {
       this.layoutObj.loadLayout(JSON.parse(savedLayout));
       this.addMissingDefaultPanels();
     } catch {
-      /* Fallback to default */
-      this.layoutObj.loadLayout(this.defaultLayout);
```
adityatoshniwal, line 228: this is good to have in cases where savedLayout is corrupted.

web/pgadmin/static/js/helpers/Layout/index.jsx
```diff
@@ -252,6 +255,22 @@ export class LayoutDocker {
...
+    const collectActiveIds = (box) => {
+      box.children.forEach((child) => {
+        if (child.children) {
+          collectActiveIds(child);
+        } else if (child.activeId) {
+          savedActiveIds.push(child.activeId);
+        }
+      });
+    };
+    collectActiveIds(this.layoutObj.getLayout().dockbox);
```
adityatoshniwal, line 273: this should have been already taken care by the rc-dock.

dpage: Unfortunately not - when openTab is called with 'middle' direction, rc-dock's addTabToPanel unconditionally sets activeId to the last newly added tab, stealing focus.

web/pgadmin/static/js/helpers/Layout/index.jsx
dpage: Good point - I've put it back.

PR #9791, opened 2026-03-26
Fixed python & feature test failures caused by Werkzeug 3.1.7 rejecting empty Host header in CSRF token generation.

anilsahoo20:
## Summary by CodeRabbit

* **Tests**
  * Improved CSRF handling in test utilities: test requests now include a sensible default host and explicit HTTP scheme, and session cookies are saved more reliably. This makes cookie/session behavior in tests more closely match production.

### DIFF
--- a/web/regression/python_test_utils/csrf_test_client.py
+++ b/web/regression/python_test_utils/csrf_test_client.py
@@ -93,9 +93,10 @@ class TestClient(testing.FlaskClient):
         # this test client, such as the secure cookie that
         # powers `flask.session`,
         # and make a test request context that has those cookies in it.
+        server_name = self.app.config.get("SERVER_NAME") or "localhost"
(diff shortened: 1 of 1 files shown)


web/regression/python_test_utils/csrf_test_client.py
```diff
@@ -94,7 +94,8 @@ def generate_csrf_token(self, *args, **kwargs):
         # powers `flask.session`,
         # and make a test request context that has those cookies in it.
         environ_overrides = {
-            'wsgi.url_scheme': ''
+            'wsgi.url_scheme': 'http',
+            'HTTP_HOST': 'localhost',
```
khushboovashi:
```suggestion
            'HTTP_HOST':  current_app.config["SERVER_NAME"] or "localhost"
```

Issue #9700, opened 2026-03-06
PGAdmin should attempt to refresh the database if the user tries to interact with a database that's disconnected


stuaxo:
As a backend developer I often `dropdb, createdb + repopulate` when I come back to the object view I have to right click and refresh, even though I have a new database with exactly the same name as the previous one.

It would be good if pgadmin4 could (optionally) try and reconnect at least once - if I do something in the object view + it's lost it's connection.

This would be good in other work-related circumstances too - there's a VPN and so I periodically have to reconnect things, and pgadmin is yet another place I have to do a reconnection ceremony that I'd like to have automated.

Issue #8279, opened 2024-12-18
Major Pain: Make the connection lost warning an option


usernamesareaholdoverfromthe90s:
**Describe the solution you'd like**

My connections to the database the drop relatively quickly.

So I'm constantly getting the "Connection Warning: Cancel/Continue" message box.

E.g., During the daily development and support of my application I may have a half dozen select queries (or "View/Edit Data" | "Filtered Rows") windows open, at any time.

And every time I dive into the code to look up a potential bug, or spend time responding to a support email, the connections to the database will frequently time out.

So when I switch back to pgAdmin, I have to click through the "Connection Warning: Cancel/Continue" message box, on all my query tabs.  This can happen a dozen times a day.

So I'm frequently seeing this dialog 50 times a day, which is over 10000 times per year.

I understand from previous chats with Dave Page that it is possible that some databases might be configured to do something special when a connection happens, so the dialog has to be shown to prevent that connection side-effect from happening.

But... I've had to click through this dialog 10000 times this year, 10000 times last year, 10000 the year before (and for the last 12 years!), and I'll have to click through it 10000 times next year, and 10000 times the following year... and never EVER will I do anything other than click "Continue".

Given that the scenario Dave is concerned about is probably pretty rare, can we not just provide a _preference_ to auto-reconnect, so those of us who don't have to worry about connection side-effects don't have to have our workflow interrupted by this irrelevant dialog time tens of thousands of times.

**Describe alternatives you've considered**

Add a new "Auto reconnect" toggle switch (default = off, so as not to change the behaviour for existing users), under "File" | "Preferences" | "Query Tool" | "Options".

**Additional context**

The usability of pgAdmin used to be awful, but it has wonderfully improved over the last few years.  Awesome!  This is now probably one of the biggest remaining usability impediments (i.e., 10000 unnecessary clicks, per year).

alexykot:
I would go further and make it a per-server configuration, not global. 

And yes, disabled by default, but available if you know there is no problem with just auto-reconnecting transparently. 

Even better if there would be another button in this dialog saying something like "Autoreconnect and don't show this again".

Issue #6308, opened 2023-05-22
When idle, connection is lost and cannot be regained. Restart necessary


birgersp:
**Describe the bug**

If I leave pgAdmin open for a while (e.g. 1 hour), the connection the database seems to be lost. If I try to view any information about the database (e.g. expand the Schemas object in the Object Explorer), pgAdmin just hangs for eternity. It seems to be waiting for a re-connection that never completes.

I just have to close pgAdmin and re-start it so that it re-connects properly.

**To Reproduce**

1. Connect to a server
2. Browse the database in any way (for instance, expand the "Tables" object to view the tables of the db)
3. Leave pgAdmin open but use other software on your computer, for about 30 min.
4. Go back to pgAdmin and expand a table to view its columns.
5. Observe how pgAdmin now just shows spinning wheels as if it is doing some kind of connection or processing.
6. Wait for 10 minutes, get frustrated, close and re-open pgAdmin.

**Expected behavior**

pgAdmin should be able to restore (or re-create) the connection to the server.

**Desktop (please complete the following information):**
 - OS: Pop Os (Ubunutu 22.04)
 - Version: 7.1
 - Mode: Desktop
 - Package type: DEB

TomKein:
Why does pgadmin even ask about reconnect? Why not just reconnect automatically after losing connection?
Does anyone like to click a hundred times a day - "yes, please reconnect, I don't need it without connection!

malix0:
Me and my colleagues had another problem, if the database doesn't respond because for example is busy, then pgAdmin also hangs for eternity. Closing and reopening the program does not guarantee a success because this can happen many times until DB start responding quickly.

Vinaum8:
My problem is the version Web. PgAdmin stay open for minutes and when idle they stop responding.

-------------------------------------------------------

**Web (please complete the following information):**

- OS: EKS with Amazon Linux 2
- Version: 8.5
- Mode: Web
- Package type: Helm Chart

**Helm Chart info:**
repoURL: https://helm.runix.net
Version: 1.24.0

Ingress controller: Nginx.

yogeshmahajan-1903:
@birgersp , @malix0 , @Vinaum8 
I am not able to reproduce the issue. Can you please check with latest version 8.14 and check? If issue persists, kindly share pgadmin logs and pgadmi configuration.

yogeshmahajan-1903:
No response from author hence closing.

PR #9800, opened 2026-03-29
Fixed infinite loading after idle connection loss on Linux Desktop. #6308

ZoroXL:
When pgAdmin is left idle for an extended period on Linux Desktop, the database connection gets silently dropped. Instead of showing the reconnect dialog, the application hangs with an infinite spinner.

Fixes:
- Added `connection_ping()` to detect stale/half-open TCP connections that `connected()` misses (it only checks local state).
- Used `connection_ping()` in the Object Explorer `children()` endpoint to detect dead connections before they cause a hang.
- Added TCP keepalive defaults to all database connections so the OS detects dead sockets faster.
- Added CONNECTION_LOST handling in the Object Explorer tree with a reconnect dialog.
- Query Tool now checks connection status immediately when the tab becomes visible again.

Refs #6308, #9700, #8279

## Summary by CodeRabbit

* **New Features**
  * Reconnection dialog when a server connection is detected as lost; server node shows "not connected" state and can be closed automatically.
  * Query tool refreshes connection status on tab visibility and uses refactored, more reliable polling.

* **Bug Fixes**
  * Improved connection health checks to detect stale TCP sessions and avoid relying on stale local state.
  * Default TCP keepalive parameters added for more stable long-lived connections.

### DIFF
--- a/web/pgadmin/utils/driver/psycopg3/connection.py
+++ b/web/pgadmin/utils/driver/psycopg3/connection.py
@@ -1413,6 +1413,31 @@ WHERE db.datname = current_database()""")
             self.conn = None
         return False
 
+    def connection_ping(self):
+        """
+        Check if the connection is actually alive by executing a lightweight
+        query.  Unlike connected(), which only inspects local state, this
+        sends traffic to the server and will detect stale / half-open TCP
+        connections that were silently dropped by firewalls or the OS while
+        pgAdmin was idle.
+
+        Returns True if alive, False otherwise.
+        """
+        if not self.connected():
+            return False
+        try:
+            cur = self.conn.cursor()
+            cur.execute("SELECT 1")
+            cur.close()
+            return True
+        except Exception:
+            try:
+                self.conn.close()
+            except Exception:
+                pass
+            self.conn = None
+            return False
+
     def _decrypt_password(self, manager):
         """
         Decrypt password
--- a/web/pgadmin/utils/driver/psycopg3/connection.py
+++ b/web/pgadmin/utils/driver/psycopg3/connection.py
@@ -1413,46 +1413,6 @@ WHERE db.datname = current_database()""")
             self.conn = None
         return False
 
-    def connection_ping(self):
-        """
-        Check if the connection is actually alive by executing a lightweight
-        query.  Unlike connected(), which only inspects local state, this
-        sends traffic to the server and will detect stale / half-open TCP
-        connections that were silently dropped by firewalls or the OS while
-        pgAdmin was idle.
-
-        Returns True if alive, False otherwise.
-        """
-        if not self.connected():
-            return False
-
-        try:
-            # Check the transaction status before executing the ping
-            # query.  If a query is already in progress (ACTIVE) or we
-            # are inside a transaction block (INTRANS / INERROR), running
-            # SELECT 1 would fail or disrupt the ongoing operation.  In
-            # those states the connection is evidently alive, so just
-            # return True.
-            txn_status = self.conn.info.transaction_status
-            if txn_status != 0:
-                # 0 = IDLE — safe to send a query
-                # 1 = ACTIVE — command in progress, connection is alive
-                # 2 = INTRANS — in transaction block, connection is alive
-                # 3 = INERROR — in failed transaction, connection is alive
-                return True
-
-            cur = self.conn.cursor()
-            cur.execute("SELECT 1")
-            cur.close()
-            return True
-        except Exception:
-            try:
-                self.conn.close()
-            except Exception:
-                pass
-            self.conn = None
-            return False
-
     def _decrypt_password(self, manager):
         """
         Decrypt password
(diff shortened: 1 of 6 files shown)


ZoroXL:
@coderabbitai review

ZoroXL:
Hey @akshay-joshi, can you review this? It fixes the idle connection hang on Linux Desktop (#6308, #9700, #8279). Thanks!

web/pgadmin/utils/driver/psycopg3/connection.py
```diff
@@ -1413,6 +1413,31 @@ def connected(self):
...
+            cur.execute("SELECT 1")
+            cur.close()
+            return True
+        except Exception:
+            try:
+                self.conn.close()
+            except Exception:
+                pass
+            self.conn = None
+            return False
```
asheshv, line 1454: This would fail or disrupt long running query execution (if in progress).
You might want to check the current transaction `transaction_status` of the connection before running this query.

asheshv (changes requested):

ZoroXL:
@asheshv Addressed your feedback — `connection_ping()` now checks `transaction_status` before running `SELECT 1`. If the connection is not idle (ACTIVE, INTRANS, or INERROR), it skips the ping and returns `True` since the connection is evidently alive. Also fixed the response status to return 503 `CONNECTION_LOST` instead of 500 when reconnect fails, so the tree reconnect dialog triggers properly.

asheshv:
### Code review

Found 3 issues:

1. **Client-side timeout silently bypasses reconnect dialog.** When `connection_ping()` blocks on a recently-dead connection (within ~60s before keepalives detect it), `SELECT 1` hangs and the client's 30s axios timeout fires. Axios timeout errors have `error.response === undefined`, so the condition `error.response?.status === 503 && error.response?.data?.info === 'CONNECTION_LOST'` is never true — the user sees a generic "Connection to pgAdmin server has been lost" toast instead of the reconnect dialog. This doesn't affect the primary use case (extended idle, where keepalives pre-detect the dead socket), but it's a gap for recent disconnections.

   **Suggested fix** — also handle timeout errors:
   ```typescript
   if ((error.response?.status === 503 && error.response?.data?.info === 'CONNECTION_LOST') ||
       error.code === 'ECONNABORTED') {
   ```

   https://github.com/pgadmin-org/pgadmin4/blob/3f3d165a4b53b98ebb08ba692b2b98cf6f1a3444/web/pgadmin/static/js/tree/tree_nodes.ts#L129-L135

2. **Multiple reconnect dialogs stack when server has expanded children.** If the server node has multiple expanded databases, each in-flight child-loading request fails independently and triggers its own `notifier.confirm()` dialog. `tree.close(serverNode)` prevents future child loads but requests already in-flight still fail. The user could see 3+ stacked "Connection lost" dialogs.

   **Suggested fix** — add a per-server dedup flag:
   ```typescript
   if (sData._reconnectPending) return;
   sData._reconnectPending = true;
   ```
   Clear it in both the OK and Cancel callbacks.

   https://github.com/pgadmin-org/pgadmin4/blob/3f3d165a4b53b98ebb08ba692b2b98cf6f1a3444/web/pgadmin/static/js/tree/tree_nodes.ts#L137-L161

3. **Improve existing `ping()` instead of adding a new `connection_ping()` method.** The existing `ping()` is a bare `execute_scalar('SELECT 1')` with no transaction status check and no error recovery — it would fail or disrupt an active query. Rather than adding a separate `connection_ping()` to the abstract class, improve `ping()` itself with the transaction status check and cleanup logic, then use it in the `children()` endpoint. This avoids two overlapping methods on the abstract class and makes the safer behavior the default for the existing caller (`driver.ping()` in `misc/__init__.py:259`).

   https://github.com/pgadmin-org/pgadmin4/blob/3f3d165a4b53b98ebb08ba692b2b98cf6f1a3444/web/pgadmin/utils/driver/psycopg3/connection.py#L1487-L1489

ZoroXL:
@asheshv Thanks for the review. Addressed all three in a9ad295d2:

1. **Timeout bypass** — added `error.code === 'ECONNABORTED'` to the connection-lost check so axios client-side timeouts also trigger the reconnect dialog (axios timeouts have no `error.response`, so the previous 503 condition was unreachable for that path).
2. **Dialog stacking** — added a per-server `_reconnectPending` guard. It's set when the first failing child-load opens the dialog and cleared in both the OK and Cancel callbacks, so concurrent in-flight failures no longer stack multiple dialogs.
3. **Method consolidation** — removed `connection_ping()` from the abstract class and the psycopg3 implementation. The transaction-status guard and close-on-failure recovery are now folded into the existing `ping()` (return type changed to `bool`; verified there are no existing callers of `conn.ping()` in the codebase). `browser/utils.py` now calls `conn.ping()`.

Re-requesting review.

ZoroXL:
Friendly ping @asheshv @akshay-joshi — I addressed all three review points in a9ad295d2 about a month ago and CI is green. Let me know if there's anything else you'd like me to change. Thanks!

dpage:
Merged to `master` as 77dc9d77e (squashed, with you as author). Thanks @ZoroXL for the thorough iteration, and @asheshv for the review.

Verified all prior review points are addressed in the merged revision (the `ECONNABORTED` timeout handling, the `_reconnectPending` dialog dedup, and `ping()` improved in place with the `transaction_status` guard + cleanup), pycodestyle/eslint clean, and added a 9.16 Bug fixes release note.

Two optional, non-blocking follow-ups for the future (not gating this merge):
- The 30s client-side node-load timeout doubles as a connection-loss signal, so a legitimately slow load on a high-latency server could surface the reconnect dialog on a live connection — a higher ceiling or a `ping()` confirmation before declaring loss would tighten that.
- `children()` now does a `SELECT 1` round-trip per node expansion; negligible locally but adds latency on remote servers. Reasonable trade-off for reliable detection.

PR #9812, opened 2026-04-02
[#9646] securityContext initContainers

### LINKED ISSUE #9646: pgadmin4-helm initContainers with hardcoded securityContext does not work on OpenShift
**Describe the bug**

Using the `pgadmin4-helm` Helm Chart does not work on OpenShift because the `initContainer`s uses a hard coded securityContext:

- https://github.com/pgadmin-org/pgadmin4/blob/master/pkg/helm/templates/deployment.yaml#L195
- https://github.com/pgadmin-org/pgadmin4/blob/master/pkg/helm/templates/deployment.yaml#L229

**To Reproduce**

helm upgrade --install pgadmin4 oci://docker.io/dpage/pgadmin4-helm

**Expected behavior**

IMO the initContainers should get the same `securityContext` as the main container ( https://github.com/pgadmin-org/pgadmin4/blob/master/pkg/helm/templates/deployment.yaml#L101 )

**Error message**

```
pods "pgadmin4-7b84bfbf6-" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, provider restricted-v2: .initContainers[0].runAsUser: Invalid value: 1001: must be in the ranges: [1003310000, 1003319999], provider restricted-v2: .initContainers[0].seLinuxOptions.level: Invalid value: "": must be s0:c58,c2, provider restricted-v2: .initContainers[1].runAsUser: Invalid value: 1001: 
```

dzabel:
- fixes #9646
- use renderSecurityContext for initContainers

## Summary by CodeRabbit

* **New Features**
  * Security context settings for container initialization are now configurable via deployment values, allowing customization of SELinux, AppArmor, user/group IDs, and privilege settings instead of using hardcoded defaults.

### DIFF
--- a/pkg/helm/templates/deployment.yaml
+++ b/pkg/helm/templates/deployment.yaml
@@ -192,24 +192,9 @@ spec:
             limits:
               cpu: 50m
               memory: 64Mi
-          securityContext:
-            seLinuxOptions: {}
-            runAsUser: 1001
-            runAsGroup: 1001
-            runAsNonRoot: true
-            privileged: false
-            readOnlyRootFilesystem: true
-            allowPrivilegeEscalation: false
-            capabilities:
-              drop: ["ALL"]
-            seccompProfile:
-              type: RuntimeDefault
-            {{- if .Values.global.compatibility.appArmor.enabled }}
-            appArmorProfile:
-              type: RuntimeDefault
-            {{- end }}
-            windowsOptions:
-              hostProcess: false
+          {{- if .Values.containerSecurityContext.enabled }}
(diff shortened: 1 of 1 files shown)


dpage:
Merged to `master` as 06300e76a (squashed, with you as author). Thanks @dzabel!

Verified with `helm template` that this is behaviour-preserving by default: the rendered init-container `securityContext` is identical to the previous hardcoded block for default values, still honours `global.compatibility.appArmor.enabled` (no `appArmorProfile` when false, present when true), and is omitted entirely when `containerSecurityContext.enabled=false`. `helm lint` clean. Added a 9.16 release note.

PR #9828, opened 2026-04-07
fix(openai): Tool name or arguments might be empty/null

tcx4c70:
Some providers might return empty/null values in streaming delta (I guess it's to keep schema consistent?). For example, streaming data for a request might be:

data: { ... "choices":[{"index":0,"delta":{"role":"assistant","tool_calls":[{"index":0, ... "function":{"name":"get_database_schema","arguments":""}}]}}] ... }
data: { ... "choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"name":null,"arguments":"{}"}}]}}] ... }
data: { ... "choices":[{"index":0,"finish_reason":"tool_calls","delta":{}}] ... }
data: [DONE]

Then pgAdmin will combine the delta to a tool call with name "null" and arguments "{}", which leads to a error that can't find the tool.

Though I don't find the official doc that mentions we should skip null value in delta, the official openai python sdk skips null values [1].

This commit is to skip null values in streaming delta to avoid empty/null tool name or arguments.

[1] https://github.com/openai/openai-python/blob/58184ad545ee2abd98e171ee09766f259d7f38cd/src/openai/lib/streaming/_deltas.py#L6

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where empty values were incorrectly overwriting valid tool call information when streaming responses from OpenAI, improving the reliability of tool call handling.

### DIFF
--- a/web/pgadmin/llm/providers/openai.py
+++ b/web/pgadmin/llm/providers/openai.py
@@ -783,9 +783,9 @@ class OpenAIClient(LLMClient):
                 if 'id' in tc_delta:
                     tc['id'] = tc_delta['id']
                 func = tc_delta.get('function', {})
-                if 'name' in func:
+                if 'name' in func and func['name']:
                     tc['name'] = func['name']
-                if 'arguments' in func:
+                if 'arguments' in func and func['arguments']:
                     tc['arguments'] += func['arguments']
 
         # Build final response


dpage:
Merged to `master` as 0223e8a02 (squashed, with you as author). Thanks @tcx4c70 — nice catch, and the OpenAI SDK reference was helpful.

Folded in two related fixes while merging, since they're the same root cause (providers emitting null fields in continuation deltas):
- Applied the same falsy-guard to the tool-call `id` — a null id in a later delta was clobbering the real id captured in the first delta, which the final build then replaced with a random uuid instead of the provider's id.
- Guarded against a null `function` object in a delta, which would otherwise raise `TypeError`.

Also added a unit test (`web/pgadmin/llm/tests/test_openai_stream.py`) covering the null-continuation, multi-chunk-arguments, and null-function cases, plus a 9.16 release note.

PR #9832, opened 2026-04-08
Fix translation for 'Refresh' in Czech

MichalBartos:
"Obnovit" is more IT relevant than "Občerstvit" ("Občerstvit" is more food relevant).

## Summary by CodeRabbit

* **Chores**
  * Updated Czech language translations to improve accuracy and user experience.


liborm85:
The word "Občerstvit" is used for "Refresh" because the next item in the context menu is "Restore", which can only be translated as "Obnovit".

This change would cause the context menu to contain two items with the same text but different actions.

I couldn't find a better Czech word to translate this - maybe only "Znovu načíst" (Reload), but that's not ideal either.

PR #9906, opened 2026-05-04
Updated Italian translation for v9.15

sbraaa:


asheshv:
### Code review

Translation file is syntactically valid (`msgfmt --check` passes; no fuzzy or untranslated entries; `.mo` semantically matches `.po`). Found 1 functional issue:

1. **Literal JSON keys/values translated in API validation error messages.** The strings extracted from `roles/__init__.py:280, 306, 385` are returned to API clients as validation error messages telling them the expected request format. The Italian translation has translated the literal JSON keys and boolean values (e.g., `role` → `ruolo`, `admin` → `amministratore`, `inherit` → `eredita`, `set` → `imposta`, `True/False` → `Vero/Falso`, `'added'/'deleted'/'updated'` → `'aggiunto'/'eliminato'/'aggiornato'`). An Italian-speaking user following this error message will construct a payload that the API rejects, since the underlying validators in `_validate_input_dict_for_new` still match on the English keys (`['role', 'admin', 'inherit', 'set']`).

   https://github.com/pgadmin-org/pgadmin4/blob/febceb0a42b3e08a88f2baa76f465caaf6e86614/web/pgadmin/translations/it/LC_MESSAGES/messages.po#L9241-L9277

   Source for context (server-side validation, English keys): https://github.com/pgadmin-org/pgadmin4/blob/febceb0a42b3e08a88f2baa76f465caaf6e86614/web/pgadmin/browser/server_groups/servers/roles/__init__.py#L294-L302

   Suggestion: keep the JSON keys, the `True/False` literals, and the `'added'/'deleted'/'updated'` keys verbatim (only translate the surrounding prose). Compare with the existing seclabels block right below, where `provider`/`label` were correctly left in English.

Minor (not blocking):

- Original indentation inside the JSON examples was dropped, slightly hurting readability of the rendered error.
- `msgstr "${error.response?.data?.errormsg || errore?.message}"` (line 10493) translates the JS identifier `error` to `errore`. In practice this entry is never matched at runtime because the source uses a backtick template literal that is evaluated before `gettext()` is called (https://github.com/pgadmin-org/pgadmin4/blob/febceb0a42b3e08a88f2baa76f465caaf6e86614/web/pgadmin/browser/static/js/browser.js#L317-L321), so it's harmless dead code, but the entry should remain identical to the msgid.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

asheshv:
I am merging this PR.
Please send an updated PR with these fixes.

Issue #9701, opened 2026-03-06
using with docker-compose


MBaechtoldUZH:
I created a docker-compose.yml

version: "3.9"

services:
  postgres:
    image: postgres:17
    container_name: pg17
    restart: always
    environment:
      POSTGRES_DB: pgwebadmin
      POSTGRES_USER: pgwebadmin4
      POSTGRES_PASSWORD: xyz   # your password
    ports:
      - "5432:5432"
    volumes:
      - pgdata:/var/lib/postgresql/data

  pgadmin:
    image: dpage/pgadmin4:latest
    container_name: pgadmin
    restart: always
    depends_on:
      - postgres
    environment:
      PGADMIN_DEFAULT_EMAIL: admin@example.com
      PGADMIN_DEFAULT_PASSWORD: abc123
      PGADMIN_CONFIG_CONFIG_DATABASE_URI: "'postgresql://pgwebadmin4:xyz@postgres:5432/pgwebadmin'"
    ports:
      - "5050:80"
    volumes:
      - pgadmin-data:/var/lib/pgadmin

volumes:
  pgdata:
  pgadmin-data:

when I check the logs:

docker logs pgadmin -f
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    from check_external_config_db import check_external_config_db; val = check_external_config_db('postgresql://pgwebadmin4:xyz@postgres:5432/pgwebadmin'); print(val)
                                                                         ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pgadmin4/pgadmin/utils/check_external_config_db.py", line 18, in check_external_config_db
    engine = create_engine(database_uri)
  File "<string>", line 2, in create_engine
  File "/venv/lib/python3.14/site-packages/sqlalchemy/util/deprecations.py", line 281, in warned
    return fn(*args, **kwargs)  # type: ignore[no-any-return]
  File "/venv/lib/python3.14/site-packages/sqlalchemy/engine/create.py", line 617, in create_engine
    dbapi = dbapi_meth(**dbapi_args)
  File "/venv/lib/python3.14/site-packages/sqlalchemy/dialects/postgresql/psycopg2.py", line 696, in import_dbapi
    import psycopg2
ModuleNotFoundError: No module named 'psycopg2'
postfix/postlog: starting the Postfix mail system
[...]

PR #9914, opened 2026-05-05
Ensure to use psycopg3 is used while connecting to postgres DB when PGADMIN_CONFIG_CONFIG_DATABASE_URI is specified. #9701

yogeshmahajan-1903:
## Summary by CodeRabbit

* **Bug Fixes**
  * Normalizes PostgreSQL-style database URIs so the app consistently uses the correct driver prefix, improving reliability when connecting to external configuration databases and avoiding connection setup errors.
* **Tests**
  * Added unit tests to validate URI normalization behavior and ensure idempotent, safe handling of non-Postgres URIs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pgadmin-org/pgadmin4/pull/9914?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


asheshv:
### Code review

Found 2 issues:

1. **Fix is incomplete — same rewrite is missing where `SQLALCHEMY_DATABASE_URI` is actually set, so the bug from issue #9701 will still reproduce.** `check_external_config_db()` is invoked once from `pkg/docker/entrypoint.sh:153` as a pre-flight check. After that pre-check passes, `pgadmin/__init__.py:340` assigns the unmodified `config.CONFIG_DATABASE_URI` directly to `app.config['SQLALCHEMY_DATABASE_URI']`, and `run_migration_for_others()` → `db_upgrade(app)` → SQLAlchemy then tries to load the psycopg2 dialect for the bare `postgresql://` URI. This is exactly the second traceback the reporter pasted in #9701 (`File "/pgadmin4/pgadmin/__init__.py", line 476` → `import psycopg2 → ModuleNotFoundError`). The same `postgresql://` → `postgresql+psycopg://` normalization needs to be applied at the URI-assignment site (or in `config.py`/`__init__.py`) for the fix to actually work.

   https://github.com/pgadmin-org/pgadmin4/blob/8d923618990099ae6f7f3e365f6b572c92224e57/web/pgadmin/__init__.py#L337-L343

2. **`postgres://` branch produces an invalid SQLAlchemy URL.** Verified against the project's SQLAlchemy 2.0.49: `postgres://` and `postgres+psycopg://` both raise `NoSuchModuleError: Can't load plugin: sqlalchemy.dialects:postgres[.psycopg]` (the `postgres://` alias was removed in SQLAlchemy 1.4). For input `postgres://...`, the current code produces `postgres+psycopg://...`, which still fails at `create_engine`. To handle the `postgres://` form, the rewrite needs to swap the scheme to `postgresql+psycopg://`, e.g.:

   ```python
   if database_uri.startswith(("postgresql://", "postgres://")):
       database_uri = re.sub(r"^postgres(ql)?://", "postgresql+psycopg://", database_uri, count=1)
   ```

   https://github.com/pgadmin-org/pgadmin4/blob/8d923618990099ae6f7f3e365f6b572c92224e57/web/pgadmin/utils/check_external_config_db.py#L17-L21

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

yogeshmahajan-1903:
> ### Code review
> Found 2 issues:
> 
> 1. **Fix is incomplete — same rewrite is missing where `SQLALCHEMY_DATABASE_URI` is actually set, so the bug from issue [using with docker-compose #9701](https://github.com/pgadmin-org/pgadmin4/issues/9701) will still reproduce.** `check_external_config_db()` is invoked once from `pkg/docker/entrypoint.sh:153` as a pre-flight check. After that pre-check passes, `pgadmin/__init__.py:340` assigns the unmodified `config.CONFIG_DATABASE_URI` directly to `app.config['SQLALCHEMY_DATABASE_URI']`, and `run_migration_for_others()` → `db_upgrade(app)` → SQLAlchemy then tries to load the psycopg2 dialect for the bare `postgresql://` URI. This is exactly the second traceback the reporter pasted in [using with docker-compose #9701](https://github.com/pgadmin-org/pgadmin4/issues/9701) (`File "/pgadmin4/pgadmin/__init__.py", line 476` → `import psycopg2 → ModuleNotFoundError`). The same `postgresql://` → `postgresql+psycopg://` normalization needs to be applied at the URI-assignment site (or in `config.py`/`__init__.py`) for the fix to actually work.
>    https://github.com/pgadmin-org/pgadmin4/blob/8d923618990099ae6f7f3e365f6b572c92224e57/web/pgadmin/__init__.py#L337-L343
> 2. **`postgres://` branch produces an invalid SQLAlchemy URL.** Verified against the project's SQLAlchemy 2.0.49: `postgres://` and `postgres+psycopg://` both raise `NoSuchModuleError: Can't load plugin: sqlalchemy.dialects:postgres[.psycopg]` (the `postgres://` alias was removed in SQLAlchemy 1.4). For input `postgres://...`, the current code produces `postgres+psycopg://...`, which still fails at `create_engine`. To handle the `postgres://` form, the rewrite needs to swap the scheme to `postgresql+psycopg://`, e.g.:
>    ```python
>    if database_uri.startswith(("postgresql://", "postgres://")):
>        database_uri = re.sub(r"^postgres(ql)?://", "postgresql+psycopg://", database_uri, count=1)
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    https://github.com/pgadmin-org/pgadmin4/blob/8d923618990099ae6f7f3e365f6b572c92224e57/web/pgadmin/utils/check_external_config_db.py#L17-L21
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Fixed.

yanehi:
I encountered this error while upgrading pgAdmin from version `9.14` to `9.15`.

asheshv (commented):
Thanks for the fix — the core change is correct. A couple of follow-ups before merge:

**1. Extract the rewrite into a shared helper.**
The same `re.sub(r"^postgres(ql)?://", "postgresql+psycopg://", ...)` lives in both `web/pgadmin/__init__.py` and `web/pgadmin/utils/check_external_config_db.py`. A single `normalize_database_uri(uri)` helper (e.g. under `web/pgadmin/utils/`) used by both call sites prevents drift if another driver (asyncpg, cockroachdb, …) ever needs similar treatment, and removes the asymmetric `startswith` guard that currently exists in only one of the two sites.

**2. Add a unit test for the rewrite.**
Small but valuable for regression prevention. Suggested cases:
- `postgres://u:p@h/db` → `postgresql+psycopg://u:p@h/db`
- `postgresql://u:p@h/db` → `postgresql+psycopg://u:p@h/db`
- `postgresql+psycopg://…` → unchanged (idempotency)
- `sqlite:///…` → unchanged

**Nits (optional):**
- `r"^postgres(ql)?://"` → `r"^postgres(?:ql)?://"` (non-capturing group; cosmetic).
- `_db_uri` in `web/pgadmin/__init__.py` is single-use — the `re.sub` result can be assigned to `SQLALCHEMY_DATABASE_URI` directly.
- A release-notes entry referencing #9701 would be nice.

Otherwise LGTM — the regex is anchored, `count=1`, idempotent for already-prefixed URIs, and both URI-writing paths (`create_app` and the Docker pre-flight via `check_external_config_db`) are covered.

yanehi:
@asheshv @yogeshmahajan-1903 - After updating from `9.14` to `9.15`, I get the following error and PgAdmin won't start:

```bash

...
File "/pgadmin4/pgadmin/utils/check_external_config_db.py", line 18, in check_external_config_db
raise exc.ArgumentError(
"Could not parse SQLAlchemy URL from given URL string"
)
...
```

Unfortunately, the suggestion in https://github.com/pgadmin-org/pgadmin4/issues/9701#issuecomment-4376678334 to adjust the connection string didn't work for me.

yogeshmahajan-1903:
> [#9701 (comment)](https://github.com/pgadmin-org/pgadmin4/issues/9701#issuecomment-4376678334)

Could you please provide an example for your DB_URI?

asheshv:
@yanehi — your error is a different bug from what this PR fixes. It's the Docker entrypoint mangling `PGADMIN_CONFIG_CONFIG_DATABASE_URI` with an extra set of quotes, causing SQLAlchemy to see literal quote characters inside the URI string. Broken in shipped images since Oct 2024 (commit `1fe840fca`), and the `#9701` workaround can't help because the shell mangles whatever you put there.

#10009 fixes it. Until that lands: mount a `config_local.py` containing `CONFIG_DATABASE_URI = 'postgresql+psycopg://…'` at `/pgadmin4/config_local.py` and unset the env var — bypasses the entrypoint path entirely.

web/pgadmin/utils/check_external_config_db.py
```diff
@@ -15,6 +17,11 @@ def check_external_config_db(database_uri):
     Check if external config database exists if it
     is being used.
     """
+    if database_uri.startswith(("postgresql://", "postgres://")):
+        database_uri = re.sub(
+            r"^postgres(ql)?://", "postgresql+psycopg://",
+            database_uri, count=1
+        )
```
asheshv: Use the util function here as well.

asheshv (changes requested):

web/pgadmin/utils/db_utils.py
```diff
@@ -0,0 +1,20 @@
...
+import re
+
+
+def normalize_database_uri(uri):
+    """
+    Rewrite a bare postgres(ql):// URI to use the psycopg3 driver scheme
+    postgresql+psycopg://.  URIs that already carry a driver specifier
+    (e.g. postgresql+psycopg://) are returned unchanged.
+    """
+    return re.sub(r"^postgres(?:ql)?://",
```
mzabuawala, line 19: (Nitpicking)
In practice, connection URIs are almost always lowercase.
However, because users can set `CONFIG_DATABASE_URI` via the config file, unconventional variants such as PostgreSQL://... would not be normalised as required.

asheshv: It's corner case, and not really a standard way to do it,
Hence - I would ignore that.

PR #9961, opened 2026-05-20
chore(deps): bump @mui/material 7 -> 9 and @mui/x-date-pickers 8 -> 9

asheshv:
> **Draft** — automated checks pass cleanly (824/824 jest, eslint silent), but this PR carries visual-regression risk that needs a manual UI sweep before marking ready.

## Summary

Supersedes dependabot #9843 and #9888, which must land together (date-pickers v9 requires @mui/material v9 as a peer). @mui/material skipped v8 on npm (7.x → 9.x), so only v9's breaking changes apply.

## Manifest changes

| Package | Was | Now (resolved) |
|---|---|---|
| `@mui/material` | `^7.3.10` | `^9.0.0` (9.0.1) |
| `@mui/icons-material` | `^7.3.10` | `^9.0.0` (9.0.1) |
| `@mui/x-date-pickers` | `^8.28.3` | `^9.0.2` (9.2.0) |

### 1. Two legacy icon imports renamed

v9 removed legacy `*Outline` icons in favor of the `*Outlined` variants. Both in `web/pgadmin/static/js/Explain/AIInsights.jsx`:

```diff
-import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline';
+import ErrorOutlineIcon from '@mui/icons-material/ErrorOutlined';
-import CheckCircleOutlineIcon from '@mui/icons-material/CheckCircleOutline';
+import CheckCircleOutlineIcon from '@mui/icons-material/CheckCircleOutlined';
```

### 2. `styled(Box)` → `styled('div')` across 50 files

In v9, `Box` stopped filtering CSS-property-named keys (`flexDirection`, `marginLeft`, `alignItems`, etc.) out of the props it forwarded to the DOM. Any `styled(Box)(({theme}) => ({ flexDirection: 'column', ... }))` leaked those CSS keys to the underlying `<div>` as React props, surfacing as a "React does not recognize the `flexDirection` prop on a DOM element" warning. Using `styled('div')` keeps the CSS-in-JS layer working unchanged and removes the now-redundant Box layer — every affected styled wrapper was a pure styling container, not using Box's sx-passthrough or `component=` swap behavior.

### 3. Box system props → `sx` (official MUI codemod)

v9 removed individual system-prop syntax from `Box`, `Stack`, `Typography`, `Link`, `Grid`, `DialogContentText`, `TimelineContent`, and `TimelineOppositeContent`. Applied via:

```sh
npx @mui/codemod@latest v9.0.0/system-props pgadmin
```

65 files transformed; 6 Jinja-templated `.js` files in templates/ skipped (parse errors, not real JS, no MUI usage).

Example transform:

```diff
-<Box flexGrow={1} className='ProcessDetails-logs'>
+<Box sx={{flexGrow: 1}} className='ProcessDetails-logs'>
```

### 4. One test updated for class-name split

MUI v9 Button no longer emits the combined `MuiButton-containedPrimary` class. It now emits `MuiButton-contained` and `MuiButton-colorPrimary` as separate classes. Updated `web/regression/javascript/components/Buttons.spec.js` accordingly.

## Test plan

- [x] `yarn install` resolves cleanly
- [x] `yarn run test:js-once` → 140/0/0 suites, 824/0/0 tests
- [x] `yarn run linter` (web) → clean (silent)
- [ ] CI matrix (Linux/Mac/Windows) passes
- [ ] **Manual visual sweep** in `yarn start` of: login screen, browser tree, query tool, ERD, schema dialogs (table/function/trigger/index), theme switcher (light/dark/high-contrast), file manager, debugger panel, backup/restore dialogs.

## Notes for reviewers

The major risk surface is **visual regression**, not API breakage. v9's "sx prop performance" and "deprecated API cleanup" work shifts a few pixels in places — most notably `ListItemIcon` spacing (changed from hardcoded `minWidth` to theme spacing). Worth a careful look at menus and toolbars.

The other Box-related changes (system props → sx) preserve behavior 1:1 — they're a pure syntactic migration.

`@mui/x-date-pickers`'s legacy `LocalizationProvider` API was removed in v9, but we already use the modern `dateAdapter={AdapterDateFns}` form in `Theme/index.jsx:959`, so no change needed there.

## Supersedes

- Dependabot #9843 (will close once this lands)
- Dependabot #9888 (will close once this lands)


asheshv:
Picking this up since it's been sitting since May. The analysis in the description holds up — I verified the parts I could:

- MUI really does skip v8 on npm (`7.3.11` → `9.0.0-alpha.0`); `latest` is now **9.4.0**.
- The icon removals are real. I unpacked `@mui/icons-material@9.0.1`: `ErrorOutline.js` and `CheckCircleOutline.js` are gone, `ErrorOutlined.js`/`CheckCircleOutlined.js` are present. The rename is required and correct.
- `LocalizationProvider` at `static/js/Theme/index.jsx:17` already uses the modern `dateAdapter={AdapterDateFns}` form — no change needed, as stated.
- Better than expected on migration surface: **zero** `@mui/styles`/`makeStyles`/`withStyles`, and **zero** Grid v1 `<Grid item>` uses, so the Grid v1→v2 migration everyone dreads doesn't apply to us at all. Only 3 files import `@mui/x-date-pickers`. The `TimelineContent` mention in the body is codemod boilerplate — we have no `@mui/lab`.
- Measured scope: 154 files import `@mui/`, 98 touched by this PR, 68 `styled(Box)` → `styled('div')` conversions.

Right call keeping it draft. Three things stand between it and merge:

**1. The codemod rewrote a vendored third-party file (`static/vendor/require/require.min.js:9`).**
It ran over the whole `pgadmin/` tree, including `static/vendor/`, and reformatted the single minified line of a RequireJS 2.3.6 bundle. An unreviewable diff to a vendored artifact doesn't belong in a dependency bump, and it destroys our ability to diff that file against upstream later. `git checkout origin/master -- web/pgadmin/static/vendor/`, and re-scope the codemod to exclude `static/vendor/` and `static/js/generated/`.

**2. More codemod collateral (`SchemaView/common.js:19-33`, `browser/static/js/keyboard.js`).**
Both are cosmetic reformats of files with no MUI usage. `common.js` is logically equivalent (`A || B && C && D` vs the parenthesised form), but the reflow moved the `// serverInfo found and it's within range` comment onto the wrong expression — in a function whose operator precedence is already easy to misread. Both should be reverted.

**3. It won't merge any more (`web/package.json:83-85`).**
The branch bumps from `^7.3.10`/`^8.28.3`; master is now `^7.3.11`/`^8.29.3`. With 98 touched files and a regenerated `yarn.lock`, that's a guaranteed conflict. And `^9.0.0` resolved to 9.0.1 back in May — 9.4.0 is out now, four minors of v9 fixes, including the settling of the exact `sx`/`Box` behaviour this PR migrates around.

On the production-bundle trap: credit where due, this one actually cleared it — the 5 `run-feature-tests-pg` jobs build with `yarn run bundle` (not `bundle-dev`) and they passed at 9.0.1, so the failure mode that's bitten us three times was genuinely exercised. But that evidence doesn't survive a rebase: re-resolving `^9.0.0` to 9.4.0 with a fresh lockfile is a different dependency graph, so the prod-bundle result has to be re-earned.

**`styled(Box)` → `styled('div')` will decay unless it's fenced.** The 68 conversions are correct, but master has already grown new `styled(Box)` wrappers since this branch was cut, and every one added after merge silently reintroduces the "React does not recognize the `flexDirection` prop" warning. An ESLint `no-restricted-syntax` rule banning `styled(Box)` should land in the same PR.

**Suggestion:** rather than salvaging this branch as one lump, split it. Land the 68 `styled(Box)` → `styled('div')` conversions plus the ESLint rule as a **separate prep PR that works fine on v7** — it's a mechanical, low-risk change and it merges today. That shrinks the actual MUI 7→9 bump to a handful of files (icon renames plus whatever the visual sweep turns up), which sidesteps the 98-file conflict window entirely. Then: rebase, re-run the codemod with `static/vendor/` and `static/js/generated/` excluded, target `^9.4.0`, and do the manual visual sweep — still unticked in the test plan, and with no visual-regression harness in the repo it's the real blocker (the `ListItemIcon` spacing shift is already flagged there).

PR #9971, opened 2026-05-22
Fix issue with Username being required for Shared server import config

### LINKED ISSUE #9226: Username is required even though SharedUsername is set
**Describe the bug**

Not sure if this is by design but when using `SharedUsername` in server definition the `Username` field is still required. If only `SharedUsername` is provided pgAdmin shows an error and server list is not populated.

**To Reproduce**

1. Use the following servers.json definition file:
   ```
   {
     "Servers": {
       "mysharedserver": {
         "Name": "My Shared Server",
         "Group": "Shared Servers",
         "Host": "postgresql",
         "Port": 5432,
         "MaintenanceDB": "postgres",
         "Shared": true,
         "SharedUsername": "postgres"
       }
     }
   }
   ```
2. Set `PGADMIN_SERVER_JSON_FILE` environment variable to point to the servers.json definition file.
3. Start pgAdmin and observe the server list. It will be empty.
4. Now add `"Username": "postgres",` into the list.
5. Restart pgAdmin. Server list is now correctly populated.

**Expected behavior**

If Shared Username is used there should be no need to duplicate it into two parameters.

**Error message**
```
'Username' attribute not found for server 'mysharedserver'
```

**Desktop**
 - OS: Linux
 - pgAdmin version: 9.8
 - Mode: Server
 - Browser (if running in server mode): Edge
 - Package type: Container

KijongHan:
Resolves https://github.com/pgadmin-org/pgadmin4/issues/9226

# What does this PR do?
This PR resolves `Username attribute not found for server 'shared_server'` errors where server json import validation logic had a hard constraint on `Username` being required even when `Shared` was set to true. This PR introduces more coherent constraint validation logic for shared servers so that you are required to either provide `SharedUsername` or `Username` (for backwards compatibility). Furthermore, the underlying data model 'Username' property is set to 'SharedUsername' (for shared servers only) to align with UI behaviour for setting this fallback.

The UX also feels better as when after registering shared servers with only the `SharedUsername` set, when trying to connect to the server for the first time it prompts for the password (instead of showing an infinite loading spinner because the underlying `Username` is not set)

# Summary of changes
add separate `Username` constraint validation logic for shared servers when importing `servers.json`
add fallback to set `Username` to `SharedUsername` for shared servers so that the UX flow matches the manual UI creation flow for registering new shared servers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved import and validation of shared database servers: shared servers are now skipped for non-admin users, username requirements are enforced conditionally (allowing a shared-specific username when appropriate), and the effective credential used for imported servers is computed reliably. This yields more accurate configuration, fewer import errors, and improved security when handling shared server entries.

### DIFF
--- a/web/pgadmin/utils/__init__.py
+++ b/web/pgadmin/utils/__init__.py
@@ -735,8 +735,9 @@ def load_database_servers(input_file, selected_servers,
 
             is_shared = obj.get("Shared", None)
             username = obj.get("Username", None)
+            shared_username = obj.get("SharedUsername", None)
             if is_shared and not username:
-                username = obj.get("SharedUsername", None)
+                username = shared_username
 
             # Create the server
             new_server = Server()
@@ -805,7 +806,7 @@ def load_database_servers(input_file, selected_servers,
 
             new_server.shared = is_shared
 
-            new_server.shared_username = obj.get("SharedUsername", None)
+            new_server.shared_username = shared_username
 
             new_server.kerberos_conn = obj.get("KerberosAuthentication", None)
 
--- a/web/pgadmin/utils/__init__.py
+++ b/web/pgadmin/utils/__init__.py
@@ -733,6 +733,11 @@ def load_database_servers(input_file, selected_servers,
                 groups_added = groups_added + 1
                 groups = ServerGroup.query.filter_by(user_id=user_id)
 
+            is_shared = obj.get("Shared", None)
+            username = obj.get("Username", None)
+            if is_shared and not username:
+                username = obj.get("SharedUsername", None)
+
             # Create the server
             new_server = Server()
             new_server.name = obj["Name"]
@@ -744,7 +749,7 @@ def load_database_servers(input_file, selected_servers,
 
             new_server.port = obj.get("Port", None)
 
-            new_server.username = obj.get("Username", None)
+            new_server.username = username
 
             new_server.role = obj.get("Role", None)
 
@@ -798,7 +803,7 @@ def load_database_servers(input_file, selected_servers,
             new_server.tunnel_keep_alive = \
                 obj.get("TunnelKeepAlive", None)
 
-            new_server.shared = obj.get("Shared", None)
+            new_server.shared = is_shared
 
             new_server.shared_username = obj.get("SharedUsername", None)
 
--- a/web/pgadmin/utils/__init__.py
+++ b/web/pgadmin/utils/__init__.py
@@ -736,7 +736,7 @@ def load_database_servers(input_file, selected_servers,
             is_shared = obj.get("Shared", None)
             username = obj.get("Username", None)
             shared_username = obj.get("SharedUsername", None)
-            if is_shared and not username:
+            if is_shared and username is None:
                 username = shared_username
 
             # Create the server
--- a/web/pgadmin/utils/__init__.py
+++ b/web/pgadmin/utils/__init__.py
@@ -601,8 +601,10 @@ def validate_json_data(data, is_admin):
     for server in data["Servers"]:
         obj = data["Servers"][server]
 
+        is_shared = obj.get("Shared", None)
+
         # Check if server is shared.Won't import if user is non-admin
-        if obj.get('Shared', None) and not is_admin:
+        if is_shared and not is_admin:
             print("Won't import the server '%s' as it is shared " %
                   obj["Name"])
             skip_servers.append(server)
@@ -627,14 +629,25 @@ def validate_json_data(data, is_admin):
         is_service_attrib_available = obj.get("Service", None) is not None
 
         if not is_service_attrib_available:
-            for attrib in ("Port", "Username"):
-                errmsg = check_attrib(attrib)
+            errmsg = check_attrib("Port")
+            if errmsg:
+                return errmsg
+            errmsg = check_is_integer(obj["Port"])
+            if errmsg:
+                return errmsg
+
+            if is_shared:
+                # Shared servers may carry either the owner's username
+                # or a per-user override, so accept either attribute.
+                if "Username" not in obj and "SharedUsername" not in obj:
+                    return gettext(
+                        "'Username' or 'SharedUsername' attribute not "
+                        "found for server '%s'" % server
+                    )
+            else:
+                errmsg = check_attrib("Username")
                 if errmsg:
                     return errmsg
-                if attrib == 'Port':
-                    errmsg = check_is_integer(obj[attrib])
-                    if errmsg:
-                        return errmsg
 
         errmsg = check_attrib("MaintenanceDB")
         if errmsg:


web/pgadmin/utils/__init__.py
```diff
@@ -720,6 +733,11 @@ def load_database_servers(input_file, selected_servers,
                 groups_added = groups_added + 1
                 groups = ServerGroup.query.filter_by(user_id=user_id)
 
+            is_shared = obj.get("Shared", None)
+            username = obj.get("Username", None)
+            if is_shared and not username:
```
hiteshjambhale: Hi @KijongHan , In case where `username` is a empty string in json data, it will fall back to `SharedUsername`. If SharedUsername is also missing, the server imports silently with username = None. We can use `username is None` to only trigger the fallback when the key is genuinely absent. 

I was able to reproduce this with following json
```
{
  "Servers": {
    "1": {
      "Group": "Servers",
      "Name": "Test",
      "Shared": true,
      "Username": "",
      "Host": "127.0.0.1",
      "Port": 5432,
      "MaintenanceDB": "postgres"
    }
  }
}
```

hiteshjambhale (changes requested):

web/pgadmin/utils/__init__.py
KijongHan: Hey @hiteshjambhale thanks alot for checking this use case! I share your concern around an empty username string being considered a valid import, but I'm thinking that we should short-circuit the import process further up the validation chain (during JSON validation) [here](https://github.com/pgadmin-org/pgadmin4/pull/9971/changes#diff-04db232fe61a1e28a3e6d6a22f7f7dae23a7ce942f07a4bdd89fa05aef141d00L611) that way we handle the validation in the right place

From what I can see, an empty string `username` being considered valid has been there for some time, so I want to check with you @asheshv or @akshay-joshi and get your opinion on whether we should tighten the validation logic on the `username` property and not allow empty strings

web/pgadmin/utils/__init__.py
```diff
@@ -720,6 +733,11 @@ def load_database_servers(input_file, selected_servers,
...
+                username = obj.get("SharedUsername", None)
+
```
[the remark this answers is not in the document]
KijongHan: > the import currently prefers Username over SharedUsername when both are present

I think this `Username` override makes sense, so that `SharedUsername` is only the default value (this is how it is currently treated in the UI creation stage as well at the moment)

> and it does not populate shared_username when only legacy Username is provided

I dont think we should be treating `shared_username` property value like a fallback. It should be "pure". This feels aligned to how `username` is used as a network login (so fallback seems reasonable) but `shared_username` is treated as the "default username for these shared servers" so it doesn't make sense to me to introduce a fallback here. Keen for your thoughts @asheshv

web/pgadmin/utils/__init__.py
```diff
@@ -785,7 +803,7 @@ def load_database_servers(input_file, selected_servers,
             new_server.tunnel_keep_alive = \
                 obj.get("TunnelKeepAlive", None)
 
-            new_server.shared = obj.get("Shared", None)
+            new_server.shared = is_shared
 
             new_server.shared_username = obj.get("SharedUsername", None)
```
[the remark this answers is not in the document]
KijongHan: added this change for now, even though the decision on normalized `shared_username` is still pending because it cleans up the `shared_username` property assignment (and no duplicate `obj.get("SharedUsername", None)`)

web/pgadmin/utils/__init__.py
KijongHan: Hey @hiteshjambhale what do you think about raising the username empty string validation logic as a separate issue?

asheshv:
Thanks for the careful analysis @KijongHan, and for not just rolling over to Copilot — your read on `Username` vs `SharedUsername` precedence matches the UI flow and is the right call. `Username` is the per-user override, `SharedUsername` is the group default, and `Username` winning when both are present is the correct semantics.

One ask before this lands: please change the fallback at line 739 from

```python
if is_shared and not username:
    username = shared_username
```

to

```python
if is_shared and username is None:
    username = shared_username
```

Reason: @hiteshjambhale's reproducer (`Username: ""` with `SharedUsername` absent) gets `username = None` after the fallback, and `Server.username` is `nullable=False` in the model. The current non-shared branch has the same latent issue with empty strings, but it's pre-existing — this PR widens the path on the shared branch, which the `is None` check closes without changing any other behaviour.

The broader question — *should `{"Username": ""}` be valid input at all, given PostgreSQL won't authenticate with an empty username?* — is worth a separate issue (and I'd argue the answer is no, validation should reject it for both shared and non-shared, parallel to the `Service` escape hatch). Happy to file that follow-up so this PR can stay scoped.

On the `shared_username` "purity" point: I agree with you — `shared_username` is the group default, it shouldn't silently inherit from a legacy `Username` field. Your handling there is correct.

KijongHan:
thanks so much for your reviews @asheshv @hiteshjambhale and appreciate the followup to some of the questions I've posed. Have addressed your suggested changes.

KijongHan:
hi @hiteshjambhale @asheshv Checking to see if you had a chance to rereview my changes that addresses your PR feedback/concerns? Keen to get these changes in!

dpage:
Thanks Thomas, and apologies for the wait. This has landed on `master` as 0ebefaf80e4545d290db9bbefc444aa7a7cba7c2, squashed with you as the author; because the SHA changed GitHub will not close the PR itself, so I am doing that by hand.

I folded three things in whilst merging, all of them small:

The shared branch of the validation now tests the values rather than merely the presence of the keys, so `{"Shared": true, "SharedUsername": null}` and its empty-string equivalent are rejected instead of being accepted and then stored as the server's username. That was the same class of problem @hiteshjambhale spotted originally, just reached through the other key, and it is worth closing on the path this change widens. The pre-existing hole on the non-shared path, where `{"Username": ""}` is still accepted, is untouched and still wants the separate fix Ashesh described.

I added `web/pgadmin/utils/tests/test_validate_json_data.py`, which is the first coverage either of these functions has had: the username requirements for shared and non-shared servers, the `Service` escape hatch, the integer port check, and shared servers being skipped rather than rejected for a non-admin user.

The JSON format section of `docs/en_US/import_export_servers.rst` still said `Username` was required in every server definition, so it now notes that a shared server may supply `SharedUsername` instead.

I left `obj.get("Shared", None)` as it was rather than defaulting it to `False`, incidentally. It looks like a tidy-up, but the dialog also stores NULL for a non-shared server, and `add_value()` would then start emitting `"Shared": false` into exported files, so it would trade one inconsistency for another.

Thanks also for pushing back on the Copilot review rather than just going along with it. Your reading of `Username` versus `SharedUsername` precedence was the right one.

dpage:
For the record, the follow-up on the non-shared path is now filed as #10309, with @hiteshjambhale's original reproducer and a note on why an empty username fails so obscurely (libpq substitutes the OS account, so the error names a user that appears nowhere in the imported configuration).

PR #9973, opened 2026-05-23
Fix issue with importing CheckBoxTree component

KijongHan:
Resolves https://github.com/pgadmin-org/pgadmin4/issues/9972

# What does this PR do?
The react-checkbox-tree v2 release added "type": "module" to its package.json, which made webpack treat its UMD bundle (lib/index.js) as ESM and ignore the module.exports assignment. This means that importing the component yielded undefined and crashed the tree views. Aliasing the package to its true ESM entry (lib/index.esm.js) makes webpack read the proper export statements and restores the component.

# Summary of changes
add forced routing to underlying esm module declaration when importing from `react-checkbox-tree`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated webpack module resolution configuration for the checkbox tree component to ensure correct import handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pgadmin-org/pgadmin4/pull/9973?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #9989, opened 2026-05-29
Fix CheckBoxTree rendering by aliasing react-checkbox-tree to its ESM bundle

### LINKED ISSUE #9972: pgAdmin can't render import/export servers menu prompt (Something went wrong)

**Describe the bug**
The react-checkbox-tree v2 release added "type": "module" to its package.json, which is making webpack treat its UMD bundle (lib/index.js) as ESM and ignore the module.exports assignment. Its possibly related to the recent changes in pgadmin to update this dependency to use v2

**To Reproduce**

1. Go to 'Tools' in top bar
3. Click on 'Import servers'
4. See the error

**Expected behavior**

No errors when rendering CheckBoxTree component

**Error message**

```
Error: Element type is invalid: expected a string
CheckBoxTree.jsx: 68
(for built-in components) or a class/function (for composite components)
but got: undefined.
You likely forgot to export your component from the
file it's defined in,
or you might have mixed up default and named
imports.
```

**Screenshots**

Image

 - OS: macOS Tahoe
 - pgAdmin version: latest from main
 - Mode: Desktop
 - Browser (if running in server mode): chrome
 - Package type: Docker Container

dpage:
## Summary

- `react-checkbox-tree` v2 (bumped to `^2.0.1` in #9870) marks the package as `"type": "module"` while still pointing its `require` exports condition at a UMD bundle. Babel transforms our `import` statements to `require()`, so webpack picks the UMD file and then treats it as ESM (because of `"type": "module"`). The UMD wrapper's `module.exports = factory(...)` never runs in that context and the default export is `undefined` — which is why dialogs that mount `CheckBoxTree` (Tools → Import/Export Servers and similar) fail with *"Element type is invalid"*.
- Add a webpack `resolve.alias` for `react-checkbox-tree` pointing at the package's ESM bundle (`lib/index.esm.js`). That file is explicitly exposed by the package's own `"./*": "./*"` exports passthrough, has a real default export, and is what webpack would have picked if our build didn't down-transform imports to requires.
- The `$` suffix makes it an exact-match alias, so any future subpath imports (e.g. CSS) continue to resolve normally.

## Test plan

- [x] `yarn run linter` — clean
- [x] `yarn run jest` — all 140 suites / 824 tests pass
- [x] `yarn run bundle:dev` and `yarn run bundle` both compile cleanly (only the pre-existing asset-size advisory warnings)
- [x] Bundle inspection: before the fix, `app.bundle.js` contained `node_modules/react-checkbox-tree/lib/index.js` (the UMD wrapper); after the fix, only `lib/index.esm.js` is bundled and the `webpackUniversalModuleDefinition` wrapper is gone
- [ ] Manual smoke test: Tools → Import Servers and Tools → Export Servers render their dialogs without the *"Element type is invalid"* error

Closes #9972

## Summary by CodeRabbit

* **Chores**
  * Updated webpack configuration to resolve compatibility issues with the react-checkbox-tree dependency by ensuring the correct module bundle is loaded during the build process.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pgadmin-org/pgadmin4/pull/9989?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/webpack.shim.js
+++ b/web/webpack.shim.js
@@ -44,6 +44,13 @@ let webpackShimConfig = {
     'graphlib': path.join(__dirname, 'node_modules/graphlib'),
     'react': path.join(__dirname, 'node_modules/react'),
     'react-dom': path.join(__dirname, 'node_modules/react-dom'),
+    // react-checkbox-tree v2 marks the package as "type": "module" but ships
+    // a UMD bundle as its CommonJS entry point. Babel transforms our ESM
+    // imports to require(), so webpack picks the UMD file and then treats it
+    // as ESM (because of "type": "module") — and module.exports never runs,
+    // leaving the default export undefined. Point directly at the ESM bundle
+    // to bypass the broken main entry.
+    'react-checkbox-tree$': path.join(__dirname, 'node_modules/react-checkbox-tree/lib/index.esm.js'),
     'stylis': path.join(__dirname, 'node_modules/stylis'),
     'popper.js': path.join(__dirname, 'node_modules/popper.js'),
 


KijongHan:
Hey @dpage ! Thanks for looking into this one. It looks like we had the same idea actually, I opened a PR for this last week over in #9973. 

Could you please take a look so we can avoid duplicating the review effort? Keen to get your feedback and your help in getting it across the line

PR #9991, opened 2026-05-29
feat: include authenticated user identity in HTTP access log

### LINKED ISSUE #9990: feat: include authenticated user identity in HTTP access log

Currently the HTTP access log (gunicorn/Apache) always shows - for the user field because pgAdmin uses session-based auth rather than HTTP Basic auth. For deployments with multiple users this makes it impossible to correlate access log entries with individual users for auditing, security review, or troubleshooting. I'd like to show the authenticated user's username on the requests in the HTTP access log.

ardentperf:
Set an X-Remote-User response header containing the authenticated username on every request. This allows the access log to be configured to include user identity via standard log format directives (%({x-remote-user}o)s in gunicorn, %{X-Remote-User}o in Apache) without requiring any changes to pgAdmin's session or auth behaviour.

Closes #9990

Default gunicorn access log format is at https://gunicorn.org/reference/settings/?h=access_log#access_log_format and https://github.com/benoitc/gunicorn/blob/9bc5891b4b06f25a8ce0e707053dcb2fb9bf638c/gunicorn/config.py#L1413 ; I confirmed that all other fields are default; this PR only changes the user name field.

Performed an end-to-end test on kubernetes with KIND

With this PR:

pgadmin-loguserid-branch

Master Branch:

pgadmin-master-branch

## Summary by CodeRabbit

* **New Features**
  * Option to include the authenticated username in responses so HTTP access logs can record user identity, improving audit trails and monitoring.
  * New configuration toggle (disabled by default) to enable or disable this behavior.
  * When enabled and a user is authenticated, a response header carries a latin-1-safe username; when disabled or unauthenticated, no username is added.

### DIFF
--- a/web/pgadmin/__init__.py
+++ b/web/pgadmin/__init__.py
@@ -863,8 +863,14 @@ def create_app(app_name=None):
 
     @app.after_request
     def after_request(response):
-        if current_user.is_authenticated:
-            response.headers['X-Remote-User'] = current_user.username
+        if config.LOG_AUTHENTICATED_USER:
+            if current_user.is_authenticated and current_user.username:
+                # Encode as latin-1 to avoid gunicorn 500s for unicode names
+                safe = current_user.username.encode(
+                    'latin-1', 'replace').decode('latin-1')
+                response.headers['X-Remote-User'] = safe
+            else:
+                response.headers.pop('X-Remote-User', None)
 
         if 'key' in request.args:
             domain = dict()
--- a/web/pgadmin/__init__.py
+++ b/web/pgadmin/__init__.py
@@ -863,6 +863,9 @@ def create_app(app_name=None):
 
     @app.after_request
     def after_request(response):
+        if current_user.is_authenticated:
+            response.headers['X-Remote-User'] = current_user.username
+
         if 'key' in request.args:
             domain = dict()
             if config.COOKIE_DEFAULT_DOMAIN and \
--- a/pkg/docker/gunicorn_config.py
+++ b/pkg/docker/gunicorn_config.py
@@ -8,7 +8,10 @@ gunicorn.SERVER_SOFTWARE = "Python"
 # Include the authenticated user identity in the access log.
 # %({x-remote-user}o)s reads the X-Remote-User response header set by pgAdmin
 # for authenticated requests; unauthenticated requests log '-'.
-access_log_format = '%(h)s %(l)s %({x-remote-user}o)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"'
+access_log_format = (
+    '%(h)s %(l)s %({x-remote-user}o)s %(t)s "%(r)s" %(s)s %(b)s '
+    '"%(f)s" "%(a)s"'
+)
 
 if JSON_LOGGER:
     logconfig_dict = {
--- a/pkg/docker/gunicorn_config.py
+++ b/pkg/docker/gunicorn_config.py
@@ -5,6 +5,11 @@ from config import JSON_LOGGER, CONSOLE_LOG_LEVEL, CONSOLE_LOG_FORMAT_JSON
 
 gunicorn.SERVER_SOFTWARE = "Python"
 
+# Include the authenticated user identity in the access log.
+# %({x-remote-user}o)s reads the X-Remote-User response header set by pgAdmin
+# for authenticated requests; unauthenticated requests log '-'.
+access_log_format = '%(h)s %(l)s %({x-remote-user}o)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"'
+
 if JSON_LOGGER:
     logconfig_dict = {
         "version": 1,
--- a/web/config.py
+++ b/web/config.py
@@ -301,6 +301,9 @@ else:
 LOG_ROTATION_SIZE = 10  # In MBs
 LOG_ROTATION_AGE = 1440  # In minutes
 LOG_ROTATION_MAX_LOG_FILES = 90  # Maximum number of backups to retain
+# Include the authenticated username in the X-Remote-User response header so
+# it can be captured in the HTTP access log. Disabled by default.
+LOG_AUTHENTICATED_USER = False
 ##########################################################################
 # Server Connection Driver Settings
 ##########################################################################


web/pgadmin/__init__.py
```diff
@@ -863,6 +863,9 @@ def before_request():
 
     @app.after_request
     def after_request(response):
+        if current_user.is_authenticated:
+            response.headers['X-Remote-User'] = current_user.username
```
dpage: HTTP header values must encode to Latin-1, but `current_user.username` isn't guaranteed to be ASCII/Latin-1. OAuth2 (`_resolve_username` can return `preferred_username`, `sub`, or a configured `OAUTH2_USERNAME_CLAIM`), Kerberos principals, and LDAP-mapped usernames can all contain Cyrillic/CJK/accented characters.

Verified against the Werkzeug currently pinned in the project:
```
Gorkov (Cyrillic)            -> UnicodeEncodeError: 'latin-1' codec can't encode...
alice\r\nX-Injected: 1       -> ValueError: Header values must not contain newline characters.
```

Because the header is set unconditionally for every authenticated user, a user with a non-Latin-1 username will raise during response serialization and get a 500 on **every** request — effectively locked out of pgAdmin. (The CR/LF case is already blocked by Werkzeug, so there's no header-injection vuln, but it would also 500.)

Suggest sanitizing before setting:
```python
if current_user.is_authenticated and current_user.username:
    # HTTP headers are latin-1 only; avoid 500s for unicode usernames
    safe = current_user.username.encode('latin-1', 'replace').decode('latin-1')
    response.headers['X-Remote-User'] = safe
```

dpage (commented):
Thanks for this — it's a clean, well-scoped change and a genuinely useful feature. One blocking concern and a couple of smaller points.

**🔴 Must fix: non-Latin-1 usernames will 500 every authenticated request.** See the inline comment — HTTP header values must encode to Latin-1, but `current_user.username` can legitimately contain characters outside that range (OAuth2 `preferred_username`/`sub`/`OAUTH2_USERNAME_CLAIM`, Kerberos principals, LDAP attributes). Since the header is set unconditionally for every authenticated user, such a user would hit `UnicodeEncodeError` during response serialization and be locked out entirely. Sanitizing the value before setting the header avoids this.

**🟡 Tests:** it would be good to add a small regression test asserting the header is present for an authenticated request and absent for an anonymous one — ideally exercising a non-ASCII username to guard the case above.

**ℹ️ Notes (non-blocking):**
- The `access_log_format` change only lands for Docker deployments; standard package/pip server installs running their own gunicorn won't pick up the username field without adding the directive themselves. Consistent with the PR's stated Docker/k8s focus — just flagging the scope.
- With `JSON_LOGGER` enabled the username ends up embedded inside the access-log *message* string rather than as a discrete JSON field, since the JSON formatter wraps gunicorn's rendered access line. Works, but operators expecting a structured field may be surprised.

dpage (commented):
Thanks for this — it's a clean, well-scoped change and a genuinely useful feature. One blocking concern and a couple of smaller points.

**🔴 Must fix: non-Latin-1 usernames will 500 every authenticated request.** See the inline comment — HTTP header values must encode to Latin-1, but `current_user.username` can legitimately contain characters outside that range (OAuth2 `preferred_username`/`sub`/`OAUTH2_USERNAME_CLAIM`, Kerberos principals, LDAP attributes). Since the header is set unconditionally for every authenticated user, such a user would hit `UnicodeEncodeError` during response serialization and be locked out entirely. Sanitizing the value before setting the header avoids this.

**🟡 Tests:** it would be good to add a small regression test asserting the header is present for an authenticated request and absent for an anonymous one — ideally exercising a non-ASCII username to guard the case above.

**ℹ️ Notes (non-blocking):**
- The `access_log_format` change only lands for Docker deployments; standard package/pip server installs running their own gunicorn won't pick up the username field without adding the directive themselves. Consistent with the PR's stated Docker/k8s focus — just flagging the scope.
- With `JSON_LOGGER` enabled the username ends up embedded inside the access-log *message* string rather than as a discrete JSON field, since the JSON formatter wraps gunicorn's rendered access line. Works, but operators expecting a structured field may be surprised.

web/pgadmin/__init__.py
dpage: HTTP header values must encode to Latin-1, but `current_user.username` isn't guaranteed to be ASCII/Latin-1. OAuth2 (`_resolve_username` can return `preferred_username`, `sub`, or a configured `OAUTH2_USERNAME_CLAIM`), Kerberos principals, and LDAP-mapped usernames can all contain Cyrillic/CJK/accented characters.

Verified against the Werkzeug currently pinned in the project:
```
Gorkov (Cyrillic)            -> UnicodeEncodeError: 'latin-1' codec can't encode...
alice\r\nX-Injected: 1       -> ValueError: Header values must not contain newline characters.
```

Because the header is set unconditionally for every authenticated user, a user with a non-Latin-1 username will raise during response serialization and get a 500 on **every** request — effectively locked out of pgAdmin. (The CR/LF case is already blocked by Werkzeug, so there's no header-injection vuln, but it would also 500.)

Suggest sanitizing before setting:
```python
if current_user.is_authenticated and current_user.username:
    # HTTP headers are latin-1 only; avoid 500s for unicode usernames
    safe = current_user.username.encode('latin-1', 'replace').decode('latin-1')
    response.headers['X-Remote-User'] = safe
```

dpage:
Please also address the python style issue causing CI to fail.

Thanks!

web/pgadmin/__init__.py
```diff
@@ -863,6 +863,9 @@ def before_request():
 
     @app.after_request
     def after_request(response):
+        if current_user.is_authenticated:
```
mzabuawala: Since not everyone wants to send user information in plain text in every response, we must make it configurable.

`LOG_AUTHENTICATED_USER=True/False`

mzabuawala:
```suggestion
        if current_user.is_authenticated:
            response.headers['X-Remote-User'] = current_user.username
        else:
            # prevents any accidental reuse if middleware or future code sets the header earlier
            response.headers.pop('X-Remote-User', None)
```

ardentperf:
Apologies, this past week has been a bit crazy - haven’t forgotten and should get to it this week 🤞

dpage:
A few additional notes to fold into your next pass, on top of the inline comments above:

**1. The pep8/CI failure** is just the new `access_log_format` line (96 chars > 79). `# noqa` won't help since we run `pycodestyle` directly, so please wrap it:
```python
access_log_format = (
    '%(h)s %(l)s %({x-remote-user}o)s %(t)s '
    '"%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"'
)
```

**2. Default the config flag to off.** Building on @mzabuawala's `LOG_AUTHENTICATED_USER` suggestion: `X-Remote-User` is sent on *every response to the client*, not just consumed by the log — so it's visible to proxies, CDNs, TLS middleboxes, browser extensions, etc. For a server-side logging feature that's broader exposure than needed, so the flag should default to **False** (opt-in), with a docs note that enabling it surfaces the username in response headers.

**3. Optional design alternative.** If cross-deployment portability (Apache `%{X-Remote-User}o`) isn't essential, you could stash the name in the WSGI environ instead of a response header — `request.environ['x_remote_user'] = current_user.username` — and log it via gunicorn's environ atom `%({x_remote_user}e)s`. That keeps the identity entirely server-side (never sent to the client) and sidesteps the Latin-1 header-encoding 500 altogether. The trade-off is it's gunicorn-specific, whereas the response-header approach also works under Apache — so it's a judgement call, not a request. (Worth confirming the `{}e` atom behaviour against the pinned gunicorn if you go that route.)

Thanks again for the contribution — no rush, just capturing these so they're in one place for your update.

ardentperf:
> **3. Optional design alternative.** If cross-deployment portability (Apache `%{X-Remote-User}o`) isn't essential, you could stash the name in the WSGI environ instead of a response header — `request.environ['x_remote_user'] = current_user.username` — and log it via gunicorn's environ atom `%({x_remote_user}e)s`. That keeps the identity entirely server-side (never sent to the client) and sidesteps the Latin-1 header-encoding 500 altogether. The trade-off is it's gunicorn-specific, whereas the response-header approach also works under Apache — so it's a judgement call, not a request. (Worth confirming the `{}e` atom behaviour against the pinned gunicorn if you go that route.)

The main use case I'm interested in right now is containerized deployments on kubernetes, so gunicorn covers the immediate case. But personally I think there's value in the overall idea to many pgAdmin users. My vote would be for the opt-in apache-compatible approach, which makes it accessible others who are also interested in using this outside kubernetes for a more complete audit picture (eg. seeing who has downloaded data from which database and how many bytes were in the download request). I also don't think there should be any concerns of sensitivity (even in regulated environments) around simply knowing which authenticated user made which requests. It's fairly standard for an access log. This being said, I'm open to either option if someone felt there was a strong argument for keeping everything server-side.

web/pgadmin/__init__.py
ardentperf: Making sure I understand this. I guess the main concern here is inconsistency - it would be a bit weird to overwrite the header sometimes and not others - and users wouldn't know who set the value they see in the field. If we're going to set a value in this field sometimes, then it's best to always control the contents of the field. That makes sense.

mzabuawala: My understanding is that the concern is less about the current implementation and more about ensuring the header is always under pgAdmin's control. If we only overwrite it for authenticated users, then in other cases it's unclear whether the value came from pgAdmin itself or was set earlier by middleware or some future code path. Explicitly removing it when there is no authenticated user keeps the behavior deterministic and avoids any ambiguity about the source of the header value.

web/pgadmin/__init__.py
ardentperf: I did a little bit of poking around but not seeing an easy way to add tests for this (or for the change as a whole) since they depend on having a server like gunicorn in the loop  :-/

I'm a little curious how you reproduced the error and what your setup is... pgadmin wouldnt let me create accounts with special chars in email, and i didn't have OAuth2 setup for testing

I repro'd the error by directly calling gunicorn's `to_bytestring()` function, just to verify it

ardentperf:
pushed updates to the PR addressing comments. CI failures on python style should be addressed.

Default pgAdmin behavior is unchanged; `X-Remote-User` header is only managed if the user enables the `LOG_AUTHENTICATED_USER` configuration setting.

I tested with the helm chart on k8s and confirmed that there's no logging by default, and logging is enabled when I add this to my `values.yaml` file:

```
extraEnvVars:
  - name: PGADMIN_CONFIG_LOG_AUTHENTICATED_USER
    value: "True"
```

Is this a good name for the config setting? Technically it controls the header, and I've updated the default gunicorn logging to automatically pick it up. For users who consume this through docker and helm, the practical effect is that the config does enable logging.

ardentperf:
after a bit more thought, i'm also testing the WSGI environ approach to see if this works.

ardentperf:
i just tried testing `request.environ['x_remote_user'] = current_user.username` and `%({x_remote_user}e)s` but gunicorn did not log usernames. now i remember trying this before too, and i never did figure out why it doesn't work...

ardentperf:
I think the culprit is here:

https://github.com/miguelgrinberg/flask-socketio/blob/v5.6.1/src/flask_socketio/__init__.py#L40

> flask_socketio wraps app.wsgi_app with its own middleware (SocketioMiddleware). Its __call__ method copies the environ before passing it to Flask. The copy at line 40 means Flask receives a new dict — a shallow copy of gunicorn's original environ. Any mutations Flask makes (including request.environ['x_remote_user'] = ... in after_request) go into the copy. Gunicorn's original environ dict is untouched. When gunicorn calls self.log.access(resp, req, environ, ...), it passes its original dict, which never has x_remote_user set. This was confirmed by logging id() of the environ in both after_request and gunicorn's handle_request — they were different on every request.

dpage:
Merged to `master` as 0c6740f14 (squashed; the SHA differs so GitHub won't auto-close this). I folded in one small hardening on top of your latest commit: as well as the latin-1 transliteration, the username now has non-printable characters (including CR/LF) stripped before being set as the `X-Remote-User` header. Those control characters are valid latin-1 so they survived the `'replace'` round-trip, and Werkzeug rejects them in header values, which would have 500'd every response for an affected user. Thanks for the contribution, and for bearing with the review back-and-forth!

ardentperf:
thanks all!

PR #10005, opened 2026-06-05
SchemaView: DataGridView audit + deferredDepChange protocol

asheshv:
## Summary

Three commits addressing silent-failure bugs in `DataGridView` and introducing a queue-based `deferredDepChange` protocol for `SchemaView`, with five schemas adopting the new contract.

### 1. `fix(datagridview): correct silent-failure bugs in DataGridView`

Four bugs that silently no-op'd:

- **Feature priority comparator typo** — `f1.priorty` is undefined, so `<` was always false and features never sorted.
- **Row className** — `classList.join[' ']` bracket-accesses `Array.prototype.join` (returns undefined). Every grid row rendered with no CSS classes.
- **Reorder hover class** — `classList?.append(...)` — Arrays have no `.append`; the optional chain silently no-op'd, so the hover-row highlight never applied during drag-reorder.
- **Inherited-column lock** — `inSchemaWithColumnCheck` short-circuited on `isNew(state)` before reading `inheritedfrom`. Inherited columns in CREATE mode (no `attnum`) bypassed the lock, making Name and datatype editable on columns that should mirror the parent.

### 2. `feat(schemaview): introduce deferredDepChange protocol`

New queue-based protocol so schemas can run async work (confirm dialogs, server fetches) atomically across a state change:

- **`SchemaState/reducer.js`** — appends resolved callbacks to `data.__deferred__` with the canonical `actionObj` shape.
- **`hooks/useSchemaState.js`** — `drainDeferredQueue` processes promises in order; rejections surface via `pgAdmin.Browser.notifier.error` (falls back to `console.error` when the notifier is unavailable). Drain `useEffect` keyed on the `__deferred__` ref identity, not `.length`, so replaced entries reliably re-trigger.
- **`DepListener.js`** — prefix-match guard using delimiter-terminated joins (so `shared` doesn't match `shared_username`), cached source/dest keys, defensive deep-copy of source path on registration.
- **`utils/listenDepChanges.js`** — full JSDoc contract: return shape, opt-out via `return undefined`, side-effect-before-resolve rule, `actionObj` fields, drain error-surfacing.

### 3. `fix(schemas): align deferredDepChange callbacks with new protocol`

Five schemas adopt the contract — real bugs fixed along the way:

- **`exclusion_constraint.ui.js`** (amname) — opt out when `amname` unchanged or `columns` empty; pre-fix code warned about clearing an empty collection on the very first selection.
- **`index.ui.js`** (amname) — explicit `return undefined` opt-out in place of `Promise.resolve(()=>{})` round-trip.
- **`table.ui.js`** (typname + coll_inherits) — `typeTable?.oftype_columns ?? []` guard against stale `ofTypeTables`; `changeColumnOptions` hoisted out of the resolved callback (per protocol); same-length inherits swap handling; stale-OID guard preventing `col.inheritedid != undefined` from dropping local user-added columns.
- **`foreign_table.ui.js`** (inherits) — same swap + stale-OID guard.
- **`azure_schema.ui.js`** (is_authenticating) — orphan-Promise fix (was never resolving in the non-match branch, leaving a permanent `__deferred__` entry); source-path comparison against `source[source.length - 1]` (arrays, not strings); auth failure surfaced via notifier with stale `auth_code` cleared.

## Test plan

- [x] `cd web && yarn run jest` — **151 suites / 884 tests passing**
- [x] `cd web && yarn linter` — **clean**
- [ ] Manual UI: exclusion-constraint create dialog with empty columns (amname pick should no longer warn)
- [ ] Manual UI: child-table inherits — add/remove/swap parent tables (columns refresh correctly)
- [ ] Manual UI: Azure interactive auth — success path, failure path (error surfaces, dialog unblocks)
- [ ] Manual UI: column inherited from parent — Name field is read-only in CREATE mode

## Notes

- 3 commits, clean linear history off `master`.
- New tests added per concern: `feature_register.spec.js`, `no_bracket_on_prototype_method.spec.js`, `dep_listener.spec.js`, `reducer_deferred.spec.js`, `deferred_drain.spec.js`, `drain_useeffect_race.spec.jsx`, plus `*.deferred.spec.js` for each of the five affected schemas.
- The `deferredDepChange` JSDoc contract in `utils/listenDepChanges.js` is the authoritative reference for schema authors.

### DIFF
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/foreign_tables/static/js/foreign_table.ui.js
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/foreign_tables/static/js/foreign_table.ui.js
@@ -129,60 +129,47 @@ export default class ForeignTableSchema extends BaseUISchema {
         options: obj.fieldOptions.tables,
         optionsLoaded: (res)=>obj.inheritedTableList=res,
         deferredDepChange: (state, source, topState, actionObj)=>{
-          return new Promise((resolve)=>{
-            // current table list and previous table list
-            let newColInherits = state.inherits || [];
-            let oldColInherits = actionObj.oldState.inherits || [];
+          const newColInherits = state.inherits || [];
+          const oldColInherits = actionObj.oldState.inherits || [];
+          const added = _.difference(newColInherits, oldColInherits);
+          const removed = _.difference(oldColInherits, newColInherits);
 
-            let tabName;
-            let tabColsResponse;
+          // REMOVE takes precedence: any removed parent means stale
+          // columns to clean up. Covers pure shrink AND same-length
+          // swap (e.g. multi-select replace) — without this branch a
+          // swap would leave the removed parent's columns sitting in
+          // the grid until the user did something else.
+          if(removed.length > 0) {
+            const removeOid = this.getTableOid(removed[0]);
+            // Guard: if inheritedTableList is stale and the removed
+            // table can't be resolved to an OID, opt out. Filtering on
+            // `inheritedid != undefined` would silently drop local
+            // user-added columns (`null == undefined` in JS).
+            if(removeOid == null) return undefined;
+            return Promise.resolve((tmpstate)=>({
+              adding_inherit_cols: false,
+              columns: (tmpstate.columns || [])
+                .filter((col)=>col.inheritedid != removeOid),
+            }));
+          }
 
-            // Add columns logic
-            // If new table is added in list
-            if(newColInherits.length > 1 && newColInherits.length > oldColInherits.length) {
-              // Find newly added table from current list
-              tabName = _.difference(newColInherits, oldColInherits);
-              tabColsResponse = obj.getColumns({attrelid: this.getTableOid(tabName[0])});
-            } else if (newColInherits.length == 1) {
-              // First table added
-              tabColsResponse = obj.getColumns({attrelid: this.getTableOid(newColInherits[0])});
-            }
-
-            if(tabColsResponse) {
-              tabColsResponse.then((res)=>{
-                resolve((tmpstate)=>{
-                  let finalCols = res.map((col)=>obj.columnsObj.getNewData(col));
-                  finalCols = [...tmpstate.columns, ...finalCols];
-                  return {
-                    adding_inherit_cols: false,
-                    columns: finalCols,
-                  };
-                });
-              });
-            }
-
-            // Remove columns logic
-            let removeOid;
-            if(newColInherits.length > 0 && newColInherits.length < oldColInherits.length) {
-              // Find deleted table from previous list
-              tabName = _.difference(oldColInherits, newColInherits);
-              removeOid = this.getTableOid(tabName[0]);
-            } else if (oldColInherits.length === 1 && newColInherits.length < 1) {
-              // We got last table from list
-              tabName = oldColInherits[0];
-              removeOid = this.getTableOid(tabName);
-            }
-            if(removeOid) {
-              resolve((tmpstate)=>{
-                let finalCols = tmpstate.columns;
-                _.remove(tmpstate.columns, (col)=>col.inheritedid==removeOid);
+          // Pure ADD: list grew without any removals.
+          if(added.length > 0) {
+            const fetchOid = this.getTableOid(added[0]);
+            if(fetchOid == null) return undefined;
+            return obj.getColumns({attrelid: fetchOid}).then((res)=>(
+              (tmpstate)=>{
+                const fetched = res.map((col)=>obj.columnsObj.getNewData(col));
                 return {
                   adding_inherit_cols: false,
-                  columns: finalCols
+                  columns: [...(tmpstate.columns || []), ...fetched],
                 };
-              });
-            }
-          });
+              }
+            ));
+          }
+
+          // Lists are equivalent — no work to do.
+          return undefined;
         },
       },
       {
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/static/js/column.ui.js
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/columns/static/js/column.ui.js
@@ -88,15 +88,19 @@ export default class ColumnSchema extends BaseUISchema {
     }
 
     if(this.nodeInfo &&  ('schema' in this.nodeInfo)) {
-      if(this.isNew(state)) {
-        return false;
-      }
-
-      // We will disable control if it's system columns
-      // inheritedfrom check is useful when we use this schema in table node
-      // inheritedfrom has value then we should disable it
+      // Parent-table inheritance locks the column in BOTH create and
+      // edit modes — its Name and datatype mirror the parent. The
+      // `isNew(state)` shortcut below must NOT run first, otherwise
+      // inherited columns become editable on a fresh child-table
+      // dialog. (`OF TYPE` inheritance uses `inheritedfromtype` and is
+      // handled by the dedicated `editable` callback further down —
+      // not by this method.)
       if (!isEmptyString(state.inheritedfrom)){
         return true;
+      }
+
+      if(this.isNew(state)) {
+        return false;
       }
 
       // ie: it's position is less than 1
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/constraints/exclusion_constraint/static/js/exclusion_constraint.ui.js
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/constraints/exclusion_constraint/static/js/exclusion_constraint.ui.js
@@ -284,6 +284,16 @@ export default class ExclusionConstraintSchema extends BaseUISchema {
       type: 'select', group: gettext('Definition'),
       options: this.fieldOptions.amname,
       deferredDepChange: (state, source, topState, actionObj)=>{
+        // Opt out cleanly when nothing would change: amname unchanged
+        // or no columns to wipe. Previously this queued a confirmation
+        // dialog unconditionally.
+        // actionObj.oldState is guaranteed by the reducer to be the
+        // pre-dispatch clone; no need for optional chaining (and the
+        // cancel branch below accesses it unconditionally).
+        if(state.amname === actionObj.oldState.amname
+            || !state.columns?.length) {
+          return undefined;
+        }
         return new Promise((resolve)=>{
           pgAdmin.Browser.notifier.confirm(
             gettext('Change access method?'),
@@ -311,11 +321,9 @@ export default class ExclusionConstraintSchema extends BaseUISchema {
               }));
             },
             function() {
-              resolve(()=>{
-                return {
-                  amname: actionObj.oldState.amname,
-                };
-              });
+              resolve(()=>({
+                amname: actionObj.oldState.amname,
+              }));
             }
           );
         });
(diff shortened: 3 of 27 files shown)


asheshv:
# Manual Test Plan

## Prerequisites

- pgAdmin 4 running from `fix/deferreddepchange-protocol` branch (e.g. `cd web && yarn webpacker:watch` + `python pgAdmin4.py`)
- Local Postgres 17 (port 5432) registered as a pgAdmin server
- A test database created on that server, e.g. `pgadmin_pr10005_test`
- A schema inside the test DB, e.g. `pr_test` (or use `public`)
- Browser DevTools open to **Console** + **Network** tabs throughout — many checks require watching console output

---

## Section A — DataGridView mechanical fixes

### A1. Feature priority ordering (`priorty` typo fix)

**Setup:**
- Open any node that uses DataGridView with multiple features (e.g. Table → Columns sub-grid).

**Steps:**
1. Navigate: Browser tree → server → DB → Schemas → `pr_test` → Tables → right-click → **Create → Table**.
2. Click the **Columns** tab.
3. Click **+** to add 3 rows.

**Expected:**
- Each row shows controls in stable column order: select checkbox (if any) → drag handle → cell controls → delete button on the right.
- No console warnings like `feature applied twice` or React `key` collision warnings.
- Resizing one column doesn't reorder features.

**Validates:** `featurePriorityCompare` now sorts (was always returning `false` due to the `priorty` typo).

---

### A2. Row CSS class rendering (`classList.join(' ')` fix)

**Setup:** Same Create Table dialog with at least 3 columns added.

**Steps:**
1. Inspect the DOM of a row in the Columns sub-grid (right-click row → Inspect).
2. Locate the row's outer `<div data-test=\"data-table-row\">`.
3. Check the `class` attribute.

**Expected:**
- `class` is **populated** (e.g. `DataGridView-tableRow some-other-feature-class`).
- **Not** `class=\"\"` or `class=\"undefined\"`.
- Hovering / selecting a row visibly changes its style (alternate background, hover highlight).

**Validates:** `className={classList.join(' ')}` now applies (was `className={classList.join[' ']}`, which evaluated to `undefined`).

---

### A3. Drag-reorder hover highlight (`.push` not `.append`)

**Setup:** Create Table dialog with **4+ columns** added.

**Steps:**
1. Grab a column's drag handle (left side of the row).
2. **Hover** the drag preview over a different row **without releasing** the mouse.
3. Visually inspect the row currently being hovered over.

**Expected:**
- The hovered target row shows a **highlight** (background tint, border color, etc. — exact style is theme-dependent).
- Move to another row → highlight follows.
- Drop → highlight clears.

**Validates:** `classList?.push('DataGridView-tableRowHovered')` (was `.append(...)`, which silently no-op'd because Arrays have no `.append`).

---

### A4. Inherited-column lock in CREATE mode

**Setup:**
```sql
CREATE TABLE pr_test.parent_a (a int PRIMARY KEY, b text);
```
Refresh the browser tree so `parent_a` is visible.

**Steps:**
1. Right-click `Tables` in `pr_test` → **Create → Table**.
2. **General** tab: name = `child_a`.
3. **Columns** tab: confirm grid is empty initially.
4. **Advanced** tab → **Inherits from table(s)**: select `pr_test.parent_a`.
5. Return to **Columns** tab.

**Expected:**
- Two rows appear (`a`, `b`) with `inheritedfrom` set to `pr_test.parent_a`.
- The **Name** field on each inherited row is **disabled / read-only**.
- The **datatype** column is also disabled.
- You cannot rename or retype `a` or `b`.

**Negative case:**
- Add a *local* column `extra` (not inherited). Its Name and datatype **are** editable.
- An OF-TYPE table (Advanced → \"Of type\"): the lock should NOT engage for `inheritedfromtype` — those columns are editable per their dedicated `editable` callback.

**Validates:** `inSchemaWithColumnCheck` now reads `inheritedfrom` before the `isNew(state)` shortcut.

---

## Section B — deferredDepChange protocol infrastructure

### B1. Drain error surfacing via `notifier.error`

**Setup:** Use the Azure cloud provisioning flow with no network (simplest reproducible failure).

**Steps:**
1. Begin the Azure auth flow: trigger `auth_btn` click, set `auth_type` to `interactive_browser_credential`.
2. Block outbound HTTPS / disconnect network.
3. Wait for `getAuthCode` to fail.

**Expected:**
- A **toast / notifier error** appears at top of the pgAdmin window (red/orange) reading something like *\"Azure authentication failed: …\"*.
- The Azure config dialog **unblocks** — `is_authenticating` returns to false, fields become re-editable.
- DevTools console does **not** show an unhandled rejection or silent `console.error` as the only signal.

**Validates:** `drainDeferredQueue` routes rejections through `pgAdmin.Browser.notifier.error`.

**Negative-case sub-test:** Temporarily delete `pgAdmin.Browser.notifier` in DevTools (`delete pgAdmin.Browser.notifier`), then retry. The console should fall back to `console.error` rather than crashing the drain.

---

### B2. DepListener prefix-match isolation

**Setup:** Any schema with fields whose IDs share a prefix (e.g. `shared` vs `shared_username`).

**Steps:**
1. From DevTools, type:
   ```js
   schemaState.getDepChange(['shared'], schemaState.data, {type:'TEST'})
   ```
2. Verify that listeners registered for `['shared_username']` are **not** invoked.

**Expected:** Only listeners with exact prefix `['shared']` fire.

**Validates:** Prefix-match guard uses delimiter-terminated joins. (Mostly covered by `dep_listener.spec.js`.)

---

### B3. Drain useEffect — replaced-entry race

**Setup:** Exclusion Constraint amname is a good repro candidate.

**Steps:**
1. Open the Exclusion Constraint create dialog.
2. Add one column.
3. Switch `amname` from `btree` → `gist` → `hash` → `btree` rapidly (within ~500ms).

**Expected:**
- Each amname change should be reflected eventually — no transient state where the dialog gets stuck mid-transition.
- The columns collection should be cleared appropriately after the final confirmation.
- No console warnings about lost dispatches.

**Validates:** Drain `useEffect` keying on `__deferred__` ref identity (not `.length`).

---

## Section C — Per-schema deferredDepChange adoption

### C1. Exclusion constraint `amname` opt-out (empty columns)

**Steps:**
1. Open Create Table → **Constraints** tab → **Exclusion Constraint** sub-tab → **+** → **Edit** the new row.
2. With **columns empty**, change `amname` from default `btree` → `gist`.

**Expected:** **No confirmation dialog opens.** State transitions silently.

3. Add one column to the exclusion constraint.
4. Change `amname` again `gist` → `hash`.

**Expected:** A confirm dialog **does** open: *\"Change access method? Changing access method will clear columns collection.\"*

5. Click **Cancel**.

**Expected:** `amname` reverts to `gist`. The column is still present.

6. Change `amname` again, click **OK**.

**Expected:** The column is removed.

7. With one column added, change `amname` to its **current** value (no actual change).

**Expected:** No dialog (opt-out fires because `amname === oldState.amname`).

**Validates:** Opt-out when amname unchanged or columns empty; cancel reverts; OK clears columns.

---

### C2. Index `amname` (sanity)

**Steps:**
1. Right-click an existing table → **Create → Index**.
2. Switch `amname` from `btree` to anything else **with no columns added**.

**Expected:** No confirm dialog.

3. Add an index column.
4. Switch `amname` again.

**Expected:** Confirm dialog appears.

5. Click Cancel.

**Expected:** `amname` reverts, column still present.

---

### C3. Table OF TYPE — stale `ofTypeTables` guard

**Setup:**
```sql
CREATE TYPE pr_test.row_t AS (x int, y text);
```

**Steps:**
1. Right-click `Tables` → **Create → Table**.
2. **Advanced** tab → **Of type**: select `pr_test.row_t`.
3. Go to **Columns** tab.

**Expected:** Two columns `x` and `y` appear, derived from `row_t`, locked.

4. **Advanced** tab → change `Of type` to **empty**.

**Expected:** Confirm dialog: *\"Remove column definitions?\"* → click OK → columns cleared.

**Stale-options edge case (optional):**
5. In DevTools, before changing `Of type`, set `obj.ofTypeTables = []` to simulate stale options.
6. Trigger the typname change.

**Expected:** No TypeError; falls back to empty columns list cleanly.

**Validates:** `typeTable?.oftype_columns ?? []` guard.

---

### C4. Table inherits — add/remove/swap

**Setup:**
```sql
CREATE TABLE pr_test.parent_b (b1 int PRIMARY KEY, b2 text);
CREATE TABLE pr_test.parent_c (c1 int PRIMARY KEY, c2 text);
```

**Steps:**
1. Open **Create Table** dialog, name = `child_bc`.
2. **Advanced** tab → **Inherits from table(s)**: add `parent_b`.

**Expected:** Columns tab now shows `b1`, `b2` (inherited).

3. Add a local column `local_extra int`.

**Expected:** All 3 columns shown.

4. **Advanced** tab → add `parent_c` to inherits.

**Expected:** Columns tab shows `b1`, `b2`, `c1`, `c2`, `local_extra`.

5. **Advanced** tab → remove `parent_b` from inherits.

**Expected:** Columns tab shows `c1`, `c2`, `local_extra`. The local `local_extra` **must still be present** (stale-OID guard verification).

6. **Swap test:** From state above (only `parent_c` inherits + `local_extra`), in one operation **replace** `parent_c` with `parent_b` (deselect c, select b — same length, different content).

**Expected:** `c1`, `c2` removed, `b1`, `b2` added. `local_extra` still present.

**Validates:** Same-length swap handling; stale-OID guard preserves local columns.

---

### C5. Foreign table inherits — same as C4

**Setup:**
```sql
CREATE EXTENSION IF NOT EXISTS postgres_fdw;
CREATE SERVER pr_fdw FOREIGN DATA WRAPPER postgres_fdw
  OPTIONS (host '127.0.0.1', dbname 'pgadmin_pr10005_test');
```

**Steps:** Repeat C4's add/remove/swap workflow under **Foreign Tables** → **Create Foreign Table**.

**Expected:** Same behavior as C4. Local columns survive parent removal/swap.

---

### C6. Azure interactive auth — success + failure

**Steps (success path):**
1. **Tools → Deploy Cloud Instance → Azure** (or your build's Azure entry).
2. Choose `interactive_browser_credential` for `auth_type`.
3. Click the **Authenticate** button.

**Expected:**
- `is_authenticating` flips to true, button text changes / disables.
- A device-code is displayed.
- Browser tab opens for Microsoft login → complete it → `auth_code` field populates.
- `is_authenticating` returns to false.

**Steps (failure path):**
4. Reload the dialog. Disconnect network / block Azure endpoints.
5. Click Authenticate.

**Expected:**
- Notifier error toast: *\"Azure authentication failed: <message>\"*.
- `is_authenticating` returns to false (dialog unblocks).
- **`auth_code` is cleared** (no stale code from a previous successful attempt).
- No orphan `__deferred__` entry (check DevTools: `schemaState.data.__deferred__.length === 0` after settle).

**Validates:** Orphan-Promise fix; source-path comparison via `source[length-1]`; auth failure surfaced via notifier; stale `auth_code` cleared.

---

## Final regression sweep

After all sections pass:

- [ ] Open at least 3 existing tables in **Edit** mode → tabs render normally, no console errors.
- [ ] Save a table with inherited columns → server-side SQL preview matches expectation.
- [ ] Save an exclusion constraint after the amname dance → produced SQL is correct.
- [ ] Close all open dialogs → no lingering deferred state in DevTools.
- [ ] Browser tree refresh / re-expand → all created objects are visible.

---

## Cleanup

```sql
DROP TABLE IF EXISTS pr_test.child_a, pr_test.child_bc CASCADE;
DROP TABLE IF EXISTS pr_test.parent_a, pr_test.parent_b, pr_test.parent_c CASCADE;
DROP TYPE  IF EXISTS pr_test.row_t;
DROP SERVER IF EXISTS pr_fdw CASCADE;
DROP SCHEMA IF EXISTS pr_test CASCADE;
DROP DATABASE IF EXISTS pgadmin_pr10005_test;
```

Issue #7811, opened 2024-08-13
Servers are re-imported at every container startup


vlaborie:
**Describe the bug**

Hello,

I use pgadmin from Docker image `dpage/pgadmin4:8.10` in server mode and mount the following file to `/pgadmin4/servers.json` :

~~~
    {
        "Servers": {
            "1": {
                "Name": "Local",
                "Group": "Servers",
                "Shared": true,
                "SharedUsername": "admin",
                "Username": "admin",
                "Comment": "Local Postgresql server",
                "Host": "postgresql-rw.postgresql.svc.cluster.local",
                "Port": 5432,
                "MaintenanceDB": "postgres",
                "ConnectionParameters": {
                    "sslmode": "verify-full",
                    "sslcert": "tls:/tls.crt",
                    "sslkey": "tls:/tls.key",
                    "sslrootcert": "tls:/ca.crt"
                }
            }
        }
    }
~~~

The server is imported at container startup but it is imported at **every** startup :

~~~
Added 0 Server Group(s) and 1 Server(s).
~~~

You can see in the following screenshot which i have 8 times this same server (because of 8 containers restart).

**Expected behavior**

I expect the server config to be updated on new import or at least to not be imported multiple time.

**Screenshots**

![pgadmin-multi-servers-import](https://github.com/user-attachments/assets/3beaf7fa-f438-495e-b0c3-d9bcd1569024)

**Additional context**

I use **oauth2** for authentification (AUTHENTICATION_SOURCES) and **postgresql** for config database (CONFIG_DATABASE_URI).

yogeshmahajan-1903:
Working fine.

Issue #5869, opened 2023-02-16
PGADMIN_CONFIG_CONFIG_DATABASE_URI from docker secrets


Mimikoo:
Hello.

There is PGADMIN_DEFAULT_PASSWORD_FILE which can be used to pass default password from docker secrets.

There is no way to pass default password for PGADMIN_CONFIG_CONFIG_DATABASE_URI from docker secrets. So the docker inspect shows the password for connecting to postgresql.

Is there an option to set DATABASE_URI from docket secrets and not use config_local.py ?

Mimikoo:
Found a workaround, but is not good for regular use.

> PGADMIN_CONFIG_CONFIG_DATABASE_URI: str(__import__('pathlib').Path("/run/secrets/pgadmin_config_config_database_uri").read_text())

m4dm4rtig4n:
Hello i try to set PGADMIN_CONFIG_CONFIG_DATABASE_URI  in Helm Release, ex :
```
env:
  - name: PGADMIN_CONFIG_CONFIG_DATABASE_URI
     value: postgresql://postgres:${password}@${name}-db-postgresql-hl.${namespace}:5432/postgres
```
But i have this error when PGadmin boot :/
```
SyntaxError: invalid syntax                                                                                                                                                                                                                  
postfix/postlog: starting the Postfix mail system                                                                                                                                                                                            
Traceback (most recent call last):                                                                                                                                                                                                           
  File "<string>", line 1, in <module>                                                                                                                                                                                                       
  File "/pgadmin4/config.py", line 906, in <module>                                                                                                                                                                                          
    from pgadmin.evaluate_config import evaluate_and_patch_config                                                                                                                                                                            
  File "/pgadmin4/pgadmin/evaluate_config.py", line 49, in <module>                                                                                                                                                                          
    import config_distro                                                                                                                                                                                                                     
  File "/pgadmin4/config_distro.py", line 12                                                                                                                                                                                                 
    CONFIG_DATABASE_URI = postgresql://postgres:XXXXXXX@pgadmin-db-postgresql-hl.pgadmin:5432/postgres
```
Have you idea how to fix this ? :/

PaulChristophel:
[...]

PR #10009, opened 2026-06-07
docker: read PGADMIN_CONFIG_CONFIG_DATABASE_URI safely via os.environ (fixes #9984)

### LINKED ISSUE #9984: incorrect use of PGADMIN_CONFIG_CONFIG_DATABASE_URI in entrypoint.sh

**Describe the bug**

I have an error when starting pgadmin container (dpage/pgadmin:9.15.0).

`
Traceback (most recent call last):
File "<string>", line 1, in <module>
from check_external_config_db import check_external_config_db; val = check_external_config_db("'postgresql+psycopg://user:pass@localhost:5432/pgadmin'"); print(val)

File "/pgadmin4/pgadmin/utils/check_external_config_db.py", line 18, in check_external_config_db
engine = create_engine(database_uri)
File "<string>", line 2, in create_engine
File "/venv/lib/python3.14/site-packages/sqlalchemy/util/deprecations.py", line 281, in warned
return fn(*args, **kwargs) # type: ignore[no-any-return]
File "/venv/lib/python3.14/site-packages/sqlalchemy/engine/create.py", line 564, in create_engine
u = _url.make_url(url)
File "/venv/lib/python3.14/site-packages/sqlalchemy/engine/url.py", line 856, in make_url
return _parse_url(name_or_url)
File "/venv/lib/python3.14/site-packages/sqlalchemy/engine/url.py", line 922, in _parse_url
raise exc.ArgumentError(
"Could not parse SQLAlchemy URL from given URL string"
)
sqlalchemy.exc.ArgumentError: Could not parse SQLAlchemy URL from given URL string
`

The check for PGADMIN_CONFIG_CONFIG_DATABASE_URI in entrypoint.sh is incorrect. 
The line 153 put double quotes around environment variable but the var should be specified with single quotes for correct generation of config_distro.py.

`
external_config_db_exists=$(cd /pgadmin4/pgadmin/utils && $SU_EXEC /venv/bin/python3 -c "from check_external_config_db import check_external_config_db; val = check_external_config_db(\"${PGADMIN_CONFIG_CONFIG_DATABASE_URI}\"); print(val)")
`

 The error doesn't prevent the start of pgadmin.

**To Reproduce**

1. configure PGADMIN_CONFIG_CONFIG_DATABASE_URI
2. start container
3. check container log

**Expected behavior**

No error reported

asheshv:
Closes #9984.

## Problem

The container entrypoint constructed a Python `-c` argument by shell-substituting `${PGADMIN_CONFIG_CONFIG_DATABASE_URI}` inside a double-quoted Python string:

```bash
val = check_external_config_db("${PGADMIN_CONFIG_CONFIG_DATABASE_URI}")
```

That collided with the long-standing **`config_distro.py` convention**, which requires the env var's value to **be** a Python literal (so the entrypoint's `CONFIG_DATABASE_URI = $val` write produces valid Python). Users therefore set the env var to `'postgresql+psycopg://...'`. The entrypoint then re-wrapped that literal in another set of quotes, producing:

```python
check_external_config_db("'postgresql+psycopg://...'")
```

— a string with literal single quotes inside. SQLAlchemy can't parse it:

```
sqlalchemy.exc.ArgumentError: Could not parse SQLAlchemy URL from given URL string
```

The Python crash made `$(...)` capture an empty string, so the downstream check `[ "${external_config_db_exists}" = "False" ]` also failed — silently skipping the entire first-launch user-setup block. That is the **second symptom** on the issue: `PGADMIN_DEFAULT_EMAIL` and `PGADMIN_DEFAULT_PASSWORD` getting ignored.

## Regression history

Bisected to commit `1fe840fca` ("Allow to pass PGADMIN_CONFIG_CONFIG_DATABASE_URI from docker secrets", #5869, **Oct 2024**), which incidentally added `"…"` shell quoting around the expansion in a PR whose actual scope was docker-secrets support (`file_env` helper). The pre-regression form (`ed211a2bb`, #7811, Sep 2024) used a bare expansion and worked correctly with the convention.

Both env-var forms have been broken in shipped images since Oct 2024:

| Env var form | Symptom |
|---|---|
| `PGADMIN_CONFIG_CONFIG_DATABASE_URI="'postgresql+psycopg://…'"` (with quotes, documented) | SQLAlchemy `ArgumentError` + silent `PGADMIN_DEFAULT_EMAIL` skip |
| `PGADMIN_CONFIG_CONFIG_DATABASE_URI=postgresql+psycopg://…` (without quotes, naive) | `config_distro.py` `SyntaxError` on startup → container exits |

## Fix

Rather than restore the bare-expansion form (a 7-year Chesterton's fence that the next shellcheck-style cleanup would reintroduce), **remove the shell from the quoting question altogether**:

- Read the env var **inside Python** via `os.environ` — the shell no longer participates in Python-literal handling.
- Use `ast.literal_eval` to unwrap the legacy `'url'` / `"url"` form when present; on `ValueError`/`SyntaxError`, fall through to the raw value (so users who set the env var without quotes also work).
- Default `external_config_db_exists` stays `"False"` on any Python failure (only overwritten when the Python call produces non-empty stdout) — fixes the secondary `PGADMIN_DEFAULT_EMAIL`-ignored regression.

## Verification (manual)

| Form | Result |
|---|---|
| `'postgresql+psycopg://…'` (legacy single-quoted) | `ast.literal_eval` unwraps → clean URI ✓ |
| `"postgresql+psycopg://…"` (double-quoted) | `ast.literal_eval` unwraps → clean URI ✓ |
| `postgresql+psycopg://…` (raw, no quotes) | `ast.literal_eval` raises → fall through to raw ✓ |
| End-to-end against live Postgres 17, no `server` table | returns `False` (first-launch path runs) ✓ |
| End-to-end against live Postgres 17, with `server` table seeded | returns `True` (skip first-launch as designed for #7811) ✓ |
| Malformed env var | Python fails → `external_config_db_exists` stays `"False"` → first-launch path runs (no silent skip) ✓ |

## Notes

- **`ast.literal_eval` safety**: parses Python literals only; cannot execute arbitrary code. Available since Python 2.6 (2008); pgAdmin's Docker image uses `python:3-alpine` (currently 3.12+), so no version-availability concern.
- **No new dependencies**, no docs changes, no behavior change for users with already-working configs (no working configs exist on master — both forms fail).
- **1 file changed**, +20/-1.

## Test plan

- [x] Bash syntax check (`bash -n pkg/docker/entrypoint.sh`) clean.
- [x] All three env-var forms validated against a live Postgres 17 (Tests 1-5 above).
- [ ] Container test (recommended for maintainers): rebuild the docker image off this branch, set `PGADMIN_CONFIG_CONFIG_DATABASE_URI="'postgresql+psycopg://…'"` + `PGADMIN_DEFAULT_EMAIL` + `PGADMIN_DEFAULT_PASSWORD`, run a fresh container → verify (1) no `ArgumentError` in container logs, (2) default admin user is created against the external DB, (3) restart with `server` table populated → first-launch is skipped as in #7811.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Docker configuration database initialization by fixing potential issues with special characters in database connection strings.

### DIFF
--- a/pkg/docker/entrypoint.sh
+++ b/pkg/docker/entrypoint.sh
@@ -195,9 +195,28 @@ EOF
 fi
 
 # Check whether the external configuration database exists if it is being used.
+#
+# The URI is read inside Python via os.environ — the shell no longer
+# participates in Python-literal quoting. `ast.literal_eval` unwraps the
+# legacy `'url'` form that the config_distro.py generation convention
+# requires; raw values pass through unchanged. On any Python failure the
+# "False" default below is preserved so first-launch user setup still
+# runs (see #9984).
 external_config_db_exists="False"
 if [ -n "${PGADMIN_CONFIG_CONFIG_DATABASE_URI}" ]; then
-     external_config_db_exists=$(cd /pgadmin4/pgadmin/utils && $SU_EXEC /venv/bin/python3 -c "from check_external_config_db import check_external_config_db; val = check_external_config_db(\"${PGADMIN_CONFIG_CONFIG_DATABASE_URI}\"); print(val)")
+    result=$(cd /pgadmin4/pgadmin/utils && $SU_EXEC /venv/bin/python3 -c "
+import os, ast
+from check_external_config_db import check_external_config_db
+raw = os.environ['PGADMIN_CONFIG_CONFIG_DATABASE_URI']
+try:
+    uri = ast.literal_eval(raw)
+except (ValueError, SyntaxError):
+    uri = raw
+print(check_external_config_db(uri))
+" 2>/dev/null)
+    if [ -n "$result" ]; then
+        external_config_db_exists="$result"
+    fi
 fi
 
 # DRY of the code to load the PGADMIN_SERVER_JSON_FILE


hiteshjambhale:
Reviewed and tested. Looks good to me.

dpage:
Merged to `master` as c88c8f612 (squashed, with you as author). Thanks @asheshv, and @hiteshjambhale for testing. Added a 9.16 Bug fixes note for #9984.

Two follow-up notes (neither blocking, kept your code as-is):
- I left `2>/dev/null` in place. Since `$result` only captures stdout, you could drop it to keep Python tracebacks visible in the container log for genuine failures (malformed URI / unreachable DB) without changing the fail-open behaviour — your call.
- The deeper diagnostic-loss is a pre-existing bug in `check_external_config_db.py`: its `finally: connection.close()` raises `NameError` when `engine.connect()` fails (connection unbound), so an unreachable DB raises instead of returning `False`. I'm opening a separate PR to fix that.

PR #10011, opened 2026-06-07
Migrations: use resolve_fks=False so reflect() survives stale FKs (fixes #9976)

asheshv:
Refs #9976 (the database-migration crash).

## Problem

A user installing pgAdmin 9.15 fresh on a machine where an old `pgadmin4.db` lingers (e.g. from a 2019-era install in `%appdata%\pgAdmin`) hits a startup failure:

```
sqlalchemy.exc.NoSuchTableError: user_old
```

The traceback (in #9976) points at migration `aff1436e3c8c_.py:32`:

```python
meta = sa.MetaData()
meta.reflect(op.get_bind(), only=('server',))
```

SQLAlchemy's `MetaData.reflect()` defaults to `resolve_fks=True`, which **auto-follows** the reflected table's FOREIGN KEY references and reflects the targets too. On the reporter's old DB, `server.user_id` carries a stale FK to a long-removed `user_old` table. The auto-cascade trips on the orphan FK target and raises `NoSuchTableError`, aborting the migration. The GUI then shows the generic "Server could not be contacted" — confusing because the migration crash is the real failure.

## Fix

None of the existing pgAdmin migrations actually need the FK target tables — they each do an UPDATE/INSERT/SELECT against the explicitly-requested table only. Sweep all 14 `meta.reflect()` call sites across 12 migration files and pass `resolve_fks=False`. Same behavior for healthy DBs (the FK targets just don't get auto-reflected, which the migrations don't care about); broken-FK DBs now survive the reflect step.

| Migration | Sites |
|---|---|
| `09d53fca90c7_.py` | role, version |
| `1f0eddc8fc79_.py` | role |
| `44b9ce549393_.py` | server, sharedserver |
| `7fedf8531802_.py` | user |
| `ac2c2e27dc2d_.py` | user_macros (×2) |
| `add_tools_ai_perm_.py` | role |
| `aff1436e3c8c_.py` | **server (the reporter's hit site)** |
| `c465fee44968_.py` | user |
| `c62bcc14c3d6_.py` | role, setting |
| `d39482714a2e_.py` | server |
| `e6ed5dac37c2_.py` | user_preferences, module_preference, … |
| `f656e56dfdc8_.py` | server/sharedserver (via parameter) |

**Total**: 12 files, +19/-15, all 14 reflect call sites guarded.

## Verification (manual, empirical)

| Backend | Result |
|---|---|
[...]

Issue #9976, opened 2026-05-24
Fresh install gives Server could not be contacted because of old pgadmin4.db


cschot:
**Describe the bug**
After a fresh install it tool a long time for pgadmin to try to start. Than I got the infamous Server could not be contacted error.
I read this could be because of old files in the Appdata folder.
So I checked  %appdata%\pgAdmin. (Note that this is not the pgadmin**4** folder)
There was an old pgadmin4.db file from 2019.
When I deleted that, the app could be started.

**Error messages**
In the GUI:  Server could not be contacted

In the log file(whole file below):
ERROR	pgadmin:	Database migration failed
(....)
sqlalchemy.exc.NoSuchTableError: user_old

**Expected behavior**
I expect 4 changes:

1. A successful database migration for old databases
2. An error message in the GUI that makes sense
	- Server could not be contacted -> Database migration failed
3. Easier troubleshooting
	- A link in the fatal error gui to the logs
	- Report the database path in the logs
	- A link in the fata error gui to a troubleshooting webpage with common error fixes
4. Easier bug reporting
	- A button in the fatal error gui to create an github issue with 

**Desktop (please complete the following information):**
 - OS: Windows 11 26100.8246
 - pgAdmin version: 9.15
 - Mode: Desktop

**Additional context**

<details><summary>Log</summary>
<pre>
2026-05-24 16:45:32,290: ERROR	pgadmin:	Database migration failed
2026-05-24 16:45:32,313: ERROR	pgadmin:	Traceback (most recent call last):
  File "D:\Program Files\pgAdmin 4\web\pgadmin\__init__.py", line 390, in upgrade_db
    db_upgrade(app)
    ~~~~~~~~~~^^^^^
  File "D:\Program Files\pgAdmin 4\web\pgadmin\setup\db_upgrade.py", line 25, in db_upgrade
    flask_migrate.upgrade(migration_folder)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "D:\Program Files\pgAdmin 4\python\Lib\site-packages\flask_migrate\__init__.py", line 111, in wrapped
    f(*args, **kwargs)
    ~^^^^^^^^^^^^^^^^^
  File "D:\Program Files\pgAdmin 4\python\Lib\site-packages\flask_migrate\__init__.py", line 200, in upgrade
    command.upgrade(config, revision, sql=sql, tag=tag)
[...]

PR #10012, opened 2026-06-07
Logging: include config-DB path/URI in migration failure messages (refs #9976)

asheshv:
Refs #9976 (the "report the DB path in the logs" improvement).

## Problem

When the configuration-DB migration fails, the log currently reads:

```
ERROR pgadmin: Database migration failed
ERROR pgadmin: Traceback (most recent call last): ...
```

It doesn't say **which** database file pgAdmin was trying to migrate. The reporter on #9976 spent time hunting through `%appdata%` looking for a stale `pgadmin4.db` left over from a 2019 install. Knowing the path up front would have shortened that investigation considerably.

## Fix

Pure additive logging. Add a small helper `_describe_config_db()` that returns one of:

- **Default (SQLite) mode**:
  `SQLite config DB at: /path/to/pgadmin4.db`
- **External (`CONFIG_DATABASE_URI`) mode**:
  `external config DB: postgresql+psycopg://user:***@host/db`
  (password between `:` and `@` is redacted before logging — the reporter on a separate issue uses Docker secrets for this URI, so we should never leak the credential in logs)

Include that string in the three migration-failure log sites:
1. `backup_db_file()` retry catch
2. `upgrade_db()` catch
3. Missing-tables check

After the change, the same scenario logs as:

```
ERROR pgadmin: Database migration failed (SQLite config DB at: C:\Users\<name>\AppData\Roaming\pgAdmin\pgadmin4.db)
ERROR pgadmin: Traceback (most recent call last): ...
```

— which is exactly what the reporter spent time hunting for.

## Redaction behaviour

| Input URI | Output |
|---|---|
| `postgresql+psycopg://user:edb@127.0.0.1:5432/pgadmin` | `postgresql+psycopg://user:***@127.0.0.1:5432/pgadmin` |
| `postgresql://user:p%40ss@host/db` (URL-encoded `@`) | `postgresql://user:***@host/db` |
| `postgresql://user:secret@host:5432/db?sslmode=require` | `postgresql://user:***@host:5432/db?sslmode=require` |
| `postgresql+psycopg://host/db` (no creds) | `postgresql+psycopg://host/db` (unchanged) |
| `postgresql://user@host/db` (no password) | `postgresql://user@host/db` (unchanged) |

The redaction is **deliberately conservative** — only the password between `:` and `@` after the scheme. Query-string secrets (some operators put e.g. `password=xxx` in URI parameters) are not redacted; that's a wider concern out of scope for this small logging improvement.

## Scope notes

- **Companion to #10011**, which fixes the actual migration crash on #9976. This PR addresses the *"easier troubleshooting"* item of the reporter's wishlist; #10011 addresses the *"successful database migration"* item. The two remaining items (clearer GUI error, GitHub-issue button) are bigger UX changes that warrant separate, more deliberate work.
- **No behavior change.** Pure additive logging.

## Test plan

- [x] `pycodestyle` clean.
- [x] Redactor unit-tested against the 5 URI shapes in the table above; all pass.
- [ ] **Recommended for maintainers**: force a migration failure (e.g. seed an orphan FK target into `pgadmin4.db` and start pgAdmin) → check the log lines now include the SQLite path.
- [ ] Same with `CONFIG_DATABASE_URI` set → check the log includes the redacted URI form.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages during database migration failures with clearer context about which configuration database is in use.
  * Enhanced logging to redact sensitive credentials and provide more detailed troubleshooting information when database operations encounter issues.

### DIFF
--- a/web/pgadmin/__init__.py
+++ b/web/pgadmin/__init__.py
@@ -362,6 +362,24 @@ def create_app(app_name=None):
     from config import SQLITE_PATH
     from pgadmin.setup import db_upgrade
 
+    def _describe_config_db():
+        """
+        Return a human-readable identifier of the configuration database
+        currently in use, for inclusion in error logs. Passwords in an
+        external CONFIG_DATABASE_URI are redacted.
+        """
+        if config.CONFIG_DATABASE_URI is not None and \
+                len(config.CONFIG_DATABASE_URI) > 0:
+            # Redact any password in the URI before logging.
+            import re
+            safe_uri = re.sub(
+                r'(://[^:/@]+:)([^@]+)(@)',
+                r'\1***\3',
+                config.CONFIG_DATABASE_URI
+            )
+            return 'external config DB: {0}'.format(safe_uri)
+        return 'SQLite config DB at: {0}'.format(SQLITE_PATH)
+
     def backup_db_file():
         """
         Create a backup of the current database file
@@ -378,7 +396,8 @@ def create_app(app_name=None):
                 'CORRUPTED_DB_BACKUP_FILE'] = backup_file_name
             app.logger.info('Database migration completed.')
         except Exception:
-            app.logger.error('Database migration failed')
+            app.logger.error(
+                'Database migration failed (%s)', _describe_config_db())
             app.logger.error(traceback.format_exc())
             raise RuntimeError('Migration failed')
 
@@ -390,7 +409,8 @@ def create_app(app_name=None):
             db_upgrade(app)
             os.environ['CORRUPTED_DB_BACKUP_FILE'] = ''
         except Exception:
-            app.logger.error('Database migration failed')
+            app.logger.error(
+                'Database migration failed (%s)', _describe_config_db())
             app.logger.error(traceback.format_exc())
             backup_db_file()
 
@@ -398,8 +418,8 @@ def create_app(app_name=None):
         is_db_error, invalid_tb_names = check_db_tables()
         if is_db_error:
             app.logger.error(
-                'Table(s) {0} are missing in the'
-                ' database'.format(invalid_tb_names))
+                'Table(s) {0} are missing in the database ({1})'.format(
+                    invalid_tb_names, _describe_config_db()))
             backup_db_file()
 
     def run_migration_for_sqlite():


hiteshjambhale (approved):

hiteshjambhale:
Hi @asheshv , this test-plan item is failing:

Same with CONFIG_DATABASE_URI set → check the log includes the redacted URI form.

When CONFIG_DATABASE_URI is set and the migration fails, no external config DB: line appears in the log.

dpage (changes requested):
Thanks for this — the SQLite half does exactly what #9976 needed (surfacing the `pgadmin4.db` path). One blocking gap, which confirms what @hiteshjambhale found in testing:

### The `external config DB:` branch is unreachable

Migration dispatch is:

```python
if config.CONFIG_DATABASE_URI ...:   # external
    run_migration_for_others()
else:                                 # SQLite
    run_migration_for_sqlite()
```

The three instrumented log sites all live in `backup_db_file()` / `upgrade_db()`, which are **only called from `run_migration_for_sqlite()`**. `run_migration_for_others()` calls `db_upgrade(app)` directly with no `try/except` and no failure logging. So:

1. When `CONFIG_DATABASE_URI` is set and migration fails, none of the new messages are emitted (the failure just propagates as a raw traceback) — hence @hiteshjambhale seeing no `external config DB:` line.
2. Because `_describe_config_db()` is only ever invoked from the SQLite path (where `CONFIG_DATABASE_URI` is empty by definition), its `external` branch — and the whole password-redaction block — is currently **dead code**.

Suggested fix — wrap the upgrade in `run_migration_for_others()` (the helper is already in scope):

```python
def run_migration_for_others():
    try:
        if get_version() == -1:
            db_upgrade(app)
        else:
            schema_version = get_version()
            if CURRENT_SCHEMA_VERSION > schema_version:
                db_upgrade(app)
                set_version(CURRENT_SCHEMA_VERSION)
                db.session.commit()
    except Exception:
        app.logger.error(
            'Database migration failed (%s)', _describe_config_db())
        app.logger.error(traceback.format_exc())
        raise
```

That makes both the external logging and the redaction actually live.

### Minor

- **CodeRabbit's query-param redaction point** is reasonable hardening, but it only matters *after* the branch above is reachable — at which point I'd either adopt the extra `re.sub` for `password=`/`token=` params or leave a comment that query-string secrets are intentionally not redacted. The core `user:password@` regex is correct (checked against all five URI shapes in the description).
- The description mentions redactor unit tests, but none are in the diff (only `__init__.py` changed) — worth adding one for `_describe_config_db()` once the external branch is live.
- Needs a `release_notes_9_16.rst` entry (refs #9976); the branch also predates the 9.16 scaffold, so a rebase will be needed.

PR #10015, opened 2026-06-07
fix(sqleditor): skip read-only columns when saving inserted rows

### LINKED ISSUE #9939: Query Editor Cannot Recognize Non-Updatable Fields
I reviewed the Query Editor documentation, and it clearly states that fields that are concatenations or that are renamed -- will not be updated. But when I write a simple query, such as:

SELECT id, first_name, last_name, first_name || ' ' || last_name as the_name
  FROM some_table

And then I insert a record, then Save Changes (F5) -- it gives me an error saying that field "the_name" does not exist in relation "some_table".

I get the same error when I try to do the following:

SELECT ST.id, ST.first_name, ST.last_name, (SELECT PT.person_type FROM person_type PT WHERE PT.id = ST.person_type_id) as the_type
  FROM some_table ST

When I try to save a new record, it tells me that field "the_type" does not exist in relation "some_table". No kidding!

So I tried ' ' || (SELECT PT.person_type FROM person_type PT WHERE PT.id = ST.person_type_id) as the_type

But this does not work either. The Query Editor still shows that column with a pencil icon instead of a lock icon.

Can you please fix the code so that it recognizes when a field is not updatable?

asheshv:
## Summary

Fixes #9939.

Query Tool result sets can include expression/alias columns (e.g. `first_name || ' ' || last_name AS the_name`) that don't exist on the underlying table. The editability detection in `is_query_resultset_updatable.py` already marks them `is_editable=False`, so the column header correctly shows the lock icon. But when a user adds a new row, the frontend populated **every** column key on the new row object and shipped the whole thing to `/sqleditor/save`. `save_changed_data` accepted those keys verbatim, and the rendered INSERT referenced the alias — Postgres replied with `column "the_name" does not exist in relation "some_table"`.

Filter on both sides:

- **Backend** — `get_columns_types` now propagates `is_editable` into the column metadata stored on the session (`is_editable=True` defaulted for View Data Tool, where columns are always real table columns).
- **Backend** — `save_changed_data` drops non-editable keys from the added-row `data` before rendering `insert.sql`. Defense-in-depth: protects against stale/buggy clients.
- **Frontend** — `ResultSet.jsx::triggerSaveData` omits cells whose column has `can_edit === false` from the added-row payload.

## Test plan

- [x] `python regression/runtests.py --pkg tools.sqleditor.utils.tests.test_save_changed_data` — 13/13 pass, including the new `TestSaveAddedRowSkipsNonEditableColumn` scenario that runs the issue's exact `first_name || ' ' || last_name AS the_name` SELECT and verifies the INSERT succeeds.
- [x] `python regression/runtests.py --pkg tools.sqleditor.utils.tests.test_is_query_resultset_updatable` — 10/10 pass (3 OID-only scenarios skipped on PG12+, as before).
- [x] `python regression/runtests.py --pkg tools.sqleditor.utils.tests.test_save_changed_uuid_data` — 3/3 pass (the other consumer of the changed helpers).
- [x] `make check-pep8`, `yarn run linter` clean.
- [ ] Manual smoke in Query Tool: run the issue's queries, add a row, F5 — INSERT should succeed and only write real columns.

## Release Notes

* **Bug Fixes**
  * SQL Editor now correctly saves newly-added rows when query results include non-editable columns (such as expression or alias columns), preventing save failures and invalid INSERT statements.

### DIFF
--- a/web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
@@ -1257,8 +1257,17 @@ export function ResultSet() {
     try {
       /* Convert the added info to actual rows */
       let added = {...dataChangeStore.added};
+      /* Strip read-only columns (expression/alias columns in Query Tool)
+       * from the new-row payload so the backend doesn't try to INSERT
+       * them into the base table. Issue #9939. */
+      let nonEditableKeys = new Set(
+        columns.filter((c)=>c.can_edit === false).map((c)=>c.key)
+      );
       Object.keys(added).forEach((clientPK)=>{
-        added[clientPK].data = _.find(rows, (r)=>rowKeyGetter(r)==clientPK);
+        let rowData = _.find(rows, (r)=>rowKeyGetter(r)==clientPK);
+        added[clientPK].data = _.omitBy(
+          rowData, (_v, k)=>nonEditableKeys.has(k)
+        );
       });
       let {data: respData} = await rsu.current.saveData({
         updated: dataChangeStore.updated,
--- a/web/pgadmin/tools/sqleditor/utils/get_column_types.py
+++ b/web/pgadmin/tools/sqleditor/utils/get_column_types.py
@@ -50,6 +50,11 @@ def get_columns_types(is_query_tool, columns_info, table_oid, conn, has_oids,
         col_type['type_name'] = None
         col_type['internal_size'] = col['internal_size']
         col_type['display_size'] = col['display_size']
+        # Propagate per-column editability so the save path can refuse to
+        # write expression/alias columns that don't exist in the base table.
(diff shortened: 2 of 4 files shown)


dpage:
Merged to `master` as f9b9bc9d4 (squashed, preserving @asheshv as author). I folded in @coderabbitai's suggestion — the added-row filter now requires `k in columns_info` so unknown client keys are dropped rather than defaulted to editable — and added a release-notes entry for #9939. The squash rewrites the SHA so GitHub won't auto-close this — closing manually. Thanks @asheshv!

hiteshjambhale:
Hi @asheshv 
This is fixed for add rows.
1. Update rows still fails with same error, We can implement this for update rows as well.
2. I would suggest resolving this on frontend. The feedback on adding a new row with non editable column is wrong. It initially appeared that the wrong value was inserted(null or added value), and it stays until we re-execute the select query

Issue #9900, opened 2026-05-01
Local file inclusion and SSRF in LLM API configuration endpoints (CVE-2026-7817)


asheshv:
## Summary

Local file inclusion (LFI) and server-side request forgery (SSRF) vulnerabilities in pgAdmin 4 LLM API configuration endpoints.

User-supplied `api_key_file` and `api_url` preferences were passed to the LLM provider clients without validation. An authenticated user could read arbitrary server-side files by pointing `api_key_file` at any path readable by the pgAdmin process, or coerce pgAdmin into making requests to internal targets (for example, cloud metadata services such as `169.254.169.254`) by setting `api_url`, exploiting the chat path and the model-list endpoints.

The fix restricts `api_key_file` to the user's private storage (server mode) or home directory (desktop mode), enforces a printable-ASCII key shape and a 1024-byte read cap, and gates `api_url` against a configurable allow-list (`config.ALLOWED_LLM_API_URLS`) at every entry point.

## Affected versions

pgAdmin 4 **>= 9.13, < 9.15**

## Fix

Fixed in pgAdmin 4 **9.15** by commit [`24485fe96`](https://github.com/pgadmin-org/pgadmin4/commit/24485fe96).

## Severity

- CVSS 3.1: **6.5 (MEDIUM)** — `AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:N/A:N`
- CVSS 4.0: **7.1 (HIGH)** — `AV:N/AC:L/AT:N/PR:L/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N`

## Credit

Reported by **j3seer**.

## References

- CVE: [CVE-2026-7817](https://www.cve.org/CVERecord?id=CVE-2026-7817)
- Release notes: [pgAdmin 4 9.15](https://github.com/pgadmin-org/pgadmin4/blob/REL-9_15/docs/en_US/release_notes_9_15.rst)

PR #10016, opened 2026-06-07
fix(llm): expand allowlist for self-hosted endpoints and stop silent fallbacks

### LINKED ISSUE #9936: Regression: pgAdmin 4.9.15 breaks custom LLM endpoints (LiteLLM, self‑hosted models, proxies)
**Version:** pgAdmin 4.9.15
**Related commit:** [https://github.com/pgadmin-org/pgadmin4/commit/24485fe9649860dcb632bb8029d4fa8efbafbc04](https://github.com/pgadmin-org/pgadmin4/commit/24485fe9649860dcb632bb8029d4fa8efbafbc04?utm_source=copilot.com)  
**Related issue:** [https://github.com/pgadmin-org/pgadmin4/issues/9900](https://github.com/pgadmin-org/pgadmin4/issues/9900?utm_source=copilot.com)

### Summary

The fix for #9900 introduced a hardcoded allow‑list of LLM provider URLs, which breaks all custom OpenAI‑compatible endpoints, including:

- LiteLLM

This is a regression from 4.9.14, where custom endpoints worked correctly.

It only allows these hardcoded urls: https://github.com/pgadmin-org/pgadmin4/blob/master/web/config.py#L1055-L1066:

```python
# Allowed LLM API URLs
# A list of scheme://host:port entries that LLM API requests are allowed
# to connect to. Only URLs matching an entry in this list will be permitted.
# This prevents SSRF attacks via user-controlled API URL fields.
# Set to an empty list to disable URL restriction (not recommended).
# Add entries for custom providers (LiteLLM, LM Studio, corporate proxies).
ALLOWED_LLM_API_URLS = [
    'https://api.anthropic.com:443',
    'https://api.openai.com:443',
    'http://localhost:11434',     # Ollama default
    'http://localhost:12434',     # Docker Model Runner default
]
```

**To Reproduce**

1. Go to AI>AI Configuration
2. Select OpenAI as Default Provider
3. Go to AI>OpenAI
4. API Key File: Configure the path of your API key file
5. API URL: Configure your custom LiteLLM URL
6. Model: Configure your preferred Model

**Expected behavior**

The AI Assistant should work without errors as it was working before on PgAdmin 4.9.14

**Error message**

It defaults to OpenAI

**Additional context**

CC @dpage

asheshv:
## Summary

Fixes #9936.

Three problems from #9900's allowlist fix compounded into the regression the reporter saw:

1. **Default allowlist was too narrow.** Only four hardcoded URLs were permitted; LiteLLM, vLLM, LM Studio, self-hosted llama.cpp, and corporate OpenAI-compatible proxies all required the admin to edit `config_local.py`.
2. **Silent URL substitution.** When the user's API URL preference was rejected by `validate_api_url()`, `get_*_api_url()` silently fell back to the trusted admin default. The user's request went through — but to a different provider, with the user's key. The reporter's symptom (`It defaults to OpenAI`) was this.
3. **Same silent-substitution for key file paths.** `get_*_api_key()` had the same shape: a preference path outside the user's private storage was dropped without any user-visible reason, leaving the AI panel asking the user to "configure the LLM first." (Comment by `jbro90` on the same issue.)

## Changes

- **`validate_api_url()`** learns a `:*` port wildcard. An entry like `http://localhost:*` matches every port on that host without opening up other hosts. Link-local cloud metadata (`169.254.169.254` etc.) remains blocked.
- **Default `ALLOWED_LLM_API_URLS`** now includes `http://localhost:*`, `http://127.0.0.1:*`, `http://[::1]:*` — covering Ollama (11434), Docker Model Runner (12434), LiteLLM (4000), vLLM (8000), LM Studio (1234), text-generation-webui (5000), and any self-hosted port out of the box.
- **No more silent substitution.** Two new helpers `_resolve_pref_url()` / `_resolve_pref_key_file()` collapse the four URL getters and two key-file getters to one-liners. When a user-set preference fails its allowlist check, they log a warning and return `''`/`None` instead of substituting the admin default. The admin fallback only happens when the user has not set a preference.
- **Clear errors in `get_llm_client()`.** New `is_pref_api_url_rejected()` / `is_pref_api_key_path_rejected()` helpers let the client raise an `LLMClientError` pointing the user at `ALLOWED_LLM_API_URLS` (in `config_local.py`) or at private user storage, instead of the misleading `API key not configured`.

## Scope notes

- jbro90's `[Comment](https://github.com/pgadmin-org/pgadmin4/issues/9936#issuecomment-4641965317)` mentioned a second symptom: after restart, the preferences page shows the old key-file path. I traced the save endpoint (`web/pgadmin/preferences/__init__.py`) and the preferences framework — there's no path-validation hook on save, the raw text is persisted, and the UI reads back the stored value. I couldn't reproduce that part. Will follow up on the issue asking for logs.

## Test plan

- [x] `python regression/runtests.py --pkg llm.tests.test_api_url_validation` — 49/49 pass, including 7 new wildcard scenarios, the rewritten resolution contract, both `is_pref_*_rejected` helpers, and the client-side error surfacing for both rejection paths.
- [x] `python regression/runtests.py --pkg llm` — 104/104 pass (no other LLM test regressed).
- [x] `pycodestyle --config=.pycodestyle` clean on all four changed files.
- [ ] Manual smoke in the AI panel:
  - [ ] Set the OpenAI API URL preference to `http://localhost:4000/v1` (LiteLLM) — refresh-models and chat both work without editing `config_local.py`.
  - [ ] Set the OpenAI API URL to `https://litellm.example.com` — refresh-models and chat both fail with the clear "not in the allowed list" error (no silent OpenAI fallback).
  - [ ] Set a key file path outside private storage — chat fails with the clear "not within your private user storage" error.

## Release Notes

* **New Features**
  * Support for self-hosted LLM endpoints on arbitrary localhost ports instead of fixed port numbers

* **Bug Fixes**
  * Enhanced API URL and key file validation with clearer error messages when configurations are rejected or invalid

* **Tests**
  * Expanded test coverage for LLM API URL allowlist behavior and key file path validation across multiple localhost forms

### DIFF
--- a/web/config.py
+++ b/web/config.py
@@ -1058,11 +1058,20 @@ DOCKER_API_MODEL = ''
 # This prevents SSRF attacks via user-controlled API URL fields.
 # Set to an empty list to disable URL restriction (not recommended).
 # Add entries for custom providers (LiteLLM, LM Studio, corporate proxies).
+#
+# Use ``:*`` to match any port on a host — useful for local self-hosting
+# where the user chooses the port. The host check still rejects
+# link-local addresses like 169.254.169.254 used by cloud metadata.
 ALLOWED_LLM_API_URLS = [
     'https://api.anthropic.com:443',
     'https://api.openai.com:443',
-    'http://localhost:11434',     # Ollama default
-    'http://localhost:12434',     # Docker Model Runner default
+    # Loopback addresses on any port: covers Ollama (11434), Docker
+    # Model Runner (12434), LiteLLM (4000), vLLM (8000), LM Studio
+    # (1234), text-generation-webui (5000), and any self-hosted
+    # OpenAI-compatible endpoint the user runs on their own machine.
+    'http://localhost:*',
+    'http://127.0.0.1:*',
+    'http://[::1]:*',
 ]
 
 # Maximum Tool Iterations
--- a/web/pgadmin/llm/client.py
+++ b/web/pgadmin/llm/client.py
@@ -172,7 +172,9 @@ def get_llm_client(
         get_anthropic_api_url, get_anthropic_api_key, get_anthropic_model,
         get_openai_api_url, get_openai_api_key, get_openai_model,
         get_ollama_api_url, get_ollama_model,
-        get_docker_api_url, get_docker_model
+        get_docker_api_url, get_docker_model,
+        is_pref_api_url_rejected,
+        is_pref_api_key_path_rejected,
     )
 
     # Determine which provider to use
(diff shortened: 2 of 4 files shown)


dpage:
Merged to `master` as 6bc434138 (squashed). I added a release-notes entry for #9936 and a short note in `config.py` flagging that the loopback `:*` defaults let any logged-in user reach the server's own loopback ports — so multi-user SERVER_MODE admins may want to pin specific ports. The squash rewrites the SHA so GitHub won't auto-close this — closing manually. Thanks @asheshv!

PR #10021, opened 2026-06-07
fix(schema-diff): aggregate child counts at parent group rows

### LINKED ISSUE #9892: Incorrect Aggregation in Schema Diff Summary
Summary
The "Schema Objects" parent node in the results tree displays "Different: 0", even though multiple child categories (Functions, Procedures, Sequences, Tables, and Types) contain objects that exist only in the Source or Target.

Description
In the Schema Diff tool, the top-level group should reflect the status of its children. In the provided image:
•	Functions: Has 7 Source Only and 1 Target Only objects.
•	Tables: Has 6 Source Only objects.
•	...and so on.
The top-level "Schema Objects" row shows Different: 0, Source Only: [Blank], and Target Only: [Blank]. While technically the objects are "Source Only" rather than "Different" (modified), a standard UI expectation is that the parent node should either:
1.	Sum the counts of all differences/mismatches below it.
2.	Highlight that differences exist so the user doesn't have to expand every single folder to find them.

Steps to Reproduce
1.	Open pgAdmin 4.
2.	Navigate to Tools > Schema Diff.
3.	Select a Source database/schema and a Target database/schema that have mismatched objects (e.g., tables that exist in Source but not Target).
4.	Click Compare.
5.	Observe the "Schema Objects" summary row.

Expected Behavior
The Schema Objects row should aggregate the statistics of its children. For example:
•	Different: Should be the sum of all "Different" counts.
•	Source Only: Should display the total count of all objects existing only in the source (e.g., $7 + 1 + 3 + 6 + 1 = 18$).
•	Target Only: Should display the total count of all objects existing only in the target (e.g., $1$).

Actual Behavior
The parent node "Schema Objects" remains at 0 for all columns or leaves them blank, providing no visual indication of the mismatches contained within the sub-folders unless they are manually expanded.
________________________________________
Suggested Priority
•	Severity: (It is a UI/Reporting bug; the underlying DDL comparison likely still works, but it causes poor visibility).
•	Component: pgAdmin 4 - Schema Diff Plugin.

 Environment Information
• pgAdmin 4 Version: (9.14)
• Mode: Desktop / Server (whichever you are using)
• OS: All
• PostgreSQL Version: (e.g., postgresql-18.3-3-windows-x64)

Snapshot:

Image

asheshv:
## Summary

Fixes #9892.

In *Tools → Schema Diff*, the top-level group rows ("Schema Objects", "Database Objects") rendered with blank counts even when their child groups had non-trivial differences. A user could see "Schema Objects" at zero/blank while Functions, Tables, and others underneath listed Source-Only / Target-Only objects, with no signal at the parent level that anything was different.

## Root cause

The results tree is three levels: a top-level *object-group* row, mid-level *object-type* rows (Functions, Tables, Sequences…), and leaf *object* rows with a `status` (Identical / Different / Source Only / Target Only).

`ResultGridComponent.jsx::setRecordCount` only knew how to handle leaf children — it iterated `row.children` and bumped counts off each child's `status` field. Mid-level rows correctly counted their leaves, but the top-level rows, whose children are *mid-level rows* (no `status`, only their own counts), got zero.

`ResultGridComponent.jsx`'s render gate `'identicalCount' in row` skipped the count rendering entirely for top-level rows because `SchemaDiffCompare.jsx::generateGridData` never seeded those four fields on the top-level row object.

## Fix

- `ResultGridComponent.jsx`: `setRecordCount` now recognises a third child shape — a counted-group child (has `identicalCount` etc.) instead of a leaf with a `status`. For those children it recurses to refresh the child's counts (so `filterParams` changes propagate) and rolls the four totals up. Leaf-only behavior is unchanged. Exported the function so the regression test can drive it directly.
- `SchemaDiffCompare.jsx`: seed `identicalCount` / `differentCount` / `sourceOnlyCount` / `targetOnlyCount` to 0 on every top-level group row in `generateGridData`, so the render gate fires and counts get drawn.

## Test plan

Adds `regression/javascript/schema_diff/result_grid_count_spec.js` with five Jest scenarios driving the exported `setRecordCount` directly:

- Leaf children counted by status (pre-existing behavior preserved).
- Leaf children skipped when their `status` is outside `filterParams`.
- **Top-level aggregation (the #9892 case)** — builds the reporter's exact tree (Functions src=7/tgt=1, Tables src=6, Procedures src=3, Sequences src=1/diff=1, Types ident=1) and asserts the top-level totals are 17/1/1/1 instead of 0/0/0/0.
- Re-running with a narrower `filterParams` correctly refreshes the mid-level child counts (no stale numbers leak into the top-level total).
- Empty `children` list is a no-op.

- [x] `yarn run eslint` clean on the changed files.
- [x] `yarn run jest -t setRecordCount`: 5/5 new tests pass.
- [x] `yarn run test:js-once`: 829/829 pass (was 824/824).
- [ ] Manual: reproduce the reporter's scenario in *Tools → Schema Diff*, confirm the "Schema Objects" row now shows the aggregated counts (and that changing the status filter chips refreshes them at every level of the tree).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed schema diff comparison to accurately count and display identical, different, source-only, and target-only items. Counts now properly aggregate through nested group hierarchies and update correctly when filters are applied.

* **Tests**
  * Added comprehensive test coverage for schema diff count calculations.

### DIFF
--- a/web/pgadmin/tools/schema_diff/static/js/components/ResultGridComponent.jsx
+++ b/web/pgadmin/tools/schema_diff/static/js/components/ResultGridComponent.jsx
@@ -159,14 +159,38 @@ function useFocusRef(isSelected) {
   };
 }
 
-function setRecordCount(row, filterParams) {
+/*
+ * Recompute the four status counts for ``row`` based on its descendants.
+ *
+ * Three row shapes can appear in the schema-diff results tree:
+ *   - Leaf row:  has a ``status`` (Identical/Different/Source Only/Target
+ *                Only); contributes 1 to the matching count when it's
+ *                visible under ``filterParams``.
+ *   - Mid-level row (object-type group, e.g. "Functions"): children are
+ *     leaves. Counts are the sum of those leaves.
+ *   - Top-level row (object-group, e.g. "Schema Objects"): children are
+ *     mid-level rows. Counts are the sum of the mid-level rows' counts.
+ *
+ * Originally this function only handled the leaf-children case, which
+ * left top-level group rows showing blank counts even when their
+ * descendants had differences. Issue #9892.
+ */
+export function setRecordCount(row, filterParams) {
   row['identicalCount'] = 0;
   row['differentCount'] = 0;
   row['sourceOnlyCount'] = 0;
   row['targetOnlyCount'] = 0;
 
   row.children.map((ch) => {
-    if (filterParams.includes(ch.status)) {
+    if ('identicalCount' in ch) {
+      // Mid-level child: refresh its counts (in case filterParams
+      // changed since the last render) and roll them up.
+      setRecordCount(ch, filterParams);
+      row['identicalCount'] += ch['identicalCount'];
+      row['differentCount'] += ch['differentCount'];
+      row['sourceOnlyCount'] += ch['sourceOnlyCount'];
+      row['targetOnlyCount'] += ch['targetOnlyCount'];
+    } else if (filterParams.includes(ch.status)) {
       if (ch.status == FILTER_NAME.IDENTICAL) {
         row['identicalCount'] = row['identicalCount'] + 1;
       } else if (ch.status == FILTER_NAME.DIFFERENT) {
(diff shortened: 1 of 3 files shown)


dpage:
Merged to `master` as 833504f1b (squashed, with you as author).

Added a 9.16 **Bug fixes** release-note entry for #9892 as part of the merge (the branch predated the 9.16 notes scaffold). Verified locally: eslint clean on the changed files and the new `result_grid_count_spec.js` passes 5/5.

Thanks @asheshv!

PR #10038, opened 2026-06-09
Let clicks pass through the object breadcrumbs popup (#9128)

### LINKED ISSUE #9128: Bug with interaction of server list

**Describe the bug**

In pgAdmin in server list I have 2 server groups: **local/personal** and **work**, when I'm working I often connect to local DB first before connection to DB from work. When DB is expanded in a server list it add a scrollbar, so a prerequisite for this is having 2 server groups and first being expanded. Now as you may know when you hover over item in a list it shows a popup at bottom left which tells which item is being hovered, but in the case like this it is blocking the work group I trying to click as it pop over it preventing me from clicking on it. That's the first issue and **a solution** for that, since you using HTML, CSS, and JS would be adding CSS rule to that popup: `pointer-events: none;`

But due to this issue I find out another interaction bug: since I cannot select element from a mouse I naturally tried to select it with **keyboard arrows** and I was able to expand my work group with **arrow right**, but since I already used a keyboard, instead of clicking with mouse on arrow ">" to expand server and connect to it, I used keyboard right arrow button which is expand **marker**, but **doesn't actually initiate connection** to the DB.

**To Reproduce**

1. Have 2 server groups
2. Expand first with enough elements that eventually create scrollbar
3. Try to hover over 2nd group and click - you wont be able because popup will block you from clicking
4. Use keyboard arrows to navigate and arrow right to open 2nd server group
5. Try to use arrow right to connect to server - it wont

**Expected behavior**

Able to click on 2nd group and able to connect to DB if I use keyboard.

**Error message**

None

**Screenshots**

Work group below that popup (not visible):

Image

The server is expanded with keyboard, but no connection initiated:

Image

 - OS: Windows 10
 - pgAdmin version: 9.7
 - Mode: Desktop
 - Browser (if running in server mode): -
 - Package type: -

**Additional context**

-

dpage:
### Summary

Fixes #9128.

The object breadcrumbs popup is an absolutely-positioned overlay (`position: absolute; bottom: 0; zIndex: 1004`) at the bottom-left of the object explorer. Because it's purely informational, it shouldn't capture mouse input — but it did, blocking clicks on the tree items underneath it. Added `pointer-events: none` to the styled container so clicks pass through.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the object breadcrumbs popup could block interactions with items in the object explorer beneath it.


asheshv (approved):
LGTM. `pointer-events: none` on the root is the right layer — the component has no interactive children so there's no risk of breaking clicks/focus inside.

Non-blocking suggestion: a one-line `getComputedStyle(...).pointerEvents` assertion in `ObjectBreadcrumbs.spec.js` would prevent a silent regression if the property gets dropped in a future restyle.

PR #10039, opened 2026-06-09
Fix index/constraint column not shown when its name requires quoting (#6481)

### LINKED ISSUE #6481: The index column  is not shown in the Properties dialogue  if the column name contains double quotes *
The index column  is not shown in the Properties dialogue  if the column name contains double quotes

**To Reproduce**

1. create the following database
CREATE DATABASE """"
    WITH
    OWNER = enterprisedb
    ENCODING = 'UTF8'
    CONNECTION LIMIT = -1
    IS_TEMPLATE = False;
2. database created
3. Click on this newly created database and expand it
4. create the following table and index
CREATE TABLE public.""""
(
    """" "char"
);

CREATE INDEX
    ON public."""" USING btree
    ("""" ASC NULLS LAST)
;
5. Tables and index created
6. Expand the index and Right Click and Expand properties
7. Error message Column and Column not empty message is also shown.

**Expected behavior**

The index column should be displayed in case of special characters.

**Error message**

If possible, include the full, unedited text of any error messages displayed.

**Screenshots**
Screen Shot 2023-06-22 at 1 49 51 PM

 - OS: [All]
 - Version: [7.3]
 - Mode: [All]
 - Package type: [All]

**Additional context**

dpage:
### Summary

Fixes #6481.

When an index (or exclusion constraint) column name requires quoting — e.g. it contains a double quote or other special characters — the column was not shown in the Properties panel, and editing reported it as empty.

**Root cause:** the `is_exp` (expression) flag is computed as `pg_get_indexdef(...) = a.attname`. `pg_get_indexdef()` returns the **SQL-quoted** identifier (e.g. `"col""x"`) while `a.attname` is the **raw** name (`col"x`). For any name needing quoting these never match, so a plain column was misclassified as an expression and dropped.

**Fix:** compare against `pg_catalog.quote_ident(a.attname)`, which yields the same quoted form as `pg_get_indexdef()` for both normal and special names. Real expression indexes (where `attname` is NULL) still evaluate as expressions.

### Changes
- `indexes/sql/11_plus/column_details.sql`, `indexes/sql/default/column_details.sql`, `exclusion_constraint/sql/default/get_constraint_cols.sql`
- Release note (9.16)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of quoted database identifiers in indexes and constraints.
  - Preserved mixed-case names, reserved words, and identifiers containing embedded double quotes.
  - Improved distinction between indexed columns and expressions.
  - Corrected identifier escaping in table and constraint operations.

- **Tests**
  - Added coverage for quoted column names, identifier parsing, expression detection, and complex escaping scenarios.

### DIFF
--- a/docs/en_US/release_notes_9_16.rst
+++ b/docs/en_US/release_notes_9_16.rst
@@ -25,3 +25,5 @@ Housekeeping
 
 Bug fixes
 *********
+
+  | `Issue #6481 <https://github.com/pgadmin-org/pgadmin4/issues/6481>`_ -  Fixed an issue where an index or exclusion constraint column was not shown in the Properties panel (and reported as empty) when the column name required quoting (e.g. it contained a double quote or special characters).
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/constraints/exclusion_constraint/utils.py
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/constraints/exclusion_constraint/utils.py
@@ -11,6 +11,7 @@
 
 from flask import render_template
 from flask_babel import gettext as _
+from pgadmin.utils import unquote_ident
 from pgadmin.utils.ajax import internal_server_error
 from pgadmin.utils.exception import ObjectGone, ExecuteError
 from functools import wraps
@@ -74,7 +75,7 @@ def _get_columns(res):
             order = True
             nulls_order = True if (row['options'] & 2) else False
 
-        columns.append({"column": row['coldef'].strip('"'),
+        columns.append({"column": unquote_ident(row['coldef']),
                         "oper_class": row['opcname'],
                         "order": order,
                         "nulls_order": nulls_order,
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/constraints/foreign_key/utils.py
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/constraints/foreign_key/utils.py
@@ -11,6 +11,7 @@
 
 from flask import render_template
 from flask_babel import gettext as _
+from pgadmin.utils import unquote_ident
 from pgadmin.utils.ajax import internal_server_error
 from pgadmin.utils.exception import ObjectGone, ExecuteError
 from functools import wraps
@@ -126,7 +127,7 @@ def search_coveringindex(conn, tid, cols, template_path=None):
 
         index_cols = set()
         for r in rest['rows']:
-            index_cols.add(r['column'].strip('"'))
+            index_cols.add(unquote_ident(r['column']))
 
         if len(cols - index_cols) == len(index_cols - cols) == 0:
             return constraint["idxname"]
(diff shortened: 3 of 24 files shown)


asheshv (changes requested):
The SQL fix (comparing against `quote_ident(a.attname)` instead of `a.attname`) is correct — `is_exp` will now classify properly. But the Python side that *displays* the column name is still broken for the exact category the PR claims to fix.

`web/pgadmin/browser/server_groups/servers/databases/schemas/tables/indexes/utils.py:123` uses `attdef.strip('"')`. `str.strip('"')` strips outer quotes but doesn't unescape doubled inner quotes. For a column named `col"x`, `pg_get_indexdef` returns `"col""x"`; `.strip('"')` yields `col""x`, not `col"x`. So the properties panel still displays the wrong name for any identifier containing a literal `"`.

Pull this into a tiny helper:

```python
def unquote_ident(s):
    if s.startswith('"') and s.endswith('"'):
        return s[1:-1].replace('""', '"')
    return s
```

Tests gap: no scenarios for quoted-identifier column names (uppercase, spaces, reserved words, embedded `"`). The regression that caused #6481 wouldn't have been caught by existing tests, and this PR doesn't add coverage for it.

dpage:
Thanks for the thorough review, Ashesh.

The Python side is now fixed with an `unquote_ident()` helper in `pgadmin.utils` that reverses `quote_ident()` properly, unescaping doubled inner quotes rather than just stripping the outer ones. It's now shared by all five call sites that had the same `.strip('"')` pattern: indexes, index constraints, exclusion constraints and foreign keys.

Test coverage added:
- `web/pgadmin/utils/tests/test_unquote_ident.py` — unit tests for the helper directly (unquoted names, outer-quote stripping, doubled inner quotes, expressions like `"a" || "b"` left untouched, edge cases).
- `web/pgadmin/browser/.../indexes/tests/test_indexes_quoted_column.py` — new regression scenarios asserting the Properties panel reports the right column name and `is_exp` is `False` for a mixed-case name, a reserved word, and a name containing a literal `"`.

Also rebased onto current `upstream/master` to pick up the CI fixes. All indexes and constraints test packages pass locally (48 + 150 tests).

Ready for re-review.

dpage:
Fixed the `run-python-tests-pg` failures, which were the resql suite rather than anything at runtime.

Six scenarios covering a table whose column name contains a double quote (create table with a primary key, create table with a primary key and a check constraint, add and delete columns, add a unique constraint, delete constraints) had their expected SQL recorded before this fix, so it still carried the doubly escaped name the old `.strip('"')` produced. In those files the constraint refers to `"col1_...\""""'`\\/#"` while the column it is declared against, two lines above, is `"col1_...\""'`\\/#"`.

Now that identifiers are unquoted properly, the generated script agrees with itself, and agrees with what PostgreSQL's own `quote_ident()` returns for that name, so it was the recorded SQL that needed correcting rather than the code. Twelve files updated, six scenarios across both `pg` and `ppas`. `--pkg resql` and the full Python suite both pass locally.

dpage:
@coderabbitai On the outside-diff finding about the `Username`/`SharedUsername` fallback in `web/pgadmin/utils/__init__.py` (around line 671): that code belongs to the server import path and is untouched by this PR, which only adds `unquote_ident()` to the same file. It's out of scope here, so I'm leaving it for a separate change rather than folding it into an identifier-quoting fix.

PR #10041, opened 2026-06-09
Apply server foreground colour to object counts (#9766)

### LINKED ISSUE #9766: Foreground color not consistent

**Describe the bug**

Not all font colors follow the foreground color configuration. The counts (i.e., the number of tables, the number of columns on a table, etc) and the column type. 

**To Reproduce**

1. Register a server with green background and black foreground. 
2. Create a table with at least one column
3. See the error - the count next to "Tables" and "Columns" and the type of the column on the table that was just created.

**Expected behavior**

All font colors should follow the foreground color configuration

**Error message**

n/a

**Screenshots**

Image

 - OS: OSX Sonoma 14.4.1
 - pgAdmin version: 9.12
 - Mode: Desktop
 - Browser (if running in server mode): n/a

dpage:
### Summary

Fixes #9766.

A server's custom **foreground** colour was applied to node labels (`span.file-name`) but not to the object **counts** shown next to collection nodes, so the counts stayed the default colour and could be hard to read against the server background colour.

**Fix:** include `span.children-count` in the per-server `fgcolor` CSS rule in `change_server_background`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where custom server foreground colors were not applied consistently in the object explorer; object counts (children counts) and column type labels now honor the server's configured foreground color for consistent theming across the interface.


asheshv (approved):
LGTM, with one small follow-up. The selector and `color-mix()` usage match the existing convention in `reactaspen.override.js`.

One gap worth fixing in this PR or a follow-up: the new rule has no `:hover` / `.pseudo-active` variant for `span.children-count`, so on hover the count snaps to the theme `textHoverFg` while the label keeps the user's custom colour. Mirroring the `:hover` block already added for `file-name` would close that inconsistency.

PR #10045, opened 2026-06-09
Pin setuptools<82 on Python 3.9 to keep pkg_resources (#9829)

### LINKED ISSUE #9829: Installation error with Python3.9
**Describe the bug**

If I create a brand new venv in a system with Python3.9 installed (RHEL-based OS), I get the following error on the first pgadmin4 run:
ModuleNotFoundError: No module named 'pkg_resources'
If I manually downgrade setuptools<82.0.0 after pgadmin4 installation, I am able to run the software (even though an error is raised during the downgrade because the pgadmin4 requirements are not satisfied).

The latest version which I can run withouth issues is the 9.12

**To Reproduce**

To reproduce the issue, use a system with the following python3 version installed:
```
$ python3 --version
Python 3.9.25
```

And then execute the following
```
python3 -m venv venv
source venv/bin/activate
pip install pgadmin4
pgadmin4
```

**Expected behavior**

The server should run as expected without manually fixing the dependencies. Instead, an exception is raised

**Error message**
```
(venv) $ pgadmin4
Traceback (most recent call last):
  File "/home/xxx/venv/bin/pgadmin4", line 5, in <module>
    from pgadmin4.pgAdmin4 import main
  File "/home/xxx/venv/lib64/python3.9/site-packages/pgadmin4/pgAdmin4.py", line 38, in <module>
    import config
  File "/home/xxx/venv/lib/python3.9/site-packages/pgadmin4/config.py", line 33, in <module>
    from pgadmin.utils import env, IS_WIN, fs_short_path
  File "/home/xxx/venv/lib/python3.9/site-packages/pgadmin4/pgadmin/__init__.py", line 32, in <module>
    from flask_security import Security, SQLAlchemyUserDatastore, current_user
  File "/home/xxx/venv/lib64/python3.9/site-packages/flask_security/__init__.py", line 15, in <module>
    from .core import (
  File "/home/xxx/venv/lib64/python3.9/site-packages/flask_security/core.py", line 86, in <module>
    from .totp import Totp
  File "/home/xxx/venv/lib64/python3.9/site-packages/flask_security/totp.py", line 18, in <module>
    from passlib.pwd import genword
  File "/home/xxx/venv/lib64/python3.9/site-packages/passlib/pwd.py", line 16, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

 - OS: Oracle Linux 9
 - pgAdmin version: 9.14
 - Mode: Server
 - Browser (if running in server mode): N/A
 - Package type: Pip

dpage:
### Summary

Fixes #9829.

A clean install on **Python 3.9** failed with `No module named 'pkg_resources'`. `setuptools==82.*` no longer ships `pkg_resources`, which `passlib` (pulled in via `Flask-Security-Too` on Python 3.9) imports at runtime.

Python 3.9 is still a supported target — RHEL/Rocky/AlmaLinux **8 and 9** build the packaged venv with the system Python 3.9, and the codebase enforces 3.9 as the minimum (`web/pgAdmin4.py`).

**Fix:** version-split the `setuptools` pin (mirroring the existing `Flask-Security-Too` split) so Python 3.9 keeps a `setuptools` that still provides `pkg_resources`:

```
setuptools<82; python_version <= '3.9'
setuptools==82.*; python_version > '3.9'
```

## Summary by CodeRabbit

* **Documentation**
  * Updated v9.16 release notes with connection warning about password-exec vs passfile precedence and corrected PostgreSQL utilities versions to 18.4.
* **Bug Fixes**
  * Expanded documentation of resolved issues.


asheshv (approved):
Pin is correct — `passlib` 1.7.x still uses `pkg_resources` and we're locked to `Flask-Security-Too` 5.6 on Python 3.9.

Two small asks before merge:

1. Tighten the lower bound: `setuptools>=70,<82` (matches the convention used elsewhere in the file and prevents a fresh-venv resolver from picking up an ancient version).
2. Add an inline comment or an issue reference noting the real fix path — FST ≥5.7 + `libpass` once Python 3.9 support is dropped — otherwise this pin will age silently.

PR #10047, opened 2026-06-09
Validate preference values set via the CLI (#7346)

### LINKED ISSUE #7346: Preferences should be validated in CLI mode.
If I run the command python setup.py set-prefs aditya.toshniwal14@gmail.com "browser:processes:process_retain_days=true" then it is success. Here, process retain days is a number.

dpage:
### Summary

Fixes #7346.

Preferences set via the CLI (`setup.py set-prefs`) were not validated: `Preferences.save_cli` wrote the raw value straight to the configuration database and always returned success, so an invalid value (e.g. a non-numeric string for an integer preference) was stored silently. The GUI path validates via `_Preference.set()`.

**Fix:**
- `_Preference.set()` now accepts an optional `user_id`, so it can be used outside a request context (the CLI has no `current_user`).
- `Preferences.save_cli` now looks up the preference (like `Preferences.save`) and delegates to `set()`, so the value is validated against the preference type.
- `setup.py set-prefs` checks the result and reports any preference whose value was invalid instead of silently claiming success.

The GUI path is unaffected (`save_cli` is only used by the CLI; `update()` calls `pref.set()` directly).

### Changes
- `web/pgadmin/utils/preferences.py`, `web/setup.py`
- Release note (9.16)

## Bug Fixes
- CLI preference setting via setup.py now validates preference values against their defined types and rejects invalid values with specific error messages instead of storing them silently.
- Enhanced error reporting distinguishes between missing preferences and invalid values.


asheshv (changes requested):
The threading of `user_id` through `set()` and routing `save_cli` via `_Preference.set()` is the right approach, but three issues:

1. **Critical — boolean prefs via `--input-file` will universally fail.** `setup.py:719` builds the value list via `str(v)`. For native JSON booleans this produces `'True'` / `'False'` (capital T/F). `save_pref` only coerces lowercase `['true', 'false']`, so the value reaches `_Preference.set()` as the string `'True'` and trips `assert isinstance(value, bool)`. Either normalize in `save_pref` (`data['value'].lower() in ['true', 'false']`) or stop stringifying booleans at the `setup.py` boundary.
2. **Error message discarded.** `res, _ = Preferences.save_cli(...)` throws away the validator's diagnostic. Users see "invalid value" with no detail — the message should propagate up through `save_pref` and into `setup.py`'s `invalid_value_prefs` report.
3. **No new tests** for `save_pref` / `save_cli` with invalid integer, valid integer-as-string, boolean, unknown options value, or missing preference. The validation surface is currently untested.

Non-blocking observation (pre-existing): `setup.py:741` uses `opt.split("=")[1]`, which truncates any value containing `=` (URLs, regexes, base64). Worth a follow-up issue.

PR #10054, opened 2026-06-09
Rebase the version-specific SQL templates to target PostgreSQL 14

### LINKED ISSUE #10050: Rebase the SQL templates
The base (default) SQL templates currently target PostgreSQL < 12. We need to re-base the templates so the default target is 14 (the current oldest supported server version), with overrides only for newer versions.

dpage:
## Summary

The base (`default`) SQL templates historically targeted PostgreSQL < 12. This re-bases the version-bucketed templates so the **default target is 14** — the oldest supported server version — keeping overrides only for newer versions.

For every "bucket container" (a directory holding `default` / `<N>_plus` subdirs), every bucket with version number ≤ 14 (`11_plus`, `12_plus`, `13_plus`, `14_plus` and the old `default`) is collapsed into a single `default`, choosing — per file — the content that a server running exactly v14 resolves today. Buckets for newer versions (`15_plus`, `16_plus`, `17_plus`, `18_plus`) are retained untouched as overrides.

- **118** obsolete sub-14 bucket directories removed; a large amount of cross-version template duplication eliminated (net ~-47k lines).
- Drops PostgreSQL / EDB Advanced Server **13** from the 9.16 supported-server list.
- Repoints the `sqleditor` explain_plan tests (which referenced the now-removed `12_plus`/`13_plus` buckets) at the new `default` template.
- The `VersionedTemplateLoader` and its generic resolution algorithm are unchanged.

## Behaviour preservation

The transform is **behaviour-preserving for every server version ≥ 14**. Template (and test-fixture) resolution was verified to be **byte-identical before and after** for all supported versions, across every bucket container, by simulating the loader's resolution at each version boundary. A server < 14 (now unsupported) resolves the new `default` instead of a dropped bucket — the intended effect.

## Test plan

- [x] `utils.tests.test_versioned_template_loader` — 6 passed
- [x] `tools.sqleditor.tests.test_explain_plan_templates` — 4 passed
- [x] `resql` reverse-engineered SQL suite vs PostgreSQL 18 — 841 scenarios passed
- [x] `tools.erd` — 9 passed; `tools.schema_diff` — 2 passed
- [x] pep8 clean on changed Python

Closes #10050


asheshv (approved):
Sampled template merges (publications, functions, aggregates, collations, tables, `explain_plan`) are semantically correct and the loader's fallback behavior covers the deleted directories cleanly. Two cleanups worth folding in:

1. `get_version_mapping_directories()` in `versioned_template_loader.py` still lists `11_plus` / `12_plus` / `13_plus` — those entries are dead now and a future contributor adding a `12_plus/` dir back wouldn't get any error. Worth removing.
2. `databases/schemas/.../directories/__init__.py` still has `min_ver = 130000` — should be bumped to `140000` to match the new baseline; otherwise the EPAS Directories node will render on PPAS 13 against PG14-only SQL.

Also — the description mentions byte-identical verification against PG18, which is exactly the path that doesn't exercise the new `default/`. A resql / properties pass on a live PG14 server before merging would catch any sneaky 12/13-only syntax that slipped into a merged template.

PR #10055, opened 2026-06-09
Fix query tool auto execution after session restore

### LINKED ISSUE #10031: Open queries autoruns when I start the app

**Describe the bug**

When I have opened queries from the last session, they automatically gets executed on the next startup of the app.

**To Reproduce**

1. Start the app and connect to a server.
2. Open a query window and write a SELECT that gets data from an existing table.
3. Close the app with the query window open.
4. Restart the app.
5. The previous query window is opened again and executed.

**Expected behavior**

The query should **not** be automatically executed. Imagine if I have a DELETE query saved and that runs.

**Error message**

If possible, include the full, unedited text of any error messages displayed.

**Screenshots**

 - OS:  Windows-11-10.0.26200-SP0
 - pgAdmin version: 9.15
 - Mode: Desktop
 - Package type: [e.g. RPM, DEB, Python, Container, etc.]

**Additional context**

dev-hari-prasad:
## What this fixes

Fixes an issue where Query Tool tabs are restored from a previous session can/start automatically executing their queries when pgAdmin starts.

## What was happening

While looking into this, I found that `is_query_tool` was only being checked against the string value 'true'`.

During the session restoration process the value can be restored as a boolean instead. When that happens, the tab gets treated as a view/edit data tab instead of a Query toool tab and can end up going through the auto execution logic during initialization. Which can lead to descurtive and unwanted queries to run against the database and this is not what a user would expect.

### This change addresses the restore flow in a few places:
- Handle both boolean and string values for `is_query_tool` 
- Prevent automatic execution while a Query Tool tab is being restored when the app starts
- Make sure the Query Tool tabs persist the correct `is_query_tool` state in the saved metadata

## Testing

- Open Query Tool
- Write a query and leave the tab open
- Close pgAdmin
- Start pgAdmin again
- Verify the query text is restored
- Verify the query is not executed automatically

PS: You will need to build the app with the changes made in this branch inorder to see this in effect otherwise the offical build will emit the same bevhiour as descirbied in the issue #10031.

Fixes #10031 

## Summary by CodeRabbit

* **Bug Fixes**
  * Query Tool now correctly recognizes the enabled state when the parameter is `true` or the string `"true"`.
  * Improved restoration behavior to prevent unintended SQL script restore and avoid local restore when appropriate.
  * Automatic execution is blocked during restore to prevent accidental runs.
  * Restoring the tool reliably re-enables the editor and clears the restore flag.
  * When activating the Query Tool, the URL parameter is updated to reflect the enabled state.


dev-hari-prasad:
The PR is ready to merge. I have addressed all concerns of CodeRabbit. Pls consider merging this after updating the branch. I have used AI in the last two commit eg. [Refactoring: Clean up state initialization and URL checks](https://github.com/pgadmin-org/pgadmin4/pull/10055/commits/b8a2a4e4570bdf587c6ff494625562592bea7486) and [Update QueryToolComponent.jsx](https://github.com/pgadmin-org/pgadmin4/pull/10055/commits/aefee3f06ccb8ceea0be4876097746e47a58ea93), but rest assured, I have thoroughly read and understood the changes, made the initial changes myself, and verified everything manually.

Pls review the changes @asheshv if any changes are required, let me know.

web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -295,14 +292,14 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
...
-    } else if (qtState.params.sql_id && qtState.params.restore != 'true') {
+    } else if (qtState.params.sql_id && qtState.params.restore !== true) {
       let sqlValue = localStorage.getItem(qtState.params.sql_id);
       localStorage.removeItem(qtState.params.sql_id);
       if (sqlValue) {
         eventBus.current.fireEvent(QUERY_TOOL_EVENTS.EDITOR_SET_SQL, sqlValue);
       }
       setQtStatePartial({ editor_disabled: false });
-    } else if (qtState.params.restore == 'true') {
+    } else if (qtState.params.restore === true) {
```
mzabuawala, line 307: Changes looks good.
I observed that `restoreToolContent()` is async but is called without `await` here, can you fix that also?

web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -939,6 +916,18 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
     connection_list: qtState.connection_list,
     current_file: qtState.current_file,
     toggleQueryTool: () => setQtStatePartial((prev)=>{
+      let panel = qtPanelDocker?.find(qtPanelId);
+      if (panel?.metaData?.toolUrl) {
+        try {
+          const toolUrl = panel.metaData.toolUrl;
+          const isAbsolute = toolUrl.includes('://');
```
mzabuawala: IMO This check is not required

dev-hari-prasad:
I'll address the suggestions and requested changes in a day or a two.

dev-hari-prasad:
The requested changes and improvements from @mzabuawala and Co-pilto are both adresssed pls review once @asheshv

dpage:
Thanks @dev-hari-prasad, this is a solid fix for a genuinely nasty bug, a restored `DELETE`/`UPDATE` auto-running on startup is exactly the sort of thing we want to stop. The root-cause analysis is right: `is_query_tool` was only compared against the string `'true'`, so when session restore rehydrated it as a boolean it was misclassified and fell through to the auto-execute path. Normalising both `is_query_tool` and `restore` to real booleans and guarding the execution gate is the correct approach.

I traced the gate at `if(!is_query_tool || (reexecute && restore !== true))` across the cases and it behaves correctly: View/Edit Data still auto-executes, a Query Tool on normal open does not, reexecute (e.g. after reconnect) does, and a restored Query Tool does **not**. I also confirmed there are no leftover `== 'true'` comparisons for either flag and that `restore` is reset to `false` once `restoreToolContent()` completes.

On the existing threads:
- **`restoreToolContent()` async/await** (mzabuawala): addressed, `getSQLScript` is now `async` and awaits it.
- **`toggleQueryTool` layout not saved** (Copilot): addressed, `saveLayout()` is now called.
- **"the restore guard also blocks restored View/Edit Data tabs"** (Copilot): this no longer applies to the current code. The restore check sits only inside the `reexecute` sub-clause, and `!is_query_tool` short-circuits to true for View/Edit Data, so those tabs still auto-execute on restore. I think this one can be resolved, it was likely written against an earlier commit.
- **"IMO this check is not required"** (mzabuawala): this one is still open and has no line anchor, so it isn't clear which check is meant, @mzabuawala could you point at the specific line? That's the main thing left holding this up.

Two minor notes, neither blocking:
- `getSQLScript()` is now `async` but its caller in the mount effect invokes it without `await` (a `useEffect` callback can't be `async` anyway), so it runs concurrently with `initializeQueryTool()`. It's fine in practice because the auto-exec guard relies on the `restore === true` value captured at render time rather than on restore finishing first, so there's no race that re-enables execution, just be aware the `async` signature implies an ordering the caller doesn't enforce. Awaiting `restoreToolContent()` inside `getSQLScript` was the right call.
- Given how destructive the underlying bug is, a small regression test would be worthwhile. The full restore flow is awkward in Jest, but the cheap, high-value part, that `is_query_tool`/`restore` normalise from both boolean and string and that the gate does not fire for `reexecute && restore`, is unit-testable.

No changelog entry needed, by the way, we now add those in a batch just before release rather than per-PR.

dev-hari-prasad:
Is this PR ready for merge, or is there something else I need to address? 

@dpage

web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -939,6 +948,18 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
     connection_list: qtState.connection_list,
     current_file: qtState.current_file,
     toggleQueryTool: () => setQtStatePartial((prev)=>{
+      let panel = qtPanelDocker?.find(qtPanelId);
+      if (panel?.metaData?.toolUrl) {
+        try {
+          const toolUrl = panel.metaData.toolUrl;
+          const url = new URL(toolUrl, window.location.origin);
+          url.searchParams.set('is_query_tool', 'true');
+          panel.metaData = Object.assign({}, panel.metaData, {toolUrl: url.toString()});
```
dpage: **This one is a blocker, I'm afraid.** `new URL(toolUrl, window.location.origin)` followed by `url.toString()` always yields an *absolute* URL, so `metaData.toolUrl` changes from `/sqleditor/panel/123?...` to `http://127.0.0.1:<port>/sqleditor/panel/123?...`, and that is what gets persisted into the saved layout.

The Desktop runtime defaults to `fixedPort: false` (`runtime/src/js/pgadmin.js:24`) and therefore calls `getAvailablePort(0)`, so the port is different on every launch. At the next start-up, `ToolForm` will POST its `action` at a dead origin and the tab simply won't come back: cross-origin, no session, no CSRF token. That breaks tool tab restore in precisely the flow this PR is trying to repair, and Desktop mode is where the issue was reported from.

Note that this is also why dropping the earlier `isAbsolute` check made no difference either way: both variants ended in `toString()`. Keeping it relative is enough:

```js
panel.metaData = {...panel.metaData, toolUrl: url.pathname + url.search};
```

web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -385,7 +394,7 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
           obtaining_conn: false,
         });
         //this condition works if user is in View/Edit Data or user does not saved server or tunnel password and disconnected the server and executing the query
-        if(!qtState.params.is_query_tool || reexecute) {
+        if(!qtState.params.is_query_tool || (reexecute && qtState.params.restore !== true)) {
```
dpage: **The other blocker.** As written this clause is either dead code or an active regression, and never useful.

On the mount path (line 445) `initializeQueryTool()` is called with no arguments, so `reexecute` is always `false` and the new clause cannot affect the reported bug at all.

The only caller that passes `reexecute=true` is `REINIT_QT_CONNECTION`, fired from `ResultSet.jsx:259` and `:268` after the 428 password prompt and after the `CRYPTKEY_MISSING` master password prompt. That listener is registered once inside the mount-only effect at line 447 (dependency array `[]`), so it holds render one's `qtState`, in which `params.restore` is `true` for any restored tab and stays that way for the lifetime of the component. The `restore: false` that `restoreToolContent` now writes lands in state that this particular closure will never see.

The net effect is that on every session-restored Query Tool tab, for as long as that tab is open, re-entering a password after a dropped connection will reconnect and then silently not run the query, which is a fairly unpleasant thing to debug from a bug report.

I'd just drop the clause: the `toggleQueryTool` change is what actually fixes the reported issue.

web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -939,6 +948,18 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
     connection_list: qtState.connection_list,
     current_file: qtState.current_file,
     toggleQueryTool: () => setQtStatePartial((prev)=>{
+      let panel = qtPanelDocker?.find(qtPanelId);
```
dpage, line 951: Two smaller points on this block.

Firstly, the docker lookup, the `panel.metaData` mutation and the `saveLayout()` call all happen inside the `setQtStatePartial((prev)=>{...})` updater. Updater functions are required to be pure, and React is free to invoke them more than once; there's no `StrictMode` in this codebase today so it won't misbehave right now, but it's a trap for whoever turns it on. Moving the whole block above the `setQtStatePartial` call costs nothing.

Secondly, if `qtPanelDocker?.find(qtPanelId)` returns `undefined`, say because the panel lives in the other workspace's docker, the promotion silently isn't persisted and nothing is logged; only a URL parse failure warns. A `console.warn` on the miss would save someone a long afternoon.

web/pgadmin/tools/sqleditor/static/js/components/QueryToolComponent.jsx
```diff
@@ -282,7 +284,7 @@ export default function QueryToolComponent({params, pgWindow, pgAdmin, selectedN
     },
   };
 
-  const getSQLScript = () => {
+  const getSQLScript = async () => {
```
dpage: Minor: `getSQLScript` is now `async` and awaits `restoreToolContent()`, but its only caller (line 444) still doesn't await it, so the await buys nothing as things stand. Either await it there too, or leave the function synchronous.

dpage (commented):
Thanks for digging into this, and apologies for the slow review. I've gone through the change in full file context rather than just the diff. The underlying bug is real and nasty, and one of the three behavioural changes here is the right fix for it, but I think the other two are wrong, so I'd hold off merging as it stands.

First, a correction to the premise, because it matters for whoever reads this in a year's time: the `params` that reach `QueryToolComponent` are built from `request.args` and `request.form` in `panel()` (`web/pgadmin/tools/sqleditor/__init__.py:364`), so `is_query_tool` always arrives as a string and can never be a boolean. The actual mechanism is the promotion path: typing into the editor of a View/Edit Data tab calls `promoteToQueryTool()` (`Query.jsx:390`), which flips `is_query_tool` in memory only, whilst the persisted `metaData.toolUrl` still carries `is_query_tool=false`. On restore the tab is therefore treated as View/Edit Data, `!qtState.params.is_query_tool` is true, and `TRIGGER_EXECUTION` fires against whatever SQL was restored into the editor. That is the bug, and the `toggleQueryTool` change is exactly the right fix for it; the `is_query_tool` and `restore` normalisation is harmless defensive tidying that does no work today. It would be worth rewriting the PR description accordingly.

Detailed comments are inline, two of which I'd treat as blockers. A few smaller notes that don't attach neatly to a line:

* Anyone upgrading with an already-promoted tab in their saved layout will still have `is_query_tool=false` persisted, so they'll hit the bug once more after upgrading. Probably acceptable, but worth being aware of.
* There are no tests. The interesting behaviour here is docker and layout bound so a full test would be awkward, but a small Jest spec around the URL rewriting would be cheap, and would have caught the first of the two blockers below.
* CI is green (32 passed, 2 skipped); the `BLOCKED` merge state is purely the unresolved review threads.

No release notes entry is needed, incidentally: we now batch those up shortly before a release rather than doing one per PR.

dev-hari-prasad:
Hello, I have pushed a follow-up commit to fix a couple of edge cases found by Dave:

- **Saved relative `toolUrl`**: Switched from `url.toString()` to `url.pathname + url.search`. In Desktop mode (`fixedPort: false`), the backend port changes on restart, so saving an absolute URL broke tab restores against the old dead port.
- **Fixed password re-prompt query re-run**: Reverted the check in `initializeQueryTool` back to `!is_query_tool || reexecute`. The old `restore !== true` clause closed over initial render state, causing restored tabs to silently fail re-executing queries after a password prompt / dropped connection.
- **State updater cleanup**: Moved the panel lookup and `saveLayout()` call outside `setQtStatePartial` to keep the React updater pure, and added a warning if the panel lookup misses.
- **Made `getSQLScript` synchronous**: Removed the misleading `async/await` since the mount `useEffect` couldn't await it anyway.

Tested in desktop mode, and tab restores + promotions are working now. Ready for review, and sorry for the delay.

dpage:
Merged as https://github.com/pgadmin-org/pgadmin4/commit/e78c7fe02, rebased onto current `master` and squashed to a single commit with you as the author. GitHub won't auto-close it given the changed SHA, so I'm closing it by hand.

I folded in one small change on the way: `restoreToolContent()` was still called without `await` or `.catch`, which was @mzabuawala's outstanding point. Nothing in the current call chain actually rejects (`getToolContent` swallows everything into `return null`, and `warnFileReload` only ever resolves), so there was no live bug, but if that ever changes the consequence is an unhandled rejection with `editor_disabled` stuck at `true`, leaving the restored tab permanently read-only with nothing logged. It now catches, reports through `HANDLE_API_ERROR` and re-enables the editor.

Thanks for sticking with this one through several rounds of review; it's a genuinely nasty bug to have had in the wild, and the fix is a good one.

PR #10074, opened 2026-06-10
Preserve jsonb number representation in the JSON editor (#9854)

### LINKED ISSUE #9854: pgadmin doesn't preserve trailing zeros in jsonb
**Describe the bug**

When storing numbers with trailing zeros in jsonb, postgres will preserve those those:

> However, jsonb will preserve trailing fractional zeroes, as seen in this example, even though those are semantically insignificant for purposes such as equality checks.

However when I open such a jsonb document in pgadmin 4, the trailing zeros will not show in the edit window. This means that opening such a document, editing it and saving it via pgadmin will strip trailing zeros of numbers that weren't modified, which can break the appliation.

**To Reproduce**

    SELECT (10.00).to_jsonb

In the row view it shows `10.00`, but double clicking the field to open the pretty-printed json-editor view will replace it by `10`.

**Expected behavior**

Preserve the trailing zeros, so that round-tripping a json document through the json editor doesn't modify data.

**Screenshots**

Screenshot, row view works

Screenshot, editor view is broken

**Desktop**
 - OS: Freedesktop SDK 24.08 (Flatpak runtime), Linux-6.18.18-1-MANJARO-x86_64-with-glibc2.40
 - pgAdmin version: 9.14
 - Mode: Desktop
 - Browser: Chrome 146.0.7680.166
 - Package type: Flatpak

dpage:
## Summary

- The Query Tool's JSON cell editor pretty-printed jsonb values by round-tripping them through `json-bignumber` (`JSONBigNumber.stringify(JSONBigNumber.parse(value), null, 2)`). That preserves big integers but normalizes decimals through a JS float, so **trailing fractional zeros are dropped** (`10.00` → `10`, `3.140` → `3.14`).
- Because the reformatted text is what gets written back on save, opening an unrelated jsonb document and saving it **silently rewrote numbers the user never edited** — corrupting data for applications that depend on the canonical jsonb text. (The data-grid cell itself renders the raw value correctly; only the editor reformats lossily.)
- Fix: switch the editor to [`lossless-json`](https://www.npmjs.com/package/lossless-json) (the maintained sibling of `json-bignumber`), which preserves the exact numeric representation — big integers **and** trailing zeros. It's also passed to the underlying `vanilla-jsoneditor` as its `parser`, so the in-editor *Format* action and tree/table modes are lossless too.
- The lossless helpers are centralized in a small `json_utils` module with unit tests.

## Test plan

- [x] New Jest spec `regression/javascript/sqleditor/json_utils.spec.js` (5 tests) covering trailing-zero preservation, big integers, indentation, and the parser interface — all passing.
- [x] `eslint -c .eslintrc.js` clean on all changed files.
- [ ] Manual: in the Query Tool, `SELECT '10.00'::jsonb`; double-click the cell — the editor shows `10.00` (not `10`); save without editing — the stored value is unchanged.

Closes #9854

## Summary by CodeRabbit

* **Bug Fixes**
  * JSON editing and saving now preserve numeric formatting (trailing fractional zeros) and large integers in JSON/JSONB values, preventing unintended rewriting or corruption.

* **Tests**
  * Added tests to verify preservation of numeric formatting and correct pretty-printing behavior.

* **Chores**
  * Introduced a lossless JSON parser to support accurate parsing/stringifying during editing.

### DIFF
--- a/docs/en_US/release_notes_9_16.rst
+++ b/docs/en_US/release_notes_9_16.rst
@@ -43,6 +43,7 @@ Bug fixes
   | `Issue #9595 <https://github.com/pgadmin-org/pgadmin4/issues/9595>`_ -  Fix missing ALTER ... SET DEFAULT statements for inherited columns in the generated table SQL/EDIT script.
   | `Issue #9677 <https://github.com/pgadmin-org/pgadmin4/issues/9677>`_ -  Fix the Unlogged table toggle in table properties not generating any ALTER TABLE ... SET LOGGED/UNLOGGED statement.
   | `Issue #9828 <https://github.com/pgadmin-org/pgadmin4/issues/9828>`_ -  Fix tool calls failing against OpenAI-compatible providers that emit empty/null name, arguments, or id fields in streaming continuation deltas.
+  | `Issue #9854 <https://github.com/pgadmin-org/pgadmin4/issues/9854>`_ -  Fix the JSON editor stripping trailing fractional zeros (e.g. 10.00) and rewriting large integers in jsonb values, which corrupted unmodified numbers when saving.
   | `Issue #9875 <https://github.com/pgadmin-org/pgadmin4/issues/9875>`_ -  Fixed an issue where EXPLAIN and EXPLAIN ANALYZE failed to execute when blank lines separated clauses in the SQL query.
   | `Issue #9810 <https://github.com/pgadmin-org/pgadmin4/issues/9810>`_ -  Use the ServerManager's passfile for the credential gate in connect() so the check matches the passfile actually used for the connection, and warn on conflicting passfile/passexec settings.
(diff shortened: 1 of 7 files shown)


dpage:
@coderabbitai review

mzabuawala:
Tested & working fine.

asheshv (approved):
LGTM. The `lossless-json` swap is the right call — it's the documented companion for vanilla-jsoneditor and the `{parse, stringify}` interface is exactly the intended pattern. Round-trip works correctly in both text and tree/table modes.

Minor nit (non-blocking): consider an `expect(() => prettifyJSONString('{bad')).toThrow()` test to lock in the throw-on-invalid contract — `jsonb` columns guarantee valid input so this is purely defensive.

PR #10075, opened 2026-06-10
Warn before opening a very large JSON value in the data grid (#9868)

### LINKED ISSUE #9868: pgAdmin 4 UI Hangs when interacting with large JSONB datasets
Status & Severity
•	Severity: High (Application unresponsiveness)
•	Component: Query Tool / Data Grid
•	Environment: Windows 10/11 x64, pgAdmin 4 v8.x (bundled with PostgreSQL 18.3)
•	Browser/Runtime: NW.js (Desktop Mode/Server)
________________________________________
Summary
The pgAdmin 4 interface becomes completely unresponsive (hangs) when attempting to interact with the "View Data" grid containing large JSONB objects. The issue is specifically triggered after expanding a large cell and subsequently attempting to use grid controls like "Filter."
Steps to Reproduce
1.	Environment Setup: Install PostgreSQL 18.3 (EnterpriseDB Windows x64).
2.	Launch: Open pgAdmin 4 and connect to a local/remote server.
3.	Data Generation: Open the Query Tool and execute the following script to create a "heavy" record:
SQL
CREATE TABLE performance_test (
    id serial PRIMARY KEY,
    wide_data jsonb
);

INSERT INTO performance_test (wide_data)
SELECT jsonb_object_agg('column_' || i, 'value_' || i)
FROM generate_series(1, 100000) AS i;
4.	Access Data: Right-click the performance_test table in the Browser tree -> View/Edit Data -> All Rows.
5.	Trigger Expansion: Double-click the result cell in the wide_data column to view the expanded JSON content.
6.	Interaction: While the cell is active or immediately after, click the Filter icon in the data grid toolbar.
Actual Result
The pgAdmin 4 window freezes entirely. No further clicks are registered, and the application must be force-closed via Task Manager.
Expected Result
The UI should remain responsive. Grid operations (Filtering/Sorting) should either execute normally or provide a "Loading" state without locking the main thread.
Error Log / Console Output
The following violation is observed in the developer console:
[Violation] Permissions policy violation: unload is not allowed in this document.
________________________________________
Preliminary Analysis
The hang appears to be a combination of:
•	DOM Exhaustion: Attempting to render or manage a single cell containing 100,000+ JSON keys in the "Edit" modal.
•	Main Thread Blocking: The UI thread is likely saturated by the string serialization of the JSONB object, causing the "unload" violation when the application attempts to transition states (like opening a filter dialog).
Workaround
Limit the JSON size using jsonb_pretty() or substring() in the Query Tool rather than using the "View Data" shortcut for large datasets.

Snapshot:

Image

dpage:
## Summary

- Opening a very large JSON/JSONB cell in the Query Tool's cell editor parses, pretty-prints and renders the whole document on the main thread. For a pathologically large value (the report uses a `jsonb` object with 100,000 keys, ~3 MB) this **blocks the UI thread and pgAdmin becomes completely unresponsive**, with no opportunity for the user to back out.
- This matches the maintainer's assessment on the issue: there should be a warning for large JSON so the user can decide whether to proceed — but currently no such warning fires.
- Fix: gate `JsonTextEditor` — when the raw cell value exceeds a size threshold (1 MB of characters), render nothing until the user confirms via a warning dialog. If they cancel, the editor closes without doing the expensive parse/render. Small values are unaffected and open immediately, exactly as before. The grid already sets `commitOnOutsideClick: false`, so the confirm dialog does not dismiss the editor.
- The size-threshold decision is factored into a small, separately unit-tested helper (`json_editor_warning.js`).

## Test plan

- [x] New Jest spec `regression/javascript/sqleditor/json_editor_warning.spec.js` (5 tests) covering the threshold boundary, custom thresholds, and non-string values — all passing.
- [x] `eslint -c .eslintrc.js` clean on all changed files.
- [ ] Manual: build the reporter's table, View/Edit Data, double-click the `wide_data` cell — a confirmation dialog appears; **Cancel** leaves the UI responsive, **Continue** opens the editor as before. A small jsonb cell opens with no prompt.

Closes #9868

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a confirmation warning before opening very large JSON/JSONB values in the data grid editor to avoid freezing; users can proceed or cancel.

* **Documentation**
  * Updated 9.16 release notes with the new bug-fix entry.

* **Tests**
  * Added regression tests covering the large-JSON warning logic and user confirmation flows.

### DIFF
--- a/web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
@@ -15,6 +15,7 @@ import CloseIcon from '@mui/icons-material/Close';
 import gettext from 'sources/gettext';
 import JSONBigNumber from 'json-bignumber';
 import JsonEditor from '../../../../../../static/js/components/JsonEditor';
+import { shouldWarnForLargeJSON } from './json_editor_warning';
 import PropTypes from 'prop-types';
 import { RowInfoContext } from '.';
 import { usePgAdmin } from '../../../../../../static/js/PgAdminProvider';
@@ -388,7 +389,7 @@ export function CheckboxEditor({row, column, onRowChange, onClose}) {
 }
 CheckboxEditor.propTypes = EditorPropTypes;
 
-export function JsonTextEditor({row, column, onRowChange, onClose}) {
+function JsonTextEditorContent({row, column, onRowChange, onClose}) {
 
   const pgAdmin = usePgAdmin();
 
@@ -474,5 +475,32 @@ export function JsonTextEditor({row, column, onRowChange, onClose}) {
       </ResizableDiv>
     </Portal>
   );
+}
+JsonTextEditorContent.propTypes = EditorPropTypes;
+
+// Wraps the JSON editor with a guard that warns before opening a very large
+// value. Rendering/parsing a huge JSON document blocks the main thread and can
+// freeze pgAdmin, so for large cells we render nothing until the user confirms
+// (#9868). The editor uses commitOnOutsideClick: false, so showing the confirm
+// dialog does not dismiss the editor.
+export function JsonTextEditor(props) {
+  const { row, column, onClose } = props;
+  const pgAdmin = usePgAdmin();
+  const [proceed, setProceed] = React.useState(
+    () => !shouldWarnForLargeJSON(row[column.key])
+  );
+
+  React.useEffect(() => {
+    if (proceed) return;
+    pgAdmin.Browser.notifier.confirm(
+      gettext('Large data'),
+      gettext('This cell contains a large amount of data. Opening it in the JSON editor may make pgAdmin slow or unresponsive. Do you want to continue?'),
+      () => setProceed(true),
+      () => onClose(false),
+    );
+  }, []);
+
+  if (!proceed) return null;
+  return <JsonTextEditorContent {...props} />;
 }
 JsonTextEditor.propTypes = EditorPropTypes;
--- a/web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
@@ -492,11 +492,22 @@ export function JsonTextEditor(props) {
 
   React.useEffect(() => {
     if (proceed) return;
+    // notifier.confirm() fires its cancel callback whenever the dialog
+    // closes, including when the user clicks OK. Track the confirmation so
+    // we only close the cell editor on an actual cancel and not after the
+    // user chose to continue (#9868).
+    let confirmed = false;
     pgAdmin.Browser.notifier.confirm(
       gettext('Large data'),
       gettext('This cell contains a large amount of data. Opening it in the JSON editor may make pgAdmin slow or unresponsive. Do you want to continue?'),
-      () => setProceed(true),
-      () => onClose(false),
+      () => {
+        confirmed = true;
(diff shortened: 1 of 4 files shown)


web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
```diff
@@ -475,4 +476,31 @@ export function JsonTextEditor({row, column, onRowChange, onClose}) {
...
+      gettext('Large data'),
+      gettext('This cell contains a large amount of data. Opening it in the JSON editor may make pgAdmin slow or unresponsive. Do you want to continue?'),
+      () => setProceed(true),
+      () => onClose(false),
+    );
+  }, []);
+
+  if (!proceed) return null;
+  return <JsonTextEditorContent {...props} />;
+}
```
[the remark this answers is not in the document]
dpage: Good catch — confirmed. `notifier.confirm()` registers the modal with `onClose: onCancelClick`, and the OK handler calls `closeModal()` with no `reason`, so `ModalContainer` fires `onClose` on the OK path too. That meant continuing past the warning ran `onClose(false)` and closed the cell editor before the JSON editor could open. Fixed in 4b57462e7 by tracking the confirmation in the effect closure so the cancel callback is a no-op once OK has been chosen, plus a regression test covering the continue path.

web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
```diff
@@ -475,4 +476,42 @@ export function JsonTextEditor({row, column, onRowChange, onClose}) {
...
   );
 }
+JsonTextEditorContent.propTypes = EditorPropTypes;
+
+// Wraps the JSON editor with a guard that warns before opening a very large
+// value. Rendering/parsing a huge JSON document blocks the main thread and can
+// freeze pgAdmin, so for large cells we render nothing until the user confirms
+// (#9868). The editor uses commitOnOutsideClick: false, so showing the confirm
+// dialog does not dismiss the editor.
+export function JsonTextEditor(props) {
```
mzabuawala, line 486: @dpage @asheshv Can we try **WebWorker** here? 
It may help us keep the UI responsive while parsing.

current `JsonTextEditor` code has - 
```
JSONBigNumber.parse(....)
JSONBigNumber.stringify(...)
```
which runs on the main thread and can freeze the UI.

Suggested flow would be - 
Cell clicked -> Open editor immediately (show spinner) -> Web Worker (parse + pretty-print) -> postMessage(formattedJson) -> Update editor content

dpage: Thanks Murtuza. I agree that the JSONBigNumber.parse/stringify on the main thread is a real contributor to the freeze, and a WebWorker is a sensible way to take that off the UI thread.

The wrinkle is that the parse/stringify is only part of the cost. Once we have the formatted string, it still gets handed to the CodeMirror-based JsonEditor, and mounting + tokenising/highlighting a multi-megabyte document is itself a heavy main-thread operation that a worker can't remove. So a worker would make the parse async but wouldn't fully fix the freeze on very large values - we'd also need editor-side mitigations (e.g. large-doc mode / disabling highlighting above a threshold).

Given that, I'd suggest we keep this PR as the immediate, low-risk safety net - it stops an accidental click from locking up the UI regardless of where the time goes. We could consider a worker as a future enhancement (once our renewed efforts have reduced the backlog a bit!)

mzabuawala: Tested working fine

PR #10076, opened 2026-06-10
fix: Adjust relationship between ServerGroup and Server/SharedServer

lkmatsumura:
- Adjusting relationship
  - ServerGroup has many Servers
  - ServerGroup has many SharedServers
  - Server has one ServerGroup not servers
  - SharedServer has one ServerGroup not servers
  - Using back_populates instead of backref for better control

- Adjusted references to ServerGroup in server (was server.serves): server.servergroup.name instead server.servers.name
- Not find references to ServerGroup in SharedServer (sharedServer.servers)

## Summary by CodeRabbit

* **Refactor**
  * Restructured database model relationships to improve server organization and grouping management.
  * Updated server grouping references in schema comparison and SQL editor tools for consistency across the application.

### DIFF
--- a/web/pgadmin/model/__init__.py
+++ b/web/pgadmin/model/__init__.py
@@ -252,6 +252,20 @@ class ServerGroup(db.Model, UserScopedMixin):
     name = db.Column(db.String(128), nullable=False)
     __table_args__ = (db.UniqueConstraint('user_id', 'name'),)
 
+    servers = db.relationship(
+        'Server',
+        back_populates='servergroup',
+        lazy='select',
+        cascade=CASCADE_STR
+    )
+
+    sharedservers = db.relationship(
+        'SharedServer',
+        back_populates='servergroup',
+        lazy='select',
+        cascade=CASCADE_STR
+    )
+
     @property
     def serialize(self):
         """Return object data in easily serializable format"""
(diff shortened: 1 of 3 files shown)


dpage (commented):
This is a reasonable readability cleanup — the `Server.servers` relationship was confusingly named (it pointed *to* the one `ServerGroup`), so `Server.servergroup` / `ServerGroup.servers` + `back_populates` reads much better.

I checked the rest of the tree: the three call sites you updated (`schema_diff` ×3, `sqleditor` ×1) are the only real users of the old `server.servers` attribute, and nothing references the old auto-generated backref names (`servergroup.server` / `.sharedserver`), so dropping them is safe. The `cascade` is correctly carried over onto the new `ServerGroup.servers`/`sharedservers` relationships, and the FK columns are unchanged — so no schema/migration impact.

Two things before merge:
- No changelog entry — this still wants a one-line "Housekeeping" entry in the current `docs/en_US/release_notes_*.rst`.
- It's a pure rename with no functional change and no test; that's fine for a cleanup, but worth confirming the relationship rename doesn't disturb any serialization that relied on the attribute name.

dpage:
Merged to `master` as dfc4ef37d (squashed; the SHA differs so GitHub won't auto-close this). I folded in a small tidy on top of your change, dropping the cosmetic blank lines between the existing column definitions so the diff stays purely semantic, and reworded the subject to `refactor:` since this is a naming/modelling cleanup with no behavioural change. I confirmed the rewired `back_populates` mappers configure cleanly at app startup and verified the two call sites were the only consumers of the old relationship. Thanks for untangling this, the new direction reads far better!

PR #10077, opened 2026-06-10
refactor(Servergroup):  Refactor server group access and shared-group handling

lkmatsumura:
Summary
-------
Refactor server group access and shared-group handling in pgAdmin4 to centralize visibility logic and simplify browser rendering.

What changed
------------
- Updated `web/pgadmin/utils/server_access.py`
  - Added `get_server_groups_for_user()` and `get_server_groups_for_user_query()`
  - Simplified `get_server_group()` to reuse shared access logic
  - Added `is_shared_group` metadata on returned groups
  - Consolidated shared-group filtering, ownership checks, and ordering

- Updated `web/pgadmin/browser/server_groups/__init__.py`
  - Replaced ad hoc shared-server checks with centralized access helpers
  - Removed redundant `ServerGroupModule.has_shared_server()` logic
  - Ensured shared server groups are rendered with the correct icon
  - Prevented deletion of shared server groups
  - Improved deletion error messages for shared groups and first group restrictions

Notes
-----
- This PR focuses on server group refactor work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deletion error handling for server groups that contain shared servers and for deletion attempts affecting the first server group.

* **Improvements**
  * Refined server-group visibility, shared-state handling, and icon selection across listings and tree nodes.
  * Updated server-group lookup and targeted retrieval to better respect shared/hide behavior.
  * Improved UI permissions by deriving edit/drop actions from the server-group’s provided flags.

### DIFF
--- a/web/pgadmin/browser/server_groups/__init__.py
+++ b/web/pgadmin/browser/server_groups/__init__.py
@@ -218,8 +218,8 @@ class ServerGroupView(NodeView):
     def update(self, gid):
         """Update the server-group properties"""
 
-        # There can be only one record at most
-        servergroup = get_server_group(gid)
+        # There can be only one record at most (only owned)
+        servergroup = get_server_group(gid, hide_shared=True)
 
         data = request.form if request.form else json.loads(
             request.data
--- a/web/pgadmin/browser/server_groups/__init__.py
+++ b/web/pgadmin/browser/server_groups/__init__.py
@@ -76,7 +76,7 @@ class ServerGroupModule(BrowserPluginModule):
                 icon_class,
                 True,
                 self.node_type,
-                can_delete=True if idx > 0 else False,
+                can_delete=True if idx > 0 and not group.is_shared_group else False,
                 user_id=group.user_id,
                 is_shared=is_shared
             )
--- a/web/pgadmin/browser/server_groups/__init__.py
+++ b/web/pgadmin/browser/server_groups/__init__.py
@@ -29,18 +29,15 @@ from pgadmin.utils.server_access import get_server_group, \
     get_server_groups_for_user
 
 
-def get_icon_css_class(group_id, group_user_id,
-                       default_val='icon-server_group'):
+def get_icon_css_class(is_shared_group, default_val='icon-server_group'):
     """
     Returns css value
-    :param group_id:
-    :param group_user_id:
+    :param is_shared_group:
     :param default_val:
     :return: default_val
     """
-    if (config.SERVER_MODE and
-        group_user_id != current_user.id and
-            ServerGroupModule.has_shared_server(group_id)):
+    if (config.SERVER_MODE
+        and is_shared_group):
         default_val = 'icon-server_group_shared'
         return default_val, True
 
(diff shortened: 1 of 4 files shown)


dpage (commented):
Thanks for taking this on — centralizing the visibility logic in `server_access.py` and replacing the per-server `has_shared_server()` loop with a single SQL `EXISTS` is a nice direction. A few things need addressing before this is mergeable:

**1. Desktop mode is broken — `(0).label('is_shared_group')` raises `AttributeError`.**
In `get_server_groups_for_user_query()`, the non-`SERVER_MODE` branch does:
```python
ServerGroup.query.add_columns((0).label('is_shared_group'))
```
`int` has no `.label()` method, so this throws at query-build time (verified locally). Since `get_nodes()` was also changed to always route through `get_all_server_groups()`, the server-group tree fails to render in desktop mode — the default deployment. Use `from sqlalchemy import literal` and `literal(0).label('is_shared_group')`.

**2. Access regression in `update()`.**
`update()` previously fetched `ServerGroup.query.filter_by(user_id=current_user.id, id=gid)`. It now uses `get_server_group(gid)`, which in `SERVER_MODE` returns groups the user does *not* own as long as they contain a shared server. The method then does `servergroup.name = data['name']; db.session.commit()` with no ownership guard, so a user can rename another user's group. Deletion is correctly gated against shared groups (`can_delete`), but rename needs the same owner check.

**3. Style checks will fail CI.**
- `Server.shared == True` → E712 (use `Server.shared.is_(True)` or just `Server.shared`).
- Only one blank line between the two new top-level functions → E302.
- A couple of lines exceed 79 chars (the `.all()` call and a docstring line) → E501.
- `for group, is_shared  in sg:` has a double space.

**4. No tests, no changelog entry.** A refactor touching access control should have a test for the owned / shared / desktop paths, and `docs/en_US/release_notes_*.rst` needs a line.

Since this isn't tied to a reported issue, it'd also help to note what user-facing problem it's solving — the diff reads as pure cleanup, so the bar is "demonstrably no behaviour change," and right now #1/#2 are behaviour changes.

lkmatsumura:
@dpage, Thank's by the review. 
I don't know what and how to do the item 4. Hope the changes commited fixes the itens 1,2 and 3.

dpage:
Thanks for the updates @lkmatsumura, this has come a long way. Items 1 and 2 from my earlier review are sorted: desktop mode no longer throws (the `literal(0)` fix), and `update()` is now owner-gated via `get_server_group(gid, hide_shared=True)` with a proper `None` guard, so a user can no longer rename another user's group. The unused `current_user` import and the missing `can_delete` on `nodes()` are also resolved. Two things still need attention before this is mergeable, and then there's item 4.

**1. `make check-pep8` is currently failing (CI red, 25 violations).** Running `pycodestyle --config=.pycodestyle` over the two Python files, the ones to fix are:
- `utils/server_access.py`: `E201`/`E202` (stray spaces in `( ServerGroup...` and `hide_shared=hide_shared )`), `E126`/`E131` continuation-indent on the `return ( ServerGroup.query.add_columns(...)` blocks, and `E303` (two blank lines inside the function, ~line 148).
- `browser/server_groups/__init__.py`: `E501` on the two new long error-message strings (lines 185 and 194) and line 391; plus `W503`/`E128` on the `can_delete=(not (... \n or ...))` and `get_icon_css_class` boolean continuations. Note our config ignores `W504` but **not** `W503`, so a line break *before* `and`/`or` trips it: move the operator to the end of the previous line, and wrap the long error strings across two lines.

**2. `canEdit` regression: `can_edit` is missing from three of the five node paths.** The base `Node.canEdit` defaults to `true`, and this PR overrides it to read `itemData.can_edit`. But `can_edit` is only passed in `get_nodes()` and `update()`; it's omitted in `create()` and both branches of `nodes()`. Nodes from those paths therefore get `can_edit === undefined`, which disables Edit/Properties even for the user's own groups (e.g. a group can't be renamed right after it's created, until a tree refresh). `can_delete` is set in all five paths, so `canDrop` is fine; `can_edit` just needs the same treatment, `can_edit=(not <group>.is_shared_group)` in those three calls (always `True` for the freshly-created group in `create()`).

**On item 4 (the part you were unsure about):**
- *Changelog: you can skip it.* We've recently moved to adding release-notes entries in a batch just before release rather than per-PR, so there's no need to touch `release_notes_*.rst` here.
- *Tests: extend the existing suite.* There's already `web/pgadmin/browser/server_groups/tests/`, and `test_sg_data_isolation.py` and `test_sg_get.py` are good models. The behaviours worth covering are the ones this refactor reshapes: an owned group is visible/editable/deletable; the first owned group is **not** deletable; a group containing another user's shared server is visible but **not** editable or deletable (and reports `is_shared_group=True` with the shared icon); and the desktop path returns owned groups with `is_shared_group=0`. Following the shape of `test_sg_data_isolation.py` is the most direct route.

One minor aside, not blocking: in `get_all_server_groups()` the `hide_shared_server` fallback defaults to `True` (hide) when the preference object is absent, which is the opposite of the preference's registered default; defaulting to the registered value would be marginally tidier, though that path shouldn't normally occur.

Thanks again, the centralised approach reads well, it's mostly the lint and the `can_edit` consistency now.

lkmatsumura:
@dpage I think the style code is ok now, The tests since I don't know how to write it, is AI generated, but it's seens correct to me.

Issue #9942, opened 2026-05-18
Helm chart support gateway api


tom171296:
**Describe the solution you'd like**

The helm chart now only supports the kubernetes ingress api. This api is no longer actively developed. The new exposure should be done via the Gateway API (HTTP Route).

**Describe alternatives you've considered**

-

**Additional context**

asheshv:
Please send a PR.

somaz94:
Opened a draft PR adding an opt-in Gateway API `HTTPRoute` template to the Helm chart (disabled by default, mirrors the existing Ingress conventions): pgadmin-org/pgadmin4#10095. Happy to adjust the values layout or naming to match your preferences.

asheshv:
Fixed in pgAdmin 4 9.17, released 2026-07-30.

PR #10095, opened 2026-06-16
Add Gateway API HTTPRoute support to the Helm chart

somaz94:
Adds an opt-in Gateway API `HTTPRoute` template to the pgAdmin Helm chart as an alternative to the existing Ingress, addressing #9942. Disabled by default (`httpRoute.enabled: false`) so existing installs are unaffected.

The template mirrors the existing `ingress.yaml` conventions: `pgadmin4.fullname` backend, `commonLabels`/`commonAnnotations` propagation, and a hostname that falls back to `ingress.hostname` when `httpRoute.hostnames` is unset. `parentRefs` is required (a `fail` guard fires when it is empty), `apiVersion` defaults to `gateway.networking.k8s.io/v1`, and the default rule forwards `/` (PathPrefix) to the pgAdmin service. Custom `hostnames` and `rules` can be supplied for full control.

Validation (local `helm` v3.18):
- Default values render **no** HTTPRoute (`kind: HTTPRoute` count = 0).
- `httpRoute.enabled=true` with `parentRefs` + `hostnames` renders a valid HTTPRoute (backendRef → service port 80, PathPrefix `/`).
- `httpRoute.enabled=true` without `parentRefs` fails fast with a clear message.
- Empty `httpRoute.hostnames` correctly falls back to `ingress.hostname`.
- Custom `httpRoute.rules` override renders as supplied.
- `helm lint` clean both with the feature disabled and enabled.

related: #9942

---

##### Was generative AI tooling used to co-author this PR?

- [X] Yes — Claude Code (Opus 4.8)

Generated-by: Claude Code (Opus 4.8)

## Summary by CodeRabbit

* **New Features**
  * Added support for Kubernetes HTTPRoute resources with Gateway API integration
  * HTTPRoute can be conditionally enabled with configurable parent references, hostnames, and custom routing rules
  * Routes all traffic to the pgAdmin4 service by default when no custom rules are specified

### DIFF
--- a/pkg/helm/templates/httproute.yaml
+++ b/pkg/helm/templates/httproute.yaml
@@ -1,0 +1,36 @@
+{{- if .Values.httpRoute.enabled }}
+{{- if not .Values.httpRoute.parentRefs }}
+{{- fail "httpRoute.parentRefs must be set when httpRoute.enabled is true" }}
+{{- end }}
+apiVersion: {{ .Values.httpRoute.apiVersion | default "gateway.networking.k8s.io/v1" }}
+kind: HTTPRoute
+metadata:
+  name: {{ template "pgadmin4.fullname" . }}
+  {{- with .Values.commonLabels }}
+  labels: {{ . | toYaml | nindent 4 }}
+  {{- end }}
+  {{- if or .Values.httpRoute.annotations .Values.commonAnnotations }}
+  annotations: {{ merge .Values.httpRoute.annotations .Values.commonAnnotations | toYaml | nindent 4 }}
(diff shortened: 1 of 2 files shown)


dpage:
Merged to `master` as 7e76fb4eb (squashed; the SHA differs so GitHub won't auto-close this). I folded in one small addition on top of your work: rows for the `httpRoute.*` values in the chart README's *Important Values* table, for parity with the existing `ingress.*` entries. Nicely done, and thanks for the thorough `helm template`/`helm lint` validation in the description!

PR #10113, opened 2026-06-19
Warn when OAuth2 provider settings are misplaced at the top level of the config

### LINKED ISSUE #10053: oauth regression: ssl verification error with self-signed certificate
I use oauth login to pgAdmin in docker with self-signed certificate and Authentik as IdP.
SSL verification is disabled because of the self-signed certificate.

Docker compose environment variable:
PGADMIN_CONFIG_OAUTH2_SSL_CERT_VERIFICATION=False

OAuth login works fine up to version 9.10.
With Version 9.11 to 9.15 I get an SSL verification error.

**To Reproduce**

1. Install pgAdmin 9.10, configure oauth with self-signed certificate and login
2. Login works
3. Update to pgAdmin 9.11 or newer
4. Login fails

**Expected behavior**

OAuth login should work.

**Error message**

{"success":0,"errormsg":"HTTPSConnectionPool(host='authentik.example.com', port=443): Max retries exceeded with url: /application/o/pgadmin/.well-known/openid-configuration (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Basic Constraints of CA cert not marked critical (_ssl.c:1081)')))","info":"","result":null,"data":null}

 - OS: Synology NAS
 - pgAdmin version: 9.10 working, 9.11+ broken
 - Mode: Server
 - Browser (if running in server mode): Firefox 151.0.3
 - Package type: docker compose container

dpage:
## Summary

pgAdmin reads OAuth2 provider settings **only** from the `OAUTH2_CONFIG` list; each provider is a single dictionary entry within it. Settings supplied at the top level of the configuration (for example a bare `OAUTH2_CLIENT_ID` or `OAUTH2_SSL_CERT_VERIFICATION`) are silently ignored.

This is an easy trap in container deployments: nearly every other option is set with an individual `PGADMIN_CONFIG_<KEY>` environment variable, so it is natural to assume `PGADMIN_CONFIG_OAUTH2_CLIENT_ID`, `PGADMIN_CONFIG_OAUTH2_SSL_CERT_VERIFICATION` and friends will behave the same way. They do not: OAuth2 must be configured through a single `PGADMIN_CONFIG_OAUTH2_CONFIG` holding the whole provider list. The result is that a user's settings, including their attempt to disable SSL certificate verification for a self-signed provider, are dropped without any feedback.

This change:

- Adds `warn_on_misplaced_oauth2_config()`, called once at startup from the OAuth2 module's `init_app`, which logs an actionable warning when per-provider `OAUTH2_*` keys are present at the top level of the configuration **and** no provider is configured in `OAUTH2_CONFIG`. There is no behavioural change: a correctly configured deployment, or one with no OAuth2 settings at all, stays silent.
- Documents the supported container-deployment approach in `oauth2.rst`, making clear that individual `PGADMIN_CONFIG_OAUTH2_*` variables do not configure a provider.

This is deliberately a *fail-loud* fix rather than a compatibility shim: `OAUTH2_CONFIG` remains the single source of truth, and we avoid quietly expanding the supported configuration surface (and the reach of `verify=False`).

Note on the "regression" in the linked issue: the OAuth2 `verify` handling did not change in code between the affected releases. The actual trigger is stricter certificate verification (OpenSSL/urllib3) in the newer image rejecting a CA certificate whose Basic Constraints extension is not marked critical; the user's `OAUTH2_SSL_CERT_VERIFICATION=False` escape hatch never took effect because it was supplied as an individual environment variable. This change makes that misconfiguration visible.

## Test plan

- [x] New test `test_oauth2_misplaced_config.py` covering: warns on misplaced keys + unconfigured `OAUTH2_CONFIG`; quiet when a provider is configured; quiet for legitimate top-level settings only; quiet with no OAuth2 config.
- [x] Existing `test_oauth2_with_mocking` still passes (58/58).
- [x] `pycodestyle` clean on changed Python files.

Closes #10053

## Summary by CodeRabbit

* **Documentation**
  * Expanded the OAuth2 setup guide to clarify that provider settings must be defined within the `OAUTH2_CONFIG` list; other top-level OAuth2 keys are ignored.
  * Added notes for container usage, including `PGADMIN_CONFIG_OAUTH2_CONFIG`, and described the startup warning behavior.

* **New Features**
  * Added startup-time validation that detects misplaced OAuth2 settings and logs a warning listing the ignored keys.

* **Tests**
  * Added automated coverage to verify warning conditions across multiple configuration scenarios.

### DIFF
--- a/docs/en_US/oauth2.rst
+++ b/docs/en_US/oauth2.rst
@@ -11,6 +11,25 @@ file (see the :ref:`config.py <config_py>` documentation) on the system where
 pgAdmin is installed in Server mode. You can copy these settings from *config.py*
 file and modify the values for the following parameters.
 
+.. warning::
+   pgAdmin reads OAuth2 provider settings **only** from the ``OAUTH2_CONFIG``
+   list; each provider is a single dictionary entry within that list. Settings
+   such as ``OAUTH2_CLIENT_ID`` or ``OAUTH2_SSL_CERT_VERIFICATION`` defined at
+   the top level of the configuration are ignored.
+
+   This matters in container deployments. Although most options can be set with
+   an individual ``PGADMIN_CONFIG_<KEY>`` environment variable, OAuth2 is the
+   exception: individual variables such as ``PGADMIN_CONFIG_OAUTH2_CLIENT_ID``
+   or ``PGADMIN_CONFIG_OAUTH2_SSL_CERT_VERIFICATION`` will **not** configure a
+   provider. Instead, configure the whole provider list through a single
+   ``PGADMIN_CONFIG_OAUTH2_CONFIG`` variable, for example::
+
+       PGADMIN_CONFIG_OAUTH2_CONFIG="[{'OAUTH2_NAME': 'my-provider', 'OAUTH2_DISPLAY_NAME': 'My Provider', 'OAUTH2_CLIENT_ID': '...', 'OAUTH2_CLIENT_SECRET': '...', 'OAUTH2_SERVER_METADATA_URL': 'https://provider.example.com/.well-known/openid-configuration', 'OAUTH2_SCOPE': 'openid email profile', 'OAUTH2_SSL_CERT_VERIFICATION': False}]"
+
+   If pgAdmin detects per-provider OAuth2 settings at the top level while no
+   provider is configured in ``OAUTH2_CONFIG``, it logs a warning at startup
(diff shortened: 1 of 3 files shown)


hiteshjambhale (approved):
This is tested and the mentioned scenario working fine. Approving it.

PR #10159, opened 2026-07-17
fix: dedupe concurrent getNodeAjaxOptions requests (#10155)

### LINKED ISSUE #10155: Duplicate concurrent get_types (and other dropdown) requests when many rows mount at once
## Describe the bug

`getNodeAjaxOptions()` in `web/pgadmin/browser/static/js/node_ajax.js` — the shared helper behind every AJAX-backed dropdown in pgAdmin (Data Type, foreign keys, triggers, backup/restore options, cloud wizards, ~35 call sites) — caches responses, but only *after* they resolve:

```js
let data = cacheNode.cache(nodeObj.type + '#' + url, treeNodeInfo, cacheLevel);
if (_.isUndefined(data) || _.isNull(data)) {
  api.get(fullUrl, {...}).then((res) => {
    ...
    cacheNode.cache(..., data);   // only written once the response comes back
  });
}
```

There is no in-flight request deduplication. If several components call this for the same URL before the first response lands, every one of them sees an empty cache and fires its own HTTP GET. Each row's Data Type dropdown (`column.ui.js`, `cacheLevel: 'table'`) does exactly this on mount.

This code path exists on `master` as-is — `node_ajax.js` isn't touched by #10146. This issue was found while testing #10146 ("fix: skip row virtualization for small data grids"); we haven't separately confirmed how many duplicate requests fire on plain `master` for the same table, so we're not claiming that PR changes the severity — only reporting what we observed.

## To Reproduce

1. Create a table with 150+ columns.
2. Open DevTools → Network tab.
3. Right-click the table → Properties → Columns tab.
4. Observe ~150 concurrent `GET /browser/table/get_types/<gid>/<sid>/<did>/<scid>/` requests, all firing within the same ~50-100ms window — one per column row, all for the identical options list.

## Expected behavior

A single `get_types` request should be shared by all concurrent callers; the rest should await that same in-flight request rather than issuing their own.

## Additional context
x
Found while testing #10146 — noting that for traceability, not to imply it's the cause. The underlying race is in `node_ajax.js`, which that PR does not modify.

## Proposed fix

Add in-flight promise sharing to `getNodeAjaxOptions`, keyed by the resolved URL + query params, so concurrent callers await the same underlying request:

```js
const _inflightAjaxRequests = new Map();
...
if (_.isUndefined(data) || _.isNull(data)) {
  const inflightKey = fullUrl + '#' + JSON.stringify(otherParams.urlParams || {});
  let request = _inflightAjaxRequests.get(inflightKey);
  if (!request) {
    request = api.get(fullUrl, { params: otherParams.urlParams }).then((res) => {
      let resData = res.data.data ?? res.data;
      otherParams.useCache && cacheNode.cache(nodeObj.type + '#' + url, treeNodeInfo, cacheLevel, resData);
      return resData;
    }).finally(() => { _inflightAjaxRequests.delete(inflightKey); });
    _inflightAjaxRequests.set(inflightKey, request);
  }
  request.then((resData) => resolve(transform(resData))).catch((err) => reject(...));
}
```

hiteshjambhale:
## Summary

Fixes #10155.

`getNodeAjaxOptions()` in `web/pgadmin/browser/static/js/node_ajax.js` — the shared helper behind every AJAX-backed dropdown — cached responses **only after they resolved**. So when many components requested the same URL before the first response landed (e.g. every column row's Data Type dropdown mounting at once on a wide table's Columns tab), each one saw an empty cache and fired its own identical `GET`. A 150-column table produced ~150 concurrent duplicate `get_types` requests.

## Fix

Added in-flight request sharing:

- A module-level `Map` tracks requests currently in progress, keyed by the resolved full URL + query params.
- On a cache miss, a caller reuses an existing in-flight request if one matches; otherwise it starts one and stores it.
- The first request to resolve populates the cache exactly as before.
- The map entry is removed on settle (`.finally()`), success or failure, so it doesn't leak.

Result: one shared `get_types` request for all concurrent callers instead of one per row. Cached behavior and response-shape handling are unchanged.

## Notes

Could not run the JS linter locally — the repo's `.eslintrc.js` (legacy format) is incompatible with the installed ESLint v9 (flat-config only), a pre-existing toolchain issue unrelated to this change. The file was verified to parse as valid ES module syntax.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate concurrent GET requests when multiple parts of the application request the same resource with identical parameters.
  * Shared a single in-progress request among concurrent callers and ensured it is properly cleared after the request completes or fails, without changing existing cache-hit behavior.

### DIFF
--- a/web/pgadmin/browser/static/js/node_ajax.js
+++ b/web/pgadmin/browser/static/js/node_ajax.js
@@ -8,7 +8,7 @@
 //////////////////////////////////////////////////////////////
 
 import _ from 'lodash';
-import getApiInstance from '../../../static/js/api_instance';
+import getApiInstance, {getInflight} from '../../../static/js/api_instance';
 import {generate_url} from 'sources/browser/generate_url';
 import pgAdmin from 'sources/pgadmin';
 
@@ -77,13 +77,6 @@ export function generateNodeUrl(treeNodeInfo, actionType, itemNodeData, withId, 
 }
 
 
-/* Tracks AJAX requests that are currently in flight, keyed by the resolved
- * URL + query params. This lets concurrent callers asking for the same options
- * (e.g. every row's Data Type dropdown mounting at once) share a single HTTP
- * request instead of each firing their own before the cache is populated.
- */
-const _inflightAjaxRequests = new Map();
-
 /* Get the nodes list as options required by select controls
  * The options are cached for performance reasons.
  */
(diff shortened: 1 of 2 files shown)


web/pgadmin/browser/static/js/node_ajax.js
```diff
@@ -114,15 +121,28 @@ export function getNodeAjaxOptions(url, nodeObj, treeNodeInfo, itemNodeData, par
...
-          data = res.data;
-          if(res.data.data) {
-            data = res.data.data;
-          }
-          otherParams.useCache && cacheNode.cache(nodeObj.type + '#' + url, treeNodeInfo, cacheLevel, data);
-          resolve(transform(data));
+        // Share a single in-flight request among all concurrent callers asking
+        // for the same URL + params, so we don't fire duplicate GETs before the
+        // first response lands and populates the cache.
+        let inflightKey = fullUrl + '#' + JSON.stringify(otherParams.urlParams || {});
```
mzabuawala: Some observations - 
1) Not the best way to create a unique key
```
{
   a:1,
   b:2
}
```
vs
```
{
   b:2,
   a:1
}
```
And a nested object will make it worse.

2) A better place to put such logic is inside the Axios wrapper, so that it will be used by the entire app

kundansable:
@hiteshjambhale fix is working, we only get one API call for column details instead of N(number of columns).
just fix the changes asked by @mzabuawala

PR #10171, opened 2026-07-22
fix: exclude non-editable/alias columns from Query Tool row UPDATE

### LINKED ISSUE #10103: Editing a row with expression/alias columns in Query Tool fails to save with "column does not exist" error
### Describe the bug

If your query includes a calculated or aliased column — for example first_name || ' ' || last_name AS the_name — and you edit any cell on an existing row and hit Save, pgAdmin throws an error. It tries to update the alias column (the_name) on the table, which doesn't actually exist as a real column. The column header shows a lock icon, so pgAdmin already knows it's not editable — but it still includes it when building the UPDATE.

### To Reproduce

Open the Query Tool on any database where you can create tables.
Run this to set up a test table with some data:

DROP TABLE IF EXISTS some_table;
CREATE TABLE some_table (id INT PRIMARY KEY, first_name TEXT, last_name TEXT);
INSERT INTO some_table VALUES (1, 'John', 'Doe');

SELECT id, first_name, last_name,
       first_name || ' ' || last_name AS the_name
FROM some_table;
In the result grid, click on an editable cell — for example, change first_name from John to Jane.
Press F5 to save.
You'll see an error in the result panel.

### Expected behavior

The save should succeed. Only real table columns (id, first_name, last_name) should be updated. The alias column the_name should be ignored completely.

### Error message

ERROR: column "the_name" of relation "some_table" does not exist
LINE 2: the_name = $1::text WHERE

kundansable:
### Summary
Fixes #10103 — editing a row in the Query Tool result grid fails to save with `column "..." does not exist` when the query includes an aliased/computed column (e.g. `first_name || ' ' || last_name AS the_name`), even though that column is already shown as read-only (lock icon) in the grid.

### Root Cause
`save_changed_data()` has two code paths — `added` (INSERT) and `updated` (UPDATE). The `added` path already filters the outgoing payload down to real, editable table columns (fixed for #9939 in #10015), using the same `columns_info[...]['is_editable']` metadata that drives the grid's lock icon. The `updated` path never got the equivalent filter, so an edited row that includes an alias/expression column sends that alias straight into the generated `UPDATE ... SET` clause — and Postgres rejects it because there's no such real column.

### Fix
Apply the same editable-columns guard to the `updated` branch: drop any key from the changed-row payload that isn't a real, editable table column, and skip the row entirely if nothing editable remains after filtering (avoiding an empty/invalid `SET` clause).

### Test Steps
1. In the Query Tool, run:
   ```sql
   DROP TABLE IF EXISTS some_table;
   CREATE TABLE some_table (id INT PRIMARY KEY, first_name TEXT, last_name TEXT);
   INSERT INTO some_table VALUES (1, 'John', 'Doe');

   SELECT id, first_name, last_name,
          first_name || ' ' || last_name AS the_name
   FROM some_table;
   ```
2. In the result grid, edit `first_name` from `John` to `Jane` on the existing row.
3. Press F5 / Save.
4. **Expected:** save succeeds; only `first_name` is updated. No `column "the_name" ... does not exist` error.
5. Regression: repeat on a query with **no** alias/expression columns and confirm normal edits still save correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved saving of edited table data when query results include calculated, aliased, or otherwise non-editable columns.
  * Prevented invalid update operations by ignoring non-editable fields and skipping rows that have no editable values.
  * Prevented read-only fields from appearing as changed or being committed through the editor.
  * Updates now reliably reference the correct row identifier when applying changes, avoiding misaligned or malformed statements.

### DIFF
--- a/web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/QueryToolDataGrid/Editors.jsx
@@ -241,6 +241,15 @@ export function TextEditor({row, column, onRowChange, onClose}) {
   }, []);
 
   const onOK = ()=>{
+    /* The editor also opens on read-only columns so that wide values can be
+     * inspected, but committing must be refused there: an expression/alias
+     * column has no counterpart in the base table and the generated UPDATE
+     * would fail with `column "..." does not exist` (#10103). The OK button
+     * is already hidden, this guards the Enter key path. */
+    if(!column.can_edit) {
+      onClose(false);
+      return;
+    }
     if(column.is_array && !isValidArray(localVal)) {
       pgAdmin.Browser.notifier.error(gettext('Arrays must start with "{" and end with "}"'));
     } else {
--- a/web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
+++ b/web/pgadmin/tools/sqleditor/static/js/components/sections/ResultSet.jsx
@@ -1621,6 +1621,15 @@ export function ResultSet() {
 
 
   const onRowsChange = (newRows, otherInfo)=>{
+    /* Never record, or even display, a change to a read-only column. The
+     * editors open on such columns so that wide values can be inspected, so a
+     * stray commit can still arrive here; letting it through would stage an
+     * expression/alias column that does not exist in the base table and the
+     * save would fail with `column "..." does not exist` (#10103). */
+    if(otherInfo.column?.can_edit === false) {
+      return;
+    }
+
     let row = newRows[otherInfo.indexes[0]];
     let clientPK = rowKeyGetter(row);
 
--- a/web/pgadmin/tools/sqleditor/utils/save_changed_data.py
+++ b/web/pgadmin/tools/sqleditor/utils/save_changed_data.py
@@ -177,6 +177,21 @@ def save_changed_data(changed_data, columns_info, conn, command_obj,
             list_of_sql[of_type] = []
             for each_row in changed_data[of_type]:
                 data = changed_data[of_type][each_row]['data']
+                # Drop any column that isn't a real editable column of the
+                # underlying table (e.g. an expression/alias column such as
+                # `first_name || ' ' || last_name as the_name`). Such columns
+                # carry the read-only lock icon in the grid, but without this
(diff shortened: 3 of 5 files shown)


asheshv:
**Consistency review: added-row vs updated-row column filtering**

Nice fix for #10103 — filtering non-editable columns out of `data` before rendering the SQL template stops the `UPDATE` from referencing expression/alias columns.

One thing worth double-checking: the **added-row** path explicitly strips the tracking keys *before* filtering:

```python
# lines 113-115
data.pop(client_primary_key, None)
data.pop('is_row_copied', None)
...
data = {
    k: v for k, v in data.items()
    if k in columns_info and columns_info[k].get('is_editable', True)
}
```

The **updated-row** path (this PR, lines 179-183) relies solely on the `k in columns_info` filter to drop them:

```python
data = changed_data[of_type][each_row]['data']
data = {
    k: v for k, v in data.items()
    if k in columns_info and columns_info[k].get('is_editable', True)
}
```

I traced `client_primary_key`/`is_row_copied` on the frontend — they're only injected into the `added` payload (`ResultSet.jsx:759` calls `setClientPK` for new rows), so today the updated-row `data` dict never actually contains them, and the `columns_info` filter alone is sufficient. So I don't think this is a live bug right now.

Two things I'd still flag:

1. **Defense-in-depth / consistency** — if a future change ever starts tagging updated rows the same way (e.g. multi-row copy-paste into existing rows), this path would silently rely on `columns_info` to save it, instead of being explicit like the added-row path. Mirroring the explicit `.pop()` calls here would make both code paths symmetric and remove the implicit assumption.
2. **`row_id` after filtering** (line 197-200):
   ```python
   list_of_sql[of_type].append({'sql': sql,
                                'data': data,
                                'row_id':
                                    data.get(client_primary_key)})
   ```
   Since `data` here is now the *filtered* dict, `data.get(client_primary_key)` will always be `None` for updated rows (as it presumably already was, since `client_primary_key` was never in the updated payload to begin with — so no regression, just flagging that this line is dead weight / possibly should read from the unfiltered `changed_data[...]['data']` if `row_id` was ever meant to be populated here).

Not blocking, just wanted these on record in case someone hits it while extending this code path later.

kundansable:
closing this
as there is already duplicate PR for this #10103

PR #10174, opened 2026-07-22
fix: don't hard-fail on an undecryptable saved password

### LINKED ISSUE #10140: OIDC users intermittently fail with "Failed to decrypt the saved password" and cannot be recovered by recreating the user
## Describe the bug

Some (but not all) users consistently experience frontend errors after logging into pgAdmin using OIDC/OAuth2.

The backend logs contain errors similar to:

```text
2026-07-03 11:00:39,895: ERROR pgadmin: Failed to decrypt the saved password.
Error: 'utf-8' codec can't decode byte 0xc4 in position 0: invalid continuation byte
```

The failing byte is not always `0xc4`; it varies between affected users, but is consistently a high-value byte (`>= 0x80`).

The issue persists for affected users even after deleting the user from the pgAdmin configuration database and allowing them to be recreated via OIDC. Other users on the same deployment continue to work normally.

## To Reproduce

Unfortunately, we have not found a deterministic way to reproduce the issue.

Observed behavior:

1. User logs into pgAdmin using OIDC.
2. The account is auto-created (`OAUTH2_AUTO_CREATE_USER = True`).
3. Some users eventually begin experiencing frontend issues.
4. The backend logs `Failed to decrypt the saved password` with UTF-8 decoding errors.
5. The affected user remains broken even after deleting and recreating the pgAdmin user.

The issue only affects a subset of users.

## Expected behavior

If pgAdmin encounters an encrypted password that cannot be decrypted, I would expect either:

- the corrupted saved password to be discarded and the user prompted again, or
- the invalid saved password entry to be ignored,

rather than leaving the user in a permanently broken state.

## Error message

Example:

```text
2026-07-03 11:00:39,895: ERROR pgadmin: Failed to decrypt the saved password.
Error: 'utf-8' codec can't decode byte 0xc4 in position 0: invalid continuation byte
```

Other affected users report different byte values, for example:

```text
Error: 'utf-8' codec can't decode byte 0xb9 in position 1: invalid start byte
```

The byte value changes, but it is always a relatively high byte rather than valid UTF-8.

## Screenshots

None available.

## Desktop (please complete the following information)

- **OS:** Kubernetes (Linux container)
- **pgAdmin version:** 9.16
- **Mode:** Server
- **Browser:** Chrome and Firefox (same behavior)
- **Package type:** Official Container image (`sha256:40fa840c5bb7c8463957f1255b01283732c2d8c9396a956d180f8e6c296753b3`)

## Additional context

- Single pgAdmin pod (not a multi-replica deployment).
- pgAdmin is configured to use an external PostgreSQL database as its configuration database (`CONFIG_DATABASE_URI`) instead of the default embedded SQLite database.
- Authentication is via OIDC/OAuth2 with `OAUTH2_AUTO_CREATE_USER = True`.
- The problem is user-specific rather than deployment-wide.
- Other users on the same deployment continue to work normally.
- Clearing the browser cache does not help.
- Using a different browser does not help.
- Restarting the pgAdmin pod does not help.
- Deleting the affected pgAdmin user from the configuration database and allowing it to be recreated via OIDC does not help.
- The issue also persists after changing both the master password and the PostgreSQL server password to newly generated passwords containing only simple ASCII letters (`a-z`), making it unlikely that special characters or Unicode passwords are contributing to the problem.
- Since the affected user can be completely recreated and the issue still persists, it appears that either some user-specific state is not being removed or the decryption failure has another underlying cause.

If useful, we can provide our OAuth2 configuration (minus secrets), additional logs, or information about our PostgreSQL-backed pgAdmin configuration database.

kundansable:
### Summary
Fixes #10140 — some OIDC/OAuth2 users intermittently get a hard `Failed to decrypt the saved password` error on every connection attempt, and the account stays broken even after being deleted and recreated via OIDC.

### Root Cause
`Connection._decode_password()` calls `decrypt()` (AES-CFB8, unauthenticated) on the stored password ciphertext and lets any `UnicodeDecodeError` from a failed decrypt propagate as a hard error. For OIDC/OAuth2 logins, the crypt key used to encrypt a saved password can differ between sessions; if it does, the ciphertext saved under a previous key can **never** be decoded correctly under the current key, and every connection attempt hits this same decode failure — surfacing the scary, unrecoverable-looking "Failed to decrypt the saved password" error. Deleting and recreating the pgAdmin user doesn't help because the problem is in the stored ciphertext for the server record, not the user record.

### Fix
Catch the decode failure in `_decode_password()`, log a warning, and return an empty/no-password result instead of propagating the exception — i.e. treat an undecryptable saved password the same as "no saved password was ever set." `connect()` then falls through to its normal password-prompt path, so the user is prompted for and can re-enter the correct password, making the account recoverable instead of permanently stuck. OIDC/OAuth2 crypt-key derivation itself is untouched by this change.

### Test Steps
*(Exact repro requires an OIDC/OAuth2 setup where the crypt key can rotate between sessions; a close approximation can be done by directly corrupting a stored password's ciphertext in the config DB for a test server.)*
1. Set up a server with a saved password (Save Password enabled).
2. Simulate a crypt-key change or corrupt the stored `password` ciphertext for that server record so it no longer decrypts under the current key.
3. Attempt to connect to that server.
4. **Before the fix:** connection attempt fails immediately with "Failed to decrypt the saved password", with no path to recovery.
5. **After the fix:** the bad saved password is silently discarded; pgAdmin instead prompts for the password normally, and entering the correct password connects successfully.
6. Regression: confirm a server with a valid, correctly-encrypted saved password still connects automatically without any prompt (no over-broad fallback).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection handling when a saved password cannot be decrypted.
  * Invalid or unusable saved passwords are now safely ignored and cleared.
  * Instead of failing immediately, the connection process now prompts you to enter your password again.
  * Improved execution reliability for operations that require single-statement processing, including retry scenarios.

### DIFF
--- a/web/pgadmin/utils/driver/psycopg3/connection.py
+++ b/web/pgadmin/utils/driver/psycopg3/connection.py
@@ -271,6 +271,13 @@ class Connection(BaseConnection):
                     'decrypted. The user will be prompted for the password. '
                     'Error: {0}'.format(str(e))
                 )
+                # Clear the cached ciphertext so it isn't silently reused
+                # on the next connect() attempt, which would just hit
+                # this same decode failure again, and (since a stale but
+                # still-truthy encpass would otherwise remain cached)
+                # skip the passexec fallback below.
+                self.password = None
+                manager.password = None
                 return False, '', None
         return False, '', password
 
--- a/web/pgadmin/utils/driver/psycopg3/connection.py
+++ b/web/pgadmin/utils/driver/psycopg3/connection.py
@@ -258,12 +258,20 @@ class Connection(BaseConnection):
                 if isinstance(password, bytes):
                     password = password.decode()
             except Exception as e:
-                manager.stop_ssh_tunnel()
-                current_app.logger.exception(e)
-                return True, \
-                    _(
-                        "Failed to decrypt the saved password.\nError: {0}"
-                    ).format(str(e)), password
+                # The saved password could not be decrypted. This happens
+                # when the stored ciphertext was encrypted with a different
+                # key (e.g. OIDC/OAuth2 logins where the derived encryption
+                # key changed between sessions), leaving un-decodable bytes
+                # (typically a "'utf-8' codec can't decode byte 0x.." error).
+                # Instead of failing every connection attempt permanently,
+                # discard the bad saved password and continue so the user is
+                # prompted for the password again.
+                current_app.logger.warning(
+                    'Ignoring the saved password as it could not be '
+                    'decrypted. The user will be prompted for the password. '
+                    'Error: {0}'.format(str(e))
+                )
+                return False, '', None
         return False, '', password
 
     def connect(self, **kwargs):


dpage (changes requested):
Thanks for picking this one up. The direction is right: turning an undecryptable saved password into a warning plus the normal prompt path is exactly what the issue asks for, and the change is pep8-clean and nicely contained. I do have a few things I would like addressed before this goes in, plus one observation about the root cause that I think matters for how we close #10140.

### Should fix

**1. The blanket `except Exception` is now silent, so real bugs disappear.** (`connection.py:260`) The catch was already broad, but it used `logger.exception` and surfaced an error, so a `TypeError` from a bad crypt key, a `binascii.Error` from a truncated column, or any future breakage in key management stayed visible. After this change all of those become a one-line warning with no traceback, and the user just sees an unexplained password prompt on every connect. Could we narrow the catch to the failures actually being targeted (`UnicodeDecodeError`, `binascii.Error`, `ValueError`) and let anything else keep the old behaviour, or at the very least log with `exc_info=True` so the traceback survives?

**2. Clearing `self.password`/`manager.password` does not clear the local `encpass` in `connect()`, and that leaks in two ways.** (`connection.py:274` versus `connection.py:310-313` and `connection.py:425`)

* At the end of `connect()`, `if status and is_update_password: manager._update_password(encpass)` re-caches the bad ciphertext into the manager and into every open connection (`server_manager.py:512-517`), whenever the connection nonetheless succeeds without a password: trust or peer auth, a `.pgpass`/passfile, a service definition, or Kerberos. That silently undoes the clearing and re-arms the remaining hard-fail sites in point 4.
* The `passexec` fallback at `connection.py:319` (`if not password and not encpass and manager.passexec`) is still skipped on this attempt, because `encpass` is unchanged. That is the half of CodeRabbit's open thread which still stands.

Both close cleanly if `_decode_password()` signals "discard the ciphertext" back to the caller (a fourth return value, or returning `encpass` alongside `password`) and `connect()` then sets `encpass = None`.

**3. The comment overstates what the clearing achieves.** (`connection.py:270-273`) The claim that the ciphertext "isn't silently reused on the next `connect()` attempt" holds only for the no-argument reconnect paths (`_restore_connections()`). The main connect route calls `manager.update(server)` when the server is not connected, and `update()` unconditionally re-populates `self.password = server.password` from the config DB (`server_manager.py:78`), after which the route passes `conn_passwd or server.password` straight back in (`browser/server_groups/servers/__init__.py:1673-1682`). The behaviour is still fine, since the user gets prompted either way, but the comment should describe what the clearing really buys.

**4. Sibling decrypt sites with the same root cause still hard-fail, so an affected user is only partly unstuck.** Once a user's stored ciphertext is undecryptable under the current key they will still hit an uncaught `UnicodeDecodeError` in `ServerManager.export_password_env()` (`server_manager.py:548`, used by Backup, Restore and Import/Export), and `Connection._decrypt_password()` (`connection.py:1464`, used by `reset()`) has the same uncaught `.decode()`, masked only for as long as the in-memory clearing survives, which per point 3 is not long. The SSH tunnel password path (`server_manager.py:565-579`) still returns the analogous "Failed to decrypt the SSH tunnel password" hard error. The tunnel path at least deserves the same treatment here, since a key change breaks both passwords identically.

**5. For SSH-tunnel servers the prompt this fix relies on may never appear.** (credit to CodeRabbit, this half of the thread is valid) On failure the route calls `get_response_for_password(server, 401, not server.save_password, ...)` (`servers/__init__.py:1731-1733`), and for a server with `save_password = 1` that third argument is `False`. The non-tunnel branch hardcodes `"prompt_password": True` (`servers/__init__.py:2219`) so it is fine, but the `use_ssh_tunnel` branch passes the argument through (`servers/__init__.py:2198`), so an SSH-tunnel server with a saved-but-undecryptable password gets a dialog with no server-password field and stays stuck. It is pre-existing, but it defeats the recovery path this PR is built on, so I think it belongs in scope.

### On the root cause

The description says the OIDC crypt key "can differ between sessions" without saying why, and I think it is worth recording, because it changes how we should close the issue. Two facts from the code: `authenticate/oauth2.py:690` sets `session['pass_enc_key'] = session['oauth2_token']['access_token']`, so the key protecting every saved server password is the OAuth2 access token, and `utils/crypto.py:pad()` truncates any key to its first 32 bytes. The rest is inference rather than something I can prove about the reporter's deployment: for a JWT access token those first 32 characters are the base64url-encoded header prefix, constant as long as the header content and field ordering are, which is why it works for most people; any deployment or moment where those bytes vary (a `kid` appearing early in the header and rotating, a different `alg`, or an IdP issuing opaque non-JWT tokens) produces a different AES key and permanently undecryptable ciphertext. That fits the reported pattern of a subset of users, no deterministic repro, and survival across user recreation.

So I am happy to take this as the mitigation, but affected users will keep being re-prompted whenever the token prefix shifts, and the real fix is a stable KEK for OAuth2 sessions (`MASTER_PASSWORD_HOOK`, the OS keyring, or a per-user key wrapped at first login). I would rather we opened a separate issue for the key derivation than closed #10140 as fully solved on the back of this.

### Minor

* After the password is discarded the user sees libpq's `fe_sendauth: no password supplied` in the prompt dialog, which is fairly opaque. The route already relays `errmsg` to the dialog, so a short explanatory message ("the saved password could not be decrypted, please re-enter it") would make the recovery obvious.
* No test accompanies the change. `web/pgadmin/utils/driver/psycopg3/tests/` currently holds only `__init__.py`, so it would be new ground, but a unit test patching `decrypt` to raise and asserting `_decode_password()` returns `(False, '', None)` whilst clearing both caches would be cheap and would pin down the contract that points 2 and 3 are about. Not a blocker.
* Dropping `manager.stop_ssh_tunnel()` from this path is correct now that we continue and still need the tunnel; noting it only so it is on the record as deliberate.

PR #10178, opened 2026-07-23
fix: persist rotated saved password and correct manager session state

### LINKED ISSUE #10128: Update of saved passwords fails in Query Tool on 9-16

**Describe the bug**
For Servers stored in pgadmin with a saved password, the Query Tool is unreachable if stored password is invalid (=after Password of the user changed), although you might enter the correct password when prompted. The saved password will overwrite the entered password and will not update.

**To Reproduce**
We will take a shortcut - not rotating user credentials but manually invalidating stored credentials in connection details. The effect is the same.

1. Create a database connection with valid credentials. No storing of password (will remove the additional step of clear stored password that would have to be performed additionally)

Image

2. Once connected to the server, disconnect

Image

3. Change password to something non-sense in connection settings, enable Save Password checkbox

Image

Image

4. Connect to Server

Image

5. Enter correct password, when prompted

Image

6. Connection established successfully

Image

7. Open Query Tool for this Connection

Image

8. Enter correct password, when prompted. This continues forever. You will not be able to access Query Tool

Image

**Expected behavior**

Saved Passwords are updated, when Checkbox "Save Password" is active and new, valid Credentials are provided

**Error message**

FATAL: LDAP authentication failed for user xyz
FATAL: no pg_hba.conf entry for host "a.b.c.d", user "xyz", database "postgres", no encryption

 - OS: Linux
 - pgAdmin version: 9-16
 - Mode: Server
 - Browser (if running in server mode): edge
 - Package type: Container, Openshift / Helm

kundansable:
For a server configured to save its password, entering a rotated password at the reconnect/password-prompt dialog appeared to work for the current connection but never actually replaced the stale saved ciphertext, causing an infinite re-prompt loop on every subsequent connection (e.g. opening the Query Tool).

Two gaps caused this:

1. Query Tool's "already connected" reconnect path (sqleditor.connect_server) cached the freshly entered password on the in-memory server manager only, never writing it back to the stored server record.

2. The main "Connect to Server" flow (ServerNode.connect()) only persisted a freshly entered password to the server's stored ciphertext when the current request's save_password flag was true. But the password-prompt dialog shown on a failed connect doesn't resend the server's existing save_password setting -- it only reports its own checkbox state, which defaults to unchecked -- so a server already configured to save its password never got its stale ciphertext replaced. Separately, the driver's in-memory manager.password fix was never persisted via manager.update_session(); Driver.managers is an in-process cache, so in a multi-worker deployment (e.g. OpenShift/Helm) the next request can land on a different worker, which restores the stale pre-fix manager from the session and loses the corrected password.

Fix all three: persist the new password to the (owned or shared) server record from the Query Tool reconnect path when save_password is set and allowed; treat save_password as true in the main connect flow whenever the server already has it enabled, not just when the current request's flag says so; and call manager.update_session() after a successful connect that used a freshly entered password.

Fixes #10128

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed repeated password prompts caused by outdated saved-password settings.
  * Preserved corrected passwords across requests and sessions after successful connections.
  * Saved newly entered passwords when enabled, including shared server connections.
  * Ensured the save-password option reflects the server’s current setting and correctly records when it is unchecked.
  * Improved password preference handling across supported connection flows.
  * Added more reliable handling when entered passwords are invalid or cannot be saved.
  * Improved shared-server connection handling when using shared login credentials.

### DIFF
--- a/web/pgadmin/browser/server_groups/servers/__init__.py
+++ b/web/pgadmin/browser/server_groups/servers/__init__.py
@@ -1685,8 +1685,16 @@ class ServerNode(PGChildNodeView):
                 password = conn_passwd or server.password
         else:
             password = data['password'] if 'password' in data else None
-            save_password = data['save_password']\
-                if 'save_password' in data else False
+            # The password-prompt dialog doesn't resend the server's
+            # existing save_password setting, only its own checkbox
+            # state. If the server is already configured to save its
+            # password, a freshly entered replacement should still be
+            # persisted, or the stale saved password is never
+            # replaced. See issue #10128.
+            save_password = (
+                data['save_password'] if 'save_password' in data
+                else False
+            ) or bool(server.save_password)
 
             try:
                 # Encrypt the password before saving with user's login
@@ -1755,6 +1763,16 @@ class ServerNode(PGChildNodeView):
                     conn = None
 
                     return internal_server_error(errormsg=str(e))
+
+            # Persist the manager's corrected in-memory password (set by
+            # the driver on a successful connect) into the Flask session.
+            # Without this, a fresh worker process handling the next
+            # request (e.g. opening the Query Tool, in a multi-worker
+            # deployment) restores the stale pre-fix manager state from
+            # the session and the new password is lost, re-triggering the
+            # password prompt indefinitely. See issue #10128.
+            if password:
+                manager.update_session()
 
             if save_tunnel_password and config.ALLOW_SAVE_TUNNEL_PASSWORD:
                 try:
--- a/web/pgadmin/browser/server_groups/servers/__init__.py
+++ b/web/pgadmin/browser/server_groups/servers/__init__.py
@@ -1750,6 +1750,12 @@ class ServerNode(PGChildNodeView):
                     # 1 is True in SQLite as no boolean type
                     if _is_non_owner(server):
                         setattr(shared_server, 'save_password', 1)
+                        # `server` is a detached overlay (see
+                        # get_shared_server_properties) built before this
+                        # write, so it won't pick up the SharedServer
+                        # change on its own -- keep it in sync since the
+                        # connect response below reports its state.
+                        server.save_password = 1
                     else:
                         setattr(server, 'save_password', 1)
 
@@ -1777,6 +1783,9 @@ class ServerNode(PGChildNodeView):
                     if _is_non_owner(server):
                         setattr(shared_server, 'save_password', 0)
                         setattr(shared_server, 'password', None)
+                        # Keep the detached overlay in sync -- see the
+                        # comment in the save_password branch above.
+                        server.save_password = 0
                     else:
                         setattr(server, 'save_password', 0)
                         setattr(server, 'password', None)
--- a/web/pgadmin/browser/server_groups/servers/__init__.py
+++ b/web/pgadmin/browser/server_groups/servers/__init__.py
@@ -1618,6 +1618,12 @@ class ServerNode(PGChildNodeView):
         passfile = None
         tunnel_password = None
         save_password = False
+        # Distinguishes "the caller explicitly said false" from "the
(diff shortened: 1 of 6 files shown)


asheshv:
I traced this one carefully because it touches stored credentials. Good news first: **there's no cross-user leak.** `get_shared_server_properties()` (`servers/__init__.py:170-234`) overlays the caller's own SharedServer row and expunges the owner's `Server` from the session, so `save_password` is read from the caller's row and the write lands on `shared_server`. The right encryption key is used (`get_crypt_key()`, the current user's), nothing is written or logged in plaintext, and `manager.update_session()` stores the encrypted form, not the raw password.

That said, I don't think this is mergeable yet.

**1. The main fix doesn't actually fix the described bug (`servers/__init__.py:1775`) — and is redundant when it does run.**
The new `if password: manager.update_session()`. Here's the sequence:

- `Connection.connect()` → `_initialize()` → `_set_server_type_and_password()` sets `manager.password = kwargs['password']`
- `_initialize` then already calls `manager.update_session()` at `connection.py:624`
- only *after* that does `connection.py:422` run `manager._update_password(encpass)` — the correction that never gets persisted

That correction differs from what's already in the session **only** when `kwargs['password']` was falsy, i.e. when `encpass` came from the `self.password or getattr(manager, 'password', None)` fallback at `connection.py:299`. And in exactly that case the new guard `if password:` is False, so `update_session()` is never reached. When `password` *is* truthy, line 624 already wrote the same value.

The fix belongs one level down, in the driver: make `connection.py:422-423` read `manager._update_password(encpass); manager.update_session()`. That repairs every caller — sqleditor, backup, debugger, schema diff — instead of just `ServerNode.connect`, and the hunk at 1775 can go.

**2. A wrong password now gets saved permanently (`tools/sqleditor/__init__.py:2780-2797`).**
`_persist_saved_password` writes to durable storage without ever checking the password against the server. `connect_server` only reaches this code because `conn.connected()` is already `True` (line 2704) from a connection that existed beforehand — the password the user just typed is never used in a `conn.connect()`. So a typo at the prompt overwrites a working stored credential, and every connection after that fails until the user notices and fixes it by hand. Compare `ServerNode.connect`, which persists only inside the post-`conn.connect()` success branch. Please validate first (throwaway connection with the supplied password), or defer persisting until the next real connect.

**3. Unchecking "Save Password" no longer works (`servers/__init__.py:1694-1697`).**
The `or bool(server.save_password)` overrides an explicit user choice. `ConnectServerContent.jsx:29` defaults `save_password` to `false` and `:122-123` only appends the field when it's true — so the backend genuinely cannot tell "the user unticked the box" apart from "the dialog didn't send the field". Result: someone who deliberately unticks Save Password on a save-enabled server gets it saved anyway. This can't be repaired server-side; the dialog has to seed the checkbox from the server's current setting (the 401/428 payload already carries `allow_save_password` at `ConnectServerContent.jsx:101` — add `save_password`) and always send the field.

**4. No tests.** `gh pr diff --name-only` shows two source files and no test file, on a change that alters *when* a credential is written and *which row* it lands on. `connect_server` has no coverage at all today. At minimum: a server-mode test asserting a non-owner's rotated password lands on their `SharedServer` row and leaves the owner's `Server.password` untouched, plus one for `save_password=0`.

Smaller items:

- **`:2797`** — `db.session.commit()` with no `rollback()` on failure; the outer `except` at 2776 swallows the error and leaves the session pending-rollback for the rest of the request.
- **`:2773`** — the truthiness list `('true', 'True', '1', 1, True)` misses `'TRUE'`, `'on'`, `'yes'`. `str(save_password).lower() in ('true', '1', 'on', 'yes')` is safer.
- **`:2788-2796`** — reimplements `_is_non_owner()` (`servers/__init__.py:62-65`) and the owner/shared branch inline. Import and reuse it; two copies of a credential-routing rule will drift apart.

PR #10216, opened 2026-07-29
 feat: add GLM (Z.ai) LLM provider

shawonis08:
Adds GLM (Z.ai) as an LLM provider for pgAdmin's AI features, alongside
Anthropic, OpenAI, Ollama and Docker Model Runner. GLM exposes an
OpenAI-compatible API, so `GLMClient` subclasses `OpenAIClient` and
overrides only where Z.ai diverges.

## Configuration

- `GLM_API_URL` — defaults to `https://api.z.ai/api/paas/v4`
- `GLM_API_KEY_FILE` — defaults to `~/.zai-api-key`
- `GLM_API_MODEL` — no default (see below)

`ALLOWED_LLM_API_URLS` gains `https://api.z.ai:443`. 
Allowlist matching is scheme+host+port, so the single `api.z.ai` entry also covers Coding Plan subscriptions at
`https://api.z.ai/api/coding/paas/v4`.

## Testing

Verified against the live Z.ai API: streaming text, tool calling, usage
reporting, the no-model guard, and the reasoning-only path.

Unit tests cover URL allowlisting for all three platforms (including the
Coding Plan path), preference resolution and rejection, the model-refresh
endpoints, and LLM status reporting.

Docs updated in `ai_tools.rst` and `preferences.rst`.

## Summary by CodeRabbit

* **New Features**
  * Added GLM (Z.ai) as a supported AI provider.
  * Added GLM configuration for API URL, API key file, and model selection.
  * Added model discovery and refresh capabilities for GLM.
  * Added validation for GLM URLs and API key file paths.
  * GLM can now be selected as the default AI provider.

* **Documentation**
  * Updated AI Reports, Preferences, and LLM configuration documentation with GLM setup and troubleshooting guidance.

### DIFF
--- a/docs/en_US/ai_tools.rst
+++ b/docs/en_US/ai_tools.rst
@@ -65,9 +65,9 @@ Select your preferred LLM provider from the dropdown:
 **GLM (Z.ai)**
   Use GLM models from Z.ai through the provider's OpenAI-compatible API.
 
-  * **API URL**: Custom API endpoint URL (leave empty for default: https://open.bigmodel.cn/api/paas/v4).
+  * **API URL**: Custom API endpoint URL (leave empty for default: https://api.z.ai/api/paas/v4). Use https://api.z.ai/api/coding/paas/v4 with a Z.ai Coding Plan subscription.
   * **API Key File**: Path to a file containing your Z.ai API key. This path refers to the filesystem where the pgAdmin server is running (e.g., inside the container if using Docker). The ``~`` prefix is expanded to the home directory of the user running the pgAdmin server process.
-  * **Model**: Select from available GLM models (e.g., glm-4.5).
+  * **Model**: Select from available GLM models (e.g., glm-4.7). There is no default; a model must be selected before the AI features can be used.
 
 **Ollama**
   Use locally-hosted open-source models via Ollama. Requires a running Ollama instance.
--- a/docs/en_US/ai_tools.rst
+++ b/docs/en_US/ai_tools.rst
@@ -26,7 +26,7 @@ Before using AI Reports, you must:
 
 **Note:**
 
- * AI Reports using cloud providers (Anthropic, OpenAI) require an active internet connection.
+ * AI Reports using cloud providers (Anthropic, OpenAI, GLM) require an active internet connection.
    Local providers (Ollama, Docker Model Runner) do not require internet access.
 
  * API usage may incur costs depending on your LLM provider's pricing model.
@@ -61,6 +61,13 @@ Select your preferred LLM provider from the dropdown:
   * **API URL**: Custom API endpoint URL (leave empty for default: https://api.openai.com/v1). Include the ``/v1`` path prefix if required by your provider.
   * **API Key File**: Path to a file containing your OpenAI API key (obtain from https://platform.openai.com/). This path refers to the filesystem where the pgAdmin server is running (e.g., inside the container if using Docker). The ``~`` prefix is expanded to the home directory of the user running the pgAdmin server process. Optional when using a custom URL with a provider that does not require authentication.
   * **Model**: Select from available GPT models (e.g., gpt-4).
+
+**GLM (Z.ai)**
+  Use GLM models from Z.ai through the provider's OpenAI-compatible API.
+
+  * **API URL**: Custom API endpoint URL (leave empty for default: https://open.bigmodel.cn/api/paas/v4).
+  * **API Key File**: Path to a file containing your Z.ai API key. This path refers to the filesystem where the pgAdmin server is running (e.g., inside the container if using Docker). The ``~`` prefix is expanded to the home directory of the user running the pgAdmin server process.
+  * **Model**: Select from available GLM models (e.g., glm-4.5).
 
 **Ollama**
   Use locally-hosted open-source models via Ollama. Requires a running Ollama instance.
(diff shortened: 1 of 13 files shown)


dpage:
Thanks for this, and apologies for the slow response. The work here is careful:
the no-default-model guard is exactly the right instinct, and I appreciate that
the comments explain *why* each override exists rather than restating what the
code does. I've read it in full and run the tests locally, rebased onto current
`master`, where `pycodestyle` is clean and `runtests.py --pkg llm` gives 199
passed with no failures, so the ten new tests all do what they claim.

I'd like to take it in a different direction, though, and I think the change
gets smaller rather than larger as a result.

## The two overrides aren't GLM-specific

`GLMClient` overrides precisely two things, and neither is really about Z.ai:

* `max_tokens` instead of `max_completion_tokens`, because `max_completion_tokens`
  is an OpenAI-ism that most compatible servers either ignore or reject.
* Never falling back to the Responses API, because `/responses` is an OpenAI
  endpoint that third-party servers don't implement.

Both of those are true of every OpenAI-compatible endpoint we already tell
people to use, so LM Studio, LiteLLM and vLLM users are hitting the same two
problems today and getting no benefit from a GLM subclass. Your comment on
`_should_use_responses_api` actually diagnoses the general case rather well:
an unrelated error whose text happens to match the inherited heuristic switches
the client onto a path that cannot work and masks the real failure. That's a
bug in `OpenAIClient` for anyone off `api.openai.com`, not a GLM quirk.

So rather than a subclass per vendor, I'd prefer we confine the OpenAI-specific
behaviour to the case where we're actually talking to OpenAI. That fixes the
whole compatible-provider cohort in one go, and it means GLM then works through
the existing *OpenAI* provider with a custom API URL, with no new provider
entry, no new preferences, and no new default allowlist entry. I'd rather keep
`ALLOWED_LLM_API_URLS` as a small curated default and have users add whichever
host they actually use to `config_local.py`, instead of growing it by one line
per vendor.

## What that looks like

All in `web/pgadmin/llm/providers/openai.py`:

1. Add a small helper that decides whether we're on the official endpoint, by
   comparing `urlparse(self._base_url).hostname` against `api.openai.com`.
   Please use host equality rather than the `startswith` currently at line 99,
   which also matches something like `https://api.openai.com.example.net/v1`.
   It fails safe today, so it's not urgent on its own, but it's worth tidying
   whilst we're in there.
2. In `_build_chat_payload` (line 204), send `max_completion_tokens` when the
   helper says official and `max_tokens` otherwise. Sending only one of the two
   matters, since some servers validate strictly and reject unknown fields.
3. In `_should_use_responses_api` (line 178), return `False` unless official,
   which is your override generalised.
4. Reuse the helper in `is_available` (line 99).

For tests, the existing scenarios in `test_openai_stream.py` are the right home:
cases asserting which token-limit key ends up in the payload for an official
versus a custom base URL, and that a custom host never switches to the Responses
API even when the error text matches the heuristic. The GLM-specific scenarios in
`test_api_url_validation.py`, `test_llm_status.py` and
`test_model_refresh_endpoints.py` can all go, along with `providers/glm.py`, the
`glm` preferences and the `config.py` additions.

## One thing worth checking before we commit to documenting GLM at all

I'm not sure a real AI Report completes against a GLM model, and I'd value your
opinion since you have a working key.

Z.ai's chat-completions reference states that `thinking` defaults to enabled
across the GLM-4.5, 4.6 and 5.x line, and that GLM-4.7 in particular "will think
compulsorily". Reasoning arrives as `reasoning_content`, which pgAdmin never
reads: both `_read_openai_stream` and `_parse_response` look only at
`delta.content` and `message.content`. Meanwhile the report pipeline caps output
at `max_tokens=500` for the planning stage
(`web/pgadmin/llm/reports/pipeline.py:221`) and 1500 for section analysis
(line 294).

If reasoning tokens count against `max_tokens`, and Z.ai's docs don't say either
way, then planning comes back with empty content and `finish_reason: length`,
and `_raise_max_tokens_error` tells the user "The request is too large for model
glm-4.7. Try using a model with a larger context window", which is both the wrong
diagnosis and unactionable advice. Your PR description mentions verifying "the
reasoning-only path", so I suspect you've seen something in this area: what did
it actually do? If it is a problem, then `thinking: {'type': 'disabled'}` would
handle the 4.6 and 5.x models but not 4.7, and we'd need to raise those two
budgets instead, which is a separate change we should make on its own merits.

## Documentation

Don't worry about the data-handling side of the docs, as I'll write a
provider-neutral section covering what gets transmitted to a cloud provider and
under whose terms, which is broader than this change and something I'd rather
word myself.

What would help from you is adding GLM (Z.ai) to the existing lists of
OpenAI-compatible providers, at `docs/en_US/ai_tools.rst:78` and
`docs/en_US/preferences.rst:102`, alongside LiteLLM, LM Studio and EXO, together
with a sentence making clear that any such host has to be added to
`ALLOWED_LLM_API_URLS` by an administrator before pgAdmin will talk to it. That
last point is CodeRabbit's, and it's a fair one: at the moment a custom remote
URL is silently rejected with no hint that the fix lies in server configuration.

Whilst you're in `web/pgadmin/llm/README.md`, the `DEFAULT_LLM_PROVIDER` list
there still omits `docker`, and the Supported Providers list omits Docker Model
Runner. Both predate your PR but the diff already touches those lines, so it'd
be good to fold them in.

No release-notes entry is needed, incidentally: we populate those in a batch
shortly before release now, so please don't add one.

Finally, the branch is 51 commits behind `master`. It rebases cleanly, so a
`git rebase upstream/master` before the next push would help.

Happy to talk any of this through if you'd rather, and thanks again for taking
the time.

PR #10234, opened 2026-08-05
feat(oauth2): Map OAuth2 claims to Visible Server Groups

### LINKED ISSUE #9704: Map OAuth claims to Server Groups

With OAuth, it would be really useful to be able to assign access to Server Groups using OIDC claims. You could use this to give users different levels of permissions based on their assigned roles.

For example:
```python
# config.py
OAUTH2_SERVER_GROUP_CLAIM = 'pgadmin_server_groups'
```

Then with claim mappers you set:
```python
{"pgadmin_server_groups": ["RO Server 1", "RO Server 2", ...]}
```

which should give you access to that list of servers.

**Describe alternatives you've considered**

Reconfiguring servers on a per-user basis (tedious, copies of database credentials). Using a shared server group (not granular). 

Or https://github.com/pgadmin-org/pgadmin4/issues/9673 - but that seems much more involved than what I was thinking

Basically, I'm thinking something like this:
```python
    @staticmethod
    def get_all_server_groups():
        """
        Returns the list of server groups to show in server mode and
        if there is any shared server in the group.
        :return: server groups
        """
        pref = Preferences.module('browser')
        hide_shared_server = pref.preference('hide_shared_server').get()

        # lookup claims
        oauth2_allowed_server_groups = None
        if getattr(current_user, 'auth_source', None) == OAUTH2:
            oauth2_allowed_server_groups = session.get('oauth2_server_group_claims')
        has_claim_based_filter = isinstance(oauth2_allowed_server_groups, list)
        if has_claim_based_filter:
            oauth2_allowed_server_groups = set(oauth2_allowed_server_groups)

        server_groups = ServerGroup.query.all()
        groups = []
        for group in server_groups:
            is_allowed_group = (
                group.user_id == current_user.id or
                (not hide_shared_server and ServerGroupModule.has_shared_server(group.id)) or
                # also check claim
                (has_claim_based_filter and group.name in oauth2_allowed_server_groups)
            )
            if is_allowed_group:
                groups.append(group)
        return groups
```

beasteers:
### Summary

Control access to server groups using OAuth2 token claims, letting you create different levels of access tied to centralized SSO roles. 

Fixes https://github.com/pgadmin-org/pgadmin4/issues/9704

### What changed

Two new per-provider config keys were added to `OAUTH2_CONFIG` in `web/config.py`:

- `OAUTH2_SERVER_GROUP_CLAIM` — the claim (from the ID token or userinfo profile) whose values are treated as server group names.
- `OAUTH2_SERVER_GROUP_CLAIM_MAPPING` — optional dict mapping a claim value → one server group name or a list of server group names, useful when the claim carries opaque codes (e.g. `readonly`) that don't match group names directly.

### How it works

In `web/pgadmin/authenticate/oauth2.py`, the new `_extract_server_group_claims()` resolves the allowed groups during login and stores them in `session['oauth2_server_group_claims']`:

- If the claim is not configured, the session key is cleared.
- Claim values may be a single value or a list.
- Mapping values may be a string or a list.

In `web/pgadmin/utils/server_access.py`, `get_server_groups_for_user_query()` now appends an additional `OR` condition matching `ServerGroup.name` against the resolved claim-based groups — for OAuth2 users only. This applies alongside the existing ownership and shared-server (`Server.shared`) conditions.

### Tests

`web/pgadmin/browser/tests/test_oauth2_with_mocking.py`:

- Two new providers: `oidc-server-groups-direct` (no mapping) and `oidc-server-groups-mapped` (with mapping).
- Two new login-success cases asserting `session['oauth2_server_group_claims']` is populated correctly.
- The existing default assertions confirm the key is **absent** when the claim is not configured.

### Backwards compatibility

Fully opt-in. When `OAUTH2_SERVER_GROUP_CLAIM` is not set, behavior is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added OAuth2/OIDC server-group visibility controls based on identity-token or profile claims.
  * Added optional mappings from claim values to server-group names.
  * OAuth2 users now see permitted owned, shared, and claim-based server groups.
  * Outdated group visibility data is cleared when filtering is disabled.

* **Documentation**
  * Documented session storage, visibility behavior, and handling of unmapped claim values.

* **Tests**
  * Added coverage for direct and mapped server-group claims during successful login.

### DIFF
--- a/web/config.py
+++ b/web/config.py
@@ -894,6 +894,19 @@ OAUTH2_CONFIG = [
         #     'wids': ["cf1c38e5-3621-4004-a7cb-879624dced7c"],
         # }
         'OAUTH2_ADDITIONAL_CLAIMS': None,
+        # Optional server-group visibility claim controls.
+        # If OAUTH2_SERVER_GROUP_CLAIM is set, pgAdmin stores the resolved
+        # list in session['oauth2_server_group_claims'] and uses it to show
+        # additional server groups in browser tree.
+        # Without mapping, claim values are treated as server group names.
+        # Example:
+        # 'OAUTH2_SERVER_GROUP_CLAIM': 'pgadmin_server_groups',
+        # Optional mapping from claim value -> server group name(s):
+        # 'OAUTH2_SERVER_GROUP_CLAIM_MAPPING': {
+        #     'readonly': ['RO Server 1', 'RO Server 2']
+        # },
+        'OAUTH2_SERVER_GROUP_CLAIM': None,
+        'OAUTH2_SERVER_GROUP_CLAIM_MAPPING': None,
         # Set this variable to False to disable SSL certificate verification
         # for OAuth2 provider.
         # This may need to set False, in case of self-signed certificates.
--- a/web/pgadmin/authenticate/oauth2.py
+++ b/web/pgadmin/authenticate/oauth2.py
@@ -537,6 +537,48 @@ class OAuth2Authentication(BaseAuthentication):
 
         return None, None
 
+    def _extract_server_group_claims(self, profile_dict, id_token_claims):
+        """
+        Resolve allowed server groups from provider claim configuration.
+
+        Config keys (per provider):
+        - OAUTH2_SERVER_GROUP_CLAIM: claim name to read
+        - OAUTH2_SERVER_GROUP_CLAIM_MAPPING: optional dict mapping
+          claim-value -> list of server group names
+
+        Returns:
+            list[str] | None
+            - list of allowed server group names if configured
+            - None when claim-based server group filtering is not configured
+        """
+        provider = self.oauth2_config.get(self.oauth2_current_client) or {}
+        claim_name = provider.get('OAUTH2_SERVER_GROUP_CLAIM')
+        if not claim_name:
+            return None
+
+        claim_values = id_token_claims.get(claim_name)
+        if claim_values is None:
+            claim_values = profile_dict.get(claim_name)
+        if claim_values is None:
+            return []
+
+        if not isinstance(claim_values, list):
+            claim_values = [claim_values]
+
+        mapping = provider.get('OAUTH2_SERVER_GROUP_CLAIM_MAPPING') or {}
+        server_groups = []
+        for value in claim_values:
+            mapped_groups = mapping.get(value)
+            if mapped_groups is None:
+                server_groups.append(value)
+            elif isinstance(mapped_groups, str):
+                server_groups.append(mapped_groups)
+            elif isinstance(mapped_groups, list):
+                server_groups.extend(mapped_groups)
+
+        # Remove duplicates
+        return list(set(server_groups))
+
     def login(self, form):
         if not self.oauth2_current_client:
             error_msg = gettext('No OAuth2 provider available.')
@@ -599,6 +641,15 @@ class OAuth2Authentication(BaseAuthentication):
                 )
             current_app.logger.error(error_msg)
             return False, error_msg
+
+        oauth2_server_group_claims = self._extract_server_group_claims(
+            profile_dict, id_token_claims
+        )
+        if oauth2_server_group_claims is None:
+            session.pop('oauth2_server_group_claims', None)
+        else:
+            session['oauth2_server_group_claims'] = \
+                oauth2_server_group_claims
 
         additional_claims = None
         if 'OAUTH2_ADDITIONAL_CLAIMS' in self.oauth2_config[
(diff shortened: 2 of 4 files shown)


dpage (commented):
Thanks for this, and for the linked issue: claim-driven server group access is something we get asked about fairly regularly, so the direction is welcome. The claim plumbing in `oauth2.py` reads well, the `needs_userinfo` handling is right, and the session round-trip is covered by tests. My concerns are almost entirely about the single consumer in `server_access.py`, where I think there are two problems that need resolving before this can go in, plus a design question that is worth agreeing first.

### Must fix

**1. An OAuth2 user can rename another user's server group.**

In `web/pgadmin/utils/server_access.py`, the `hide_shared` branch previously applied `ServerGroup.user_id == current_user.id` as a standalone filter, whereas it is now folded into `or_(*conditions)` alongside the claim condition. `ServerGroupView.update()` in `web/pgadmin/browser/server_groups/__init__.py:226` passes `hide_shared=True` specifically to obtain an owned-only lookup, as its comment says, and it then assigns `servergroup.name` and commits without any further ownership check. So an OAuth2 user whose claim list happens to contain a name such as `Production` can `PUT` the id of somebody else's group of that name and rename it. The claim condition needs to stay out of the `hide_shared=True` branch; ideally the write paths would not share this query at all, since it is now doing double duty as both a visibility helper and an ownership check.

**2. As written, the claim grants visibility but not access, so the feature is close to a no-op.**

Server visibility is decided entirely by `get_user_server_query()` (`server_access.py`), which is still own-or-shared and is untouched here, and `ServerModule.get_nodes()` (`browser/server_groups/servers/__init__.py:292`) goes through it. That leaves three cases and none of them do what the issue asks for: a claim-granted group owned by another user shows up in the tree but is empty, because that owner's private servers are still filtered out; a group that does contain `shared=True` servers is already visible to every authenticated user via the existing `has_shared_servers` condition, so the new `OR` adds nothing; and with the `hide_shared_server` preference enabled, `get_servers()` skips non-owned servers anyway, so the group is empty again. To actually deliver #9704 the claim has to feed server access as well as group listing, not just the group query.

### Design question

The description frames this as controlling access and offering different levels of access, but the implementation can only ever widen visibility, so an administrator cannot use it to confine a user to a subset of groups. The only existing mechanism that hides shared groups is the per-user `hide_shared_server` preference, which the user can simply switch off, so that is not an access control either. I think the semantics we want are "when the claim is configured, visibility becomes owned groups plus claim-granted groups", which is restrictive rather than additive, but it would be worth agreeing that before you spend more time on the implementation.

### Should fix

**Matching `ServerGroup.name` across all owners.** `ServerGroup` rows are per-user, so `ServerGroup.name.in_(...)` matches every user's group with a given name. Two users each owning a group called "Production" will present the claim holder with two indistinguishable tree nodes, and `ServerGroupView.properties()` (`server_groups/__init__.py:280`) returns the owner's `user_id`, so it also discloses which accounts exist. The lookup should be scoped to a defined owner rather than matched globally by name.

**Ordering of the session write.** In `web/pgadmin/authenticate/oauth2.py`, `session['oauth2_server_group_claims']` is written before the `OAUTH2_ADDITIONAL_CLAIMS` validation that can still reject the login. It is not exploitable, since the consumer is gated on `current_user.auth_source == OAUTH2`, but moving it below the `valid_combined` check is cleaner and costs nothing.

**Test coverage for the behaviour that changed.** The new cases assert what lands in the session, which is useful, but nothing exercises `get_server_groups_for_user_query()` with claims present, and in particular nothing covers the `hide_shared=True` path where the rename problem above lives. CodeRabbit also asked for a regression test covering a server group claim that exists only in the userinfo response; the `needs_userinfo` fix was made but that test has not been added.

**Documentation.** `docs/en_US/oauth2.rst` documents every other `OAUTH2_*` key in its configuration table, with a worked example section for `OAUTH2_ADDITIONAL_CLAIMS`, so the two new keys need equivalent rows and a short example covering what happens to unmapped claim values.

### Minor

`list(set(server_groups))` gives non-deterministic ordering, which is why the test needed `assertCountEqual`; `list(dict.fromkeys(server_groups))` dedupes whilst preserving order. `mapping.get(value)` will raise `TypeError: unhashable type` if a claim value is a list or a dict, which some providers emit when groups are objects rather than strings, and since the call is not wrapped that would surface as a 500 during login rather than a clear message, so skipping non-scalar values would be safer. Finally, claim values are matched to group names exactly, with no trimming or case folding, which is a defensible choice but worth stating in the docs given how often provider group names arrive `/`-prefixed or differently cased.

### Notes

Claims are resolved only at login, so a role change at the identity provider takes effect on the user's next login. That is reasonable, but it deserves a line in the documentation. The module docstring at the top of `server_access.py` also still says these helpers restrict users to servers they own or that have been explicitly shared, and only the function docstring was updated.

### Verification

I ran `browser.tests.test_oauth2_with_mocking` against the branch and all 23 cases pass, including both new scenarios, and `pycodestyle` is clean on all four changed files. One unrelated failure appeared in `ReverseEngineeredSQLTestCases`, but it reproduces on an unmodified `master` in the same working tree, so it is environmental rather than anything to do with this change. Both CodeRabbit findings were valid and look correctly addressed in 19c0599.

Issue #10221, opened 2026-08-01
Lazy-load non-essential blueprints to improve desktop startup


dev-hari-prasad:
## Problem

Some blueprints are imported during `create_app()` even though they are not required for desktop startup (`SERVER_MODE=False`). If we only import essential blueprints on startup and lazy load the other this would definetly result in faster load times.

After tracing the startup flow here is a breakdown of the blureprints and how we should ideally load them:

## Modules that should remain eagerly loaded

| Module | Reason |
|--------|--------|
| `misc` | Provides the `/misc/ping` endpoint that Electron polls while waiting for the backend to start. Without it, the application never gets past the splash screen. |
| `browser` | Has`index()`, the first page Electron loads after startup. Deferring it only shifts the wait from the splash screen to a blank window. `preferences` also depends on it. |
| Core `authenticate` | Required for session and login handling, even in desktop mode. |

## Good lazy-loading candidates

| Module | Load when | Reason | Approx. size |
|--------|-----------|--------|--------------|
| `authenticate` (kerberos, ldap, mfa, oauth2, webserver) | Never in desktop mode | Gate imports in `authenticate/registry.py` behind `config.SERVER_MODE`. Desktop runs with `SERVER_MODE=False` and doesn't use these providers. | ~14 files |
| `tools` | First tool opens | Query Tool, Debugger, Backup/Restore, ERD, Import/Export, etc. aren't needed during startup. | ~155 files |
| `llm` | AI assistant opened | Depends on `pgadmin.tools.user_management`, so it would likely move together with `tools`. | ~32 files |
| `dashboard` | First server/database opened | The dashboard isn't shown until after a connection is opened. | ~7 files |

> Note: Considering how often tools are used, we can keep the tools blueprint in the early-loading category for a few select tools.

## Not worth lazy-loading

These modules are tiny (1–4 files each):

- `preferences`
- `settings`
- `about`
- `help`
- `redirects`

## Estimated impact

Potentially removes eager imports for roughly:

- `tools`
- `llm`
- External authentication backends

≈ **190 of ~915 Python files**

> File count is only a rough proxy and **not** an estimate of startup improvement.

## Validation

Profile import costs:

```bash
python -X importtime -c "from pgadmin.tools import blueprint"
python -X importtime -c "from pgadmin.llm import blueprint"
python -X importtime -c "from pgadmin.authenticate.registry import AuthSourceRegistry"
```

Then compare against overall `create_app()` startup time.

## Proposal

[...]

PR #10240, opened 2026-08-09
Gate external auth providers behind SERVER_MODE on Desktop Mode 

dev-hari-prasad:
Gate external authentication providers (`kerberos`, `ldap`, `mfa`, `oauth2`, `webserver`) behind `config.SERVER_MODE`. Desktop mode (`SERVER_MODE = False`) now only imports and registers `internal` authentication, avoiding unnecessary module overhead at startup.

## Changes
- **`web/pgadmin/authenticate/registry.py`**: Gate external auth imports & `init_app` inside `load_modules()` with `if config.SERVER_MODE:`.
- **`web/pgadmin/authenticate/tests/test_auth_gating.py`**: Added unit tests verifying external providers are not imported in desktop mode and are loaded in server mode.

## Verification & Performance
- **Tests**: `pytest` **2/2 PASSED**, `pycodestyle` **0 errors**.
- **Startup Improvement**: `load_modules()` execution dropped from **~138 ms** (Server Mode) to **< 1 ms** (Desktop Mode).

Partially addresses #10221

## Summary by CodeRabbit

* **Bug Fixes**
  * Desktop mode now loads only internal authentication, preventing unnecessary external authentication providers from being initialized.
  * Server mode continues to support configured external providers, including Kerberos, LDAP, MFA, OAuth2, and webserver authentication.

* **Tests**
  * Added coverage to verify authentication-provider loading behavior in both desktop and server modes.

### DIFF
--- a/web/pgadmin/authenticate/registry.py
+++ b/web/pgadmin/authenticate/registry.py
@@ -10,6 +10,7 @@
 """External Authentication Registry."""
 
 
+import config
 from pgadmin.utils.dynamic_registry import create_registry_metaclass
 
 
@@ -19,20 +20,21 @@ def load_modules(cls, app=None):
     from . import internal as module
     submodules.append(module)
 
-    from . import kerberos as module
-    submodules.append(module)
+    if config.SERVER_MODE:
+        from . import kerberos as module
+        submodules.append(module)
 
-    from . import ldap as module
-    submodules.append(module)
+        from . import ldap as module
+        submodules.append(module)
 
-    from . import mfa as module
-    submodules.append(module)
+        from . import mfa as module
+        submodules.append(module)
 
-    from . import oauth2 as module
-    submodules.append(module)
+        from . import oauth2 as module
+        submodules.append(module)
 
-    from . import webserver as module
-    submodules.append(module)
+        from . import webserver as module
+        submodules.append(module)
 
     for module in submodules:
         if "init_app" in module.__dict__.keys():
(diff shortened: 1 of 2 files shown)


dpage:
Merged as 2de30f25c, thank you. The gating itself was sound, and I checked each provider's desktop-mode behaviour before taking it: MFA is already inert because `mfa_enabled()` requires `SERVER_MODE`, `get_logout_url()` guards its kerberos and oauth2 branches on `SERVER_MODE`, and `AUTHENTICATION_SOURCES` defaults to `internal`.

I folded in one fix on the way through, because the new test would never have run in CI:

1. `web/pgadmin/authenticate/tests/` had no `__init__.py`, so `pkgutil.iter_modules()` did not report it and `find_modules()`, which `TestsGeneratorRegistry.load_generators()` walks, never reached the module.
2. The class derived from `unittest.TestCase` rather than `BaseTestGenerator`. Registration happens through the `TestsGeneratorRegistry` metaclass on that base class, and `get_suite()` builds the suite from that registry, so a plain `TestCase` is invisible to it.

Both are fixed, and the two scenarios now run and pass under `python regression/runtests.py --pkg authenticate`. Worth knowing for future contributions: we do not run pytest in CI, everything goes through `regression/runtests.py`, so a test needs to derive from `BaseTestGenerator` and live in a proper package to count.

I also dropped the assertion that `pgadmin.authenticate.mfa` is absent from `sys.modules` in desktop mode, because it is not: `pgadmin/browser/__init__.py` and `pgadmin/user_login_check.py` both import `pgadmin.authenticate.mfa.utils` unconditionally, the latter from `pga_login_required`, which decorates nearly every authenticated view. The original assertion passed only because the test popped the module itself beforehand. The saving is real for the other four providers; MFA was always going to be imported.

Since the commit was rebased and squashed the SHA changed, so GitHub will not close this automatically, hence the manual close.

dev-hari-prasad:
Thanks for merging❕

PR #10248, opened 2026-08-12
Schema Diff: ignore defseqrelid to avoid false SERIAL/BIGSERIAL diffs and invalid ALTER SQL (#10236)

### LINKED ISSUE #10236: Schema Diff: False positive and invalid SQL generated for BIGSERIAL / SERIAL columns
**Bug Description**
When comparing two identical tables using Schema Diff (where both tables use BIGSERIAL / SERIAL columns with identical definitions), Schema Diff incorrectly flags the table as different. Furthermore, it generates an invalid SQL script attempting to change the column type to bigserial (ALTER COLUMN ... TYPE bigserial;), which is syntactically invalid in PostgreSQL.

**To Reproduce**
_Steps to reproduce the behavior:_

- Open Schema Diff tool in pgAdmin 4.

- Select a Source and Target database that both contain a table with a BIGSERIAL column created as follows:

SQL
```
CREATE TABLE public.ges_adelantos
(
    adl_id bigserial NOT NULL,
    adl_grupo integer NOT NULL,
    adl_empresa integer NOT NULL,
    adl_persona integer NOT NULL,
    adl_ciclo integer,
    adl_moneda integer,
    adl_valor_adelanto numeric(25,6),
    adl_valor_saldo numeric(25,6),
    adl_bajado numeric(25,6),
    adl_situacion character varying(1) COLLATE pg_catalog."default",
    adl_tipo character varying(1) COLLATE pg_catalog."default",
    adl_id_factura_inclusion integer,
    CONSTRAINT ges_adelantos_pkey PRIMARY KEY (adl_id)
)
TABLESPACE pg_default;

ALTER TABLE IF EXISTS public.ges_adelantos
    OWNER to postgres;
```

- Run the Schema Diff process between Source and Target.

- Observe that the table is marked as modified.

- Inspect the generated DDL script under Differences / DDL Comparison.

**Expected Behavior**

Schema Diff should recognize that both table schemas are structurally identical and should not mark them as different (or should gracefully ignore minor underlying sequence metadata mismatches when the table definitions match).

The generated diff SQL should NOT contain invalid syntax such as ALTER COLUMN <col_name> TYPE bigserial;, since bigserial is a pseudo-type in PostgreSQL and cannot be used in ALTER TABLE ... ALTER COLUMN ... TYPE statements.

**Generated DDL (Incorrect)**

SQL
```
ALTER TABLE public.ges_adelantos
    ALTER COLUMN adl_id TYPE bigserial;

ALTER TABLE IF EXISTS public.ges_adelantos
    ALTER COLUMN adl_id DROP DEFAULT;

ALTER TABLE IF EXISTS public.ges_adelantos
    ALTER COLUMN adl_id SET INCREMENT 1 RESTART SET START 1 SET MINVALUE 1 SET MAXVALUE 9223372036854775807 SET CACHE 1;
```

**Desktop Environment**
OS: Windows 11
pgAdmin Version: 9.17
Mode: Desktop Mode
PostgreSQL Version: 15

agu2347:
## Bug

Schema Diff reports a false-positive difference (and generates invalid SQL) for structurally identical SERIAL/BIGSERIAL/SMALLSERIAL columns across two databases.

Fixes #10236

## Root cause

`get_formatted_columns()` reprojects a column as SERIAL/BIGSERIAL/SMALLSERIAL when it owns (`pg_depend`, `deptype='a'` -> `seqrelid`) the same sequence referenced by its own `nextval()` DEFAULT (`pg_attrdef` -> `defseqrelid`; see #9896/#10100/#10101). Both `seqrelid` and `defseqrelid` survive into the column dict that Schema Diff compares, but they are raw sequence OIDs -- meaningless to compare across two independently-created databases. Two databases can have byte-for-byte identical BIGSERIAL columns while their owned sequences get assigned completely different OIDs.

`SchemaDiffTableCompare.column_keys_to_ignore` already excluded `seqrelid`, but not its sibling `defseqrelid`. That gap caused two symptoms from the same root cause:

1. `compare()` / `are_dictionaries_identical()` reports a table as different purely because of the benign `defseqrelid` mismatch, even when every user-visible column property (including `cltype`) is identical.
2. `compare_target_cols()` adds the column to the `changed` diff list. Downstream SQL generation (`columns/sql/*/update.sql`) then renders `ALTER COLUMN ... TYPE {cltype}` for that column -- even though `data.cltype == o_data.cltype` (`'bigserial'` on both sides) -- producing the invalid statement reported in the issue: `ALTER COLUMN adl_id TYPE bigserial;`.

## Fix

Add `'defseqrelid'` to `SchemaDiffTableCompare.column_keys_to_ignore`, alongside its existing sibling `'seqrelid'`. This fixes both symptoms at the source: the column is no longer considered different at all, so neither the false-positive diff nor the invalid `ALTER COLUMN` statement is generated.

## Testing

Added `tests/test_schema_diff_serial_column_ignore.py`, mirroring the existing `columns/tests/test_serial_detection_unit.py` unit-test convention (`BaseTestGenerator` + mock dict fixtures, no live DB needed for the logic under test):

- Two BIGSERIAL columns differing only by `defseqrelid` are not flagged as added/changed.
- A genuine difference (`NOT NULL` toggled) on an otherwise-identical BIGSERIAL column is still detected, proving the fix doesn't mask real diffs.

Since standing up the full Postgres-backed regression harness in my sandbox was impractical for this change, I additionally verified the *actual* patched `SchemaDiffTableCompare.compare_target_cols()` logic in isolation (loading the real class source directly, bypassing only the unrelated Flask/DB-heavy base-class import) with 6 checks, including a check that reverting the fix reproduces the exact false positive described in the issue (proving this is a genuine regression guard, not a tautological test). All passed.

`pycodestyle --config=.pycodestyle` (the project's actual CI style gate, confirmed via `.github/workflows/check-python-style.yml`) is clean on both changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed schema comparisons for serial columns across independently created databases.
  * Prevented differing sequence identifiers from being reported as false differences or generating invalid alteration statements.
  * Genuine column changes, such as `NOT NULL` differences, continue to be detected.

### DIFF
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/schema_diff_table_utils.py
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/schema_diff_table_utils.py
@@ -24,8 +24,19 @@ class SchemaDiffTableCompare(SchemaDiffObjectCompare):
                             'rows_cnt', 'hastoasttable', 'relhassubclass',
                             'relacl_str', 'setting']
 
+    # 'seqrelid' (the sequence a column owns) and 'defseqrelid' (the
+    # sequence referenced by the column's nextval() DEFAULT) are compared
+    # against EACH OTHER by get_formatted_columns() to reproject a column
+    # as SERIAL/BIGSERIAL/SMALLSERIAL (see columns/utils.py, #9896/#10100/
+    # #10101) -- but the resulting raw sequence OIDs are otherwise
+    # meaningless across two independently-created databases, even when
+    # both sides have an identical SERIAL column. Without ignoring
+    # 'defseqrelid' here too, Schema Diff falsely reports such columns (and
+    # therefore their whole table) as different, and generates an invalid
+    # `ALTER COLUMN ... TYPE bigserial` statement despite both sides having
+    # the exact same reprojected cltype.
     column_keys_to_ignore = ['atttypid', 'edit_types', 'elemoid', 'seqrelid',
-                             'indkey', 'seqtypid']
+                             'indkey', 'seqtypid', 'defseqrelid']
 
     constraint_keys_to_ignore = ['relname', 'nspname', 'parent_tbl',
                                  'attrelid', 'adrelid', 'fknsp', 'confrelid',
--- a/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/tests/test_schema_diff_serial_column_ignore.py
+++ b/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/tests/test_schema_diff_serial_column_ignore.py
@@ -1,0 +1,83 @@
+##########################################################################
+#
+# pgAdmin 4 - PostgreSQL Tools
+#
+# Copyright (C) 2013 - 2026, The pgAdmin Development Team
+# This software is released under the PostgreSQL Licence
+#
+##########################################################################
+
+"""Unit tests for SchemaDiffTableCompare's column comparison, verifying
+that a SERIAL/BIGSERIAL column's raw sequence OID ('defseqrelid') does not
+cause Schema Diff to report a false-positive difference (#10236).
+"""
+
+from pgadmin.browser.server_groups.servers.databases.schemas.tables.\
+    schema_diff_table_utils import SchemaDiffTableCompare
+from pgadmin.utils.route import BaseTestGenerator
+
+
+def _make_bigserial_column(defseqrelid, **overrides):
+    """A column dict as returned by get_formatted_columns() for a
+    genuine, already-reprojected BIGSERIAL column."""
+    defaults = dict(
+        name='adl_id', cltype='bigserial', typname='bigserial',
+        atttypid=20, attlen=8, attnum=1, elemoid=20, seqtypid=20,
+        indkey=None, seqrelid=defseqrelid, defseqrelid=defseqrelid,
+        defval='', attnotnull=True, attacl=[],
+    )
+    defaults.update(overrides)
+    return defaults
+
+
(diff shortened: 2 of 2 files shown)


dpage:
Thanks for this, and for the unusually thorough write-up: the diagnosis was spot on, and I've merged it as 01e5bf099 (rebased and squashed, so GitHub won't close this automatically).

I did fold two things in whilst merging. Your change stops an unchanged serial column being reported as different, but the invalid SQL had a second source that survived it: the diff data carries the reprojected pseudo-type whilst `_check_for_column_update()` reads the old column straight from the catalogue, so as soon as a serial column has *any* genuine difference (a comment, a NOT NULL, a privilege) it still rendered `ALTER COLUMN ... TYPE bigserial`, a spurious `DROP DEFAULT`, and sequence options that only an identity column accepts. Reproducing that is easy: take your test case and add a differing column comment on the `bigserial` column. The changed column is now normalised back onto the underlying integer type before `update.sql` sees it, and the owned sequence is left to the sequence node's own comparison.

I also added an end-to-end Schema Diff test (`web/pgadmin/tools/schema_diff/tests/test_schema_diff_serial_column.py`) which compares two independently created databases, asserts the identical serial table comes out identical, and applies the generated SQL to the target before re-comparing, so an invalid statement now fails the suite rather than passing quietly. The existing fixtures had no serial column at all, which is why this got through in the first place.

One note for future contributions: the branch was based far enough back that the test suite couldn't authenticate at that commit, which is unrelated to your change but does mean the tests can't run as submitted; rebasing on current `master` before submitting avoids that. Thanks again.

PR #10254, opened 2026-08-13
feat: toggle Object Explorer by re-clicking Default workspace icon

### LINKED ISSUE #9631: Toggle visibility of left panel by clicking icon when an area is already seleted.
The object panel on the left that shows object-explorer and other functionality should be able to be toggled.

If you press one of the icons (e.g. the one to toggle to object panel) while that panel is open it should close that area (and open it again if you press it again).

This is similar to the same area in vscode.

byshy:
## Summary
- Fixes #9631 by allowing the Default workspace icon to toggle Object Explorer visibility when that workspace is already selected (VS Code-like behavior).
- Collapses Object Explorer with CSS so the tree stays mounted and keeps its state.
- Persists visibility across refresh via `Browser/ObjectExplorerVisible`, and restores it on Reset Layout.

## Test plan
- [x] Switch Preferences → Layout to Workspace layout
- [x] Click the Default (tree) workspace icon while Object Explorer is open → panel collapses and main area expands
- [x] Click the same icon again → Object Explorer reappears with prior tree expansion/selection intact
- [x] Collapse Object Explorer, refresh the browser → remains collapsed
- [x] Collapse Object Explorer, use File → Reset Layout → Object Explorer is shown again
- [x] Switch to Query Tool / PSQL / Schema Diff and back → Object Explorer visibility matches the saved preference
- [x] Classic layout still shows Object Explorer as before (no workspace toolbar toggle)

## Summary by CodeRabbit

* **New Features**
  * Added the ability to show or hide the Object Explorer from the workspace toolbar.
  * Object Explorer visibility is preserved across sessions and restored after layout resets.
  * Selecting the active Default workspace now toggles Object Explorer visibility.
  * The Object Explorer can be collapsed without losing its current state.

* **Bug Fixes**
  * Resetting the layout now reliably restores the Object Explorer to its visible state.


dpage:
Thanks @byshy, this is merged: it landed on `master` as 549d71a20, squashed into a single commit with you as the author. Because the squash changes the SHA, GitHub will not close this automatically, so I am closing it by hand.

The feature works nicely, and hiding the panel with CSS rather than unmounting it was the right call: the tree keeps its expansion and selection state, and because rc-dock sizes its box children with flex, the main panel takes the freed width with nothing left behind. I checked that in a browser rather than assuming it.

I folded four things in whilst reviewing, all of them small:

The Object Explorer keyboard shortcut did nothing useful whilst the sidebar was collapsed, so it now asks for the panel to be shown first through a new `pgadmin:object-explorer:show` event on `pgAdmin.Browser.Events`. Worth knowing for future reference that the `focus()` call in `bindLeftTree()` was already inert, since `id-object-explorer` is a plain div with no `tabindex`; the useful part of that shortcut is the tree selection that follows it.

`resetLayout()` is no longer monkey-patched onto the docker instance. `LayoutDocker` fires a `RESET` event and `BrowserComponent` listens for it, which is how the rest of that area is wired and avoids mutating an object other code holds a reference to.

A failed attempt to persist the setting is now logged rather than discarded, because otherwise the sidebar could quietly disagree with the server after the next refresh with nothing to explain why.

The icon's tooltip says Show or Hide Object Explorer whilst it is acting as a toggle, and `aria-pressed` exposes the state. That one needed care: `{...props}` is spread after `title` in `WorkspaceButton`, and the caller passes its own `title`, so the existing `title={menuItem?.label}` was already dead for these buttons.

I also added `web/regression/javascript/workspaces/WorkspaceProvider.spec.jsx`, which is the first coverage this area has had, covering the default, the saved setting being honoured, each toggle persisting its value, two toggles in one tick resolving in order, and the reveal event.

Two things I deliberately left alone, in case you or anyone else fancies a follow-up: `id-object-explorer` has no `tabindex`, so nothing can actually focus the panel; and there is no menu item or keyboard shortcut for the toggle itself, which VS Code does offer alongside the icon.

PR #10281, opened 2026-08-17
Don't follow HTTP redirects on LLM API requests

dpage:
`ALLOWED_LLM_API_URLS` is checked against the URL pgAdmin has been configured with, but the providers then issued the request through `urllib.request.urlopen()`, whose default handler will happily follow a `Location` header on to a destination that check was never applied to. Validating the first URL does not imply validating every subsequent hop.

No legitimate LLM API replies with a redirect, so this refuses them outright rather than re-validating each hop. `urlopen_no_redirect()` in `web/pgadmin/llm/utils.py` mirrors `urlopen()`, including its handling of an explicit SSL context, but installs an `HTTPRedirectHandler` subclass that raises `HTTPError` instead of following. All eleven outbound call sites now go through it: two each in the Anthropic, OpenAI and Docker providers, three in Ollama (including the `is_available()` probe), and the four model-refresh helpers in `web/pgadmin/llm/__init__.py`. Every one of those sites already caught `HTTPError` or `URLError`, so a redirect surfaces as an ordinary provider error and no call site needed restructuring.

I should be clear that this is hardening rather than a fix for an exploitable flaw, because returning the redirect in the first place requires control of a host that is already on the allowlist: either one of the vendor entries, which would mean the vendor itself had been compromised, or a loopback entry, which needs code execution on the pgAdmin host and therefore grants direct access to the same services anyway. The shipped configuration already documents that the loopback entries let any logged-in user reach any port on the server's own loopback interface. The residual case worth closing is an administrator who allowlists an internal self-hosted endpoint that happens to carry an open redirect, and in any event there is no reason for us to be following redirects here at all.

Tests are in `web/pgadmin/llm/tests/test_no_redirect.py`. They stand up a real loopback HTTP server rather than mocking, and cover 301, 302, 303, 307 and 308 across both GET and POST, asserting each time that the redirect target recorded no hit. There is also a control case asserting that a plain `urllib.request.urlopen` *does* follow the same fixture, so the refusal tests cannot pass vacuously against a fixture that never redirected.

Reported by Ziya Abdullayev, whose report prompted the change.

## Summary by CodeRabbit

* **Security**
  * LLM provider requests now block HTTP redirects, helping prevent requests from being sent to unintended destinations.
  * Redirects are rejected consistently for standard and streaming requests.

* **Bug Fixes**
  * Preserved existing request timeouts, SSL settings, response parsing, and error handling.
  * Added coverage for redirect responses and normal successful requests.

### DIFF
--- a/web/pgadmin/llm/__init__.py
+++ b/web/pgadmin/llm/__init__.py
@@ -22,7 +22,7 @@ from pgadmin.user_login_check import pga_login_required
 from pgadmin.utils.constants import MIMETYPE_APP_JS
 from pgadmin.utils.csrf import pgCSRFProtect
 import config
-from pgadmin.llm.utils import LLMApiError
+from pgadmin.llm.utils import LLMApiError, urlopen_no_redirect
 from pgadmin.tools.user_management.PgAdminPermissions import \
     AllPermissionTypes
 
@@ -807,7 +807,7 @@ def _fetch_anthropic_models(api_key, api_url=''):
     req = urllib.request.Request(url, headers=headers)
 
     try:
-        with urllib.request.urlopen(
+        with urlopen_no_redirect(
             req, timeout=30, context=SSL_CONTEXT
         ) as response:
             data = json.loads(response.read().decode('utf-8'))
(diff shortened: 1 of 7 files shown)


hiteshjambhale (approved):
Tested locally: all 12 redirect cases pass, no regressions vs master. build_opener still keeps the default handlers, so nothing's lost. LGTM.
Kindly handle failing checks.

hiteshjambhale (approved):

PR #10296, opened 2026-08-17
feat: add a keyboard shortcut for collapsing the Object Explorer

dpage:
#10254 made the Object Explorer collapsible by clicking the current workspace icon, but that click is currently the only way to do it. This adds a keyboard shortcut, `Ctrl+Alt+B` by default, registered as its own `toggle_object_explorer` preference so it can be changed like any other.

### Why Ctrl+Alt+B

The `Shift+Alt+<letter>` family is full: b, f, o, l, h, q, v, s, n, e, d and g are all taken, and b belongs to the existing Object Explorer shortcut, which keeps its present meaning of focusing the tree. `Ctrl+Alt+B` is free, sits close enough to VS Code's `Ctrl+B` for the side bar to be guessable, and keeps a family resemblance to the bracket keys used for tab navigation. `Ctrl+Shift+B` was the other obvious candidate and I avoided it, since Chrome and Firefox both use it for the bookmarks bar.

One caveat worth raising rather than burying: on keyboard layouts where AltGr is Ctrl+Alt, this combination can be swallowed by the layout and produce a character instead. That applies to any `Ctrl+Alt` binding, the preference is user-configurable, and `Shift+Ctrl+Alt+B` is available if the default turns out to irritate people. Happy to change the default if reviewers prefer.

### Shape of the change

The shortcut only fires a `pgadmin:object-explorer:toggle` event, because the workspace provider owns the visibility state, and the provider ignores it in classic layout, where there is no workspace toolbar and the Object Explorer is always shown.

### Testing

`web/regression/javascript/browser/keyboard_toggle_object_explorer_spec.js` covers the shortcut asking for the toggle, and `WorkspaceProvider.spec.jsx` gains a case for the provider acting on it in both directions.

Also driven in a browser: in the workspace layout, `Ctrl+Alt+B` collapses the panel and a second press restores it, with `aria-pressed` on the workspace icon following along, and the new preference appears alongside the existing shortcuts.

Documented in `docs/en_US/keyboard_shortcuts.rst`. No release note entry, per the usual practice of batching those shortly before release.

## Summary by CodeRabbit

* **New Features**
  * Added a keyboard shortcut to toggle the Object Explorer:
    * Windows/Linux: `Ctrl + Alt + B`
    * macOS: `Ctrl + Option + B`
  * The shortcut collapses or restores the Object Explorer in workspace layouts.

* **Documentation**
  * Updated keyboard shortcut documentation with the new toggle command.

* **Tests**
  * Added coverage confirming the shortcut and visibility toggle work correctly.


hiteshjambhale (commented):
Jest specs pass and verified the ⌃⌥B toggle + persistence in-browser on macOS.
LGTM.
Kindly handle failing checks.

hiteshjambhale (approved):

dpage:
@hiteshjambhale I've rebased this branch onto current `upstream/master`, which picks up #10306 (a CI infrastructure fix for the Yarn/Corepack setup that was causing the broad check failures here). Checks should now pass on the rerun.

PR #10299, opened 2026-08-17
Javascript dependency: Bump use-resize-observer from 9.1.0 to 10.0.0 (#10265)

dpage:
Supersedes #10265, which could not be merged as it stood.

use-resize-observer 10.0.0 removes the default export in favour of a named one, whilst `web/pgadmin/static/js/PgTreeView/index.jsx:19` imported it as a default. Left alone, `useResizeObserver` resolves to `undefined` and the Object Explorer dies the moment the hook is called, so the bump needs the import change to go with it:

```js
-import useResizeObserver from 'use-resize-observer';
+import { useResizeObserver } from 'use-resize-observer';
```

I checked the published package rather than relying on the release notes: v10 exports `useResizeObserver` as a named export only, in both `dist/index.mjs` and `dist/index.cjs`, with no default in either. The hook signature still returns the `{ ref, width, height }` shape that line 81 destructures, this is the only use of the package in the tree, and there is no webpack or jest alias for it.

The rest of v10 costs us nothing: it drops the `/polyfilled` entrypoint and the `@juggle/resize-observer` dependency, no longer takes `react-dom` as a peer, and requires React >= 18.2, comfortably met on React 19.

### Testing

Verified by hand in both Chrome and Safari against a local PostgreSQL 18, with the tree expanded through databases, schemas, casts, tables and tablespaces, resizing both the Object Explorer panel and the window. The tree renders and redraws correctly in both.

Worth stating why this one was hand-tested rather than merged on a green board: `run-javascript-tests` cannot see this class of breakage, since the component only fails when it mounts. That is exactly how the pickr bump in #10264 reached master and took the feature tests down for three days (reverted in #10278), and why #10287 was closed rather than merged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tree view resizing compatibility following an update to the resize observer dependency.
  * Preserved existing tree view behavior while ensuring resize detection continues to work correctly.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -152,7 +152,7 @@
     "sql-formatter": "^15.8.2",
     "uplot": "^1.6.32",
     "uplot-react": "^1.2.4",
-    "use-resize-observer": "^9.1.0",
+    "use-resize-observer": "^10.0.0",
     "valid-filename": "^4.0.0",
     "vanilla-jsoneditor": "^3.12.0",
     "wkx": "^0.5.0",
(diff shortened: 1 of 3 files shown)


hiteshjambhale (approved):
Confirmed v10 exports useResizeObserver as named-only (no default), so the import fix is required. Verified in-browser: tree renders and resizes cleanly. LGTM.

hiteshjambhale (approved):

PR #10313, opened 2026-08-19
Fix existingSecret path in Helm deployment template

### LINKED ISSUE #10214: Fix existingSecret in helm Deployment
Hi everyone,

https://github.com/pgadmin-org/pgadmin4/blob/5fc319551d00d95cd3331ed5075416e88dc3ba5d/pkg/helm/templates/deployment.yaml#L40-L42

i think, here have to be ` .Values.**auth**.existingSecret`, 
because .Values.existingSecret is not exists and secret.yaml has a condition `{{- if empty .Values.auth.existingSecret }}`

dpage:
## Summary

`deployment.yaml`'s annotation conditions referenced the non-existent top-level `.Values.existingSecret` instead of `.Values.auth.existingSecret`, so the secret checksum annotation and the empty-secret gating never worked as intended when an existing secret wasn't supplied. `secret.yaml` already uses the correct `.Values.auth.existingSecret` path.

Fixes #10214.

## Test plan

- [x] Manually traced every `existingSecret` reference across the chart's templates and values.yaml to confirm `.Values.auth.existingSecret` is the only path that actually exists
- [ ] `helm lint`/`helm template` (not available in this environment; the change is a straightforward path correction matching the existing pattern used elsewhere in the same file)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Helm deployment configuration so secret checksum annotations correctly use the configured authentication secret.
  * Prevented deployment annotations from referencing an outdated secret configuration value.


hiteshjambhale (approved):
Tested this hands-on with helm template.
Looks good.

PR #10317, opened 2026-08-19
Schema Diff: stop injecting whitespace into applied function/procedure bodies

### LINKED ISSUE #10302: Schema Diff: applying a function or procedure change leaves a whitespace-only difference behind
**Bug Description**

Applying a Schema Diff script that updates a function or procedure leaves a whitespace-only difference behind, so comparing the two databases again still reports the object as different unless 'Ignore whitespace' is ticked.

The generated `CREATE OR REPLACE` wraps the body in newlines:

```sql
CREATE OR REPLACE PROCEDURE test_schema_diff.proc1(
	IN arg1 bigint)
LANGUAGE 'sql'
AS $BODY$
select 1;
$BODY$;
```

so the target's `prosrc` becomes `"\nselect 1;\n"` whilst the source's remains `"select 1;"`. The bodies now differ by exactly the newlines pgAdmin itself introduced, and the comparison duly reports it.

It is cosmetic in that the two functions behave identically, and ticking 'Ignore whitespace' hides it, but a script that cannot settle the difference it was generated for is confusing: re-running the comparison suggests the script did not work.

**Expected Behaviour**

Applying the generated script should leave the object comparing as identical without needing 'Ignore whitespace', which means either rendering the body exactly as it is stored, or normalising the stored body on both sides before comparing.

**Context**

Found by making the Schema Diff regression test assert that applying the generated script leaves the two databases identical (#10293). The test lists this as a known outstanding difference until it is fixed.

dpage:
## What this is

Applying a Schema Diff script that creates a function or procedure that only exists on the source ("Source Only") left a whitespace-only difference behind, so re-comparing the two databases kept reporting the object as different unless "Ignore whitespace" was ticked.

The "Source Only" path routed through `FunctionView.sql()`, which always calls `reformat_prosrc_code()` to wrap the body with a leading and trailing newline around `$BODY$...$BODY$` purely for readability in the SQL panel. That reformatted body was then executed verbatim against the target, so PostgreSQL stored the extra newlines in `prosrc`: the target's body no longer matched the source's byte for byte.

Schema Diff's update path (`CREATE OR REPLACE` against an existing target object) already suppressed this formatting via an `allow_code_formatting` flag, added for #7102, but that flag was never threaded through the create/"Source Only" path.

## The fix

Extends the existing `allow_code_formatting` guard to `sql()` itself, and passes `allow_code_formatting=False` from the `get_sql_from_diff()` branch that builds the DDL actually applied to the target.

## Testing

Added a dedicated regression test covering both the "Source Only" and "Different" cases for a function and a procedure: applying the generated script now leaves every object comparing as Identical without needing "Ignore whitespace".

`tools.schema_diff` passes against PostgreSQL 18; `pycodestyle` is clean.

Fixes #10302.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved function and procedure body whitespace in schema-diff-generated SQL.
  * Prevented unwanted leading or trailing newlines around stored code bodies.
  * Ensured source-only schema differences can be applied and subsequently compare as identical.
  * Added clear errors when unsupported new arguments are added to existing functions or procedures.

* **Tests**
  * Added regression coverage for whitespace preservation in function and procedure definitions.


hiteshjambhale (approved):
Tested this hands-on. Created a source-only function with a spaced body, ran Schema Diff, and checked the generated script — the body now stays exactly as written ($BODY$ SELECT a + b; $BODY$), no extra newlines.

also applied it to the target and the stored body matched the source exactly, so a re-compare shows Identical without ticking "Ignore whitespace"

Looks good to me.

PR #10319, opened 2026-08-19
Schema Diff: reinstate dependency ordering of the generated script

### LINKED ISSUE #10295: Schema Diff: generated script is no longer ordered by dependency (dependLevel is never set since the React port)
**Bug Description**

Schema Diff no longer orders the script it generates by object dependency, so a script that contains an object and something that object depends on can easily be written the wrong way round and fails when it is run.

`getGenerateScriptData()` in `web/pgadmin/tools/schema_diff/static/js/components/SchemaDiffCompare.jsx` buckets each selected row by `data.dependLevel`:

```js
        if (!(data.dependLevel in script_array)) script_array[data.dependLevel] = [];
        checkAndGetSchemaQuery(data, script_array);
        script_array[data.dependLevel].push(data.diff_ddl);
```

and `generateFinalScript()` then walks those buckets in reverse so that the deepest dependencies are written first. The catch is that **nothing assigns `dependLevel` any more**: it is read in exactly one file and written in none, so it is always `undefined`, every object lands in a single `undefined` bucket, and the script comes out in whatever order the rows happen to be in.

`dependLevel` used to be computed in `web/pgadmin/tools/schema_diff/static/js/schema_diff_dependency.js`, added for #5730, which walked each checked row's dependencies and assigned levels. That file was deleted wholesale by e1942d8c9 ("Port schema diff to React"), and the calculation was never reinstated, so the ordering half of #5730 has been silently absent since. The data it needs is still there: each row carries `dependencies` and `dependenciesOid`.

**To Reproduce**

Comparing two databases where a range type differs in kind (say a range in the source against an enum of the same name in the target) produces a script whose very first statement is the cast that the range's multirange type needs:

```sql
CREATE CAST (test_schema_diff.typ_range_enum_diff AS test_schema_diff.typ_multirange_enum_diff)
	WITH FUNCTION typ_multirange_enum_diff(test_schema_diff.typ_range_enum_diff);
```

whilst the `DROP TYPE`/`CREATE TYPE` that creates those very types appears several hundred lines further down, so running the script fails with:

```
ERROR:  type "test_schema_diff.typ_multirange_enum_diff" does not exist
```

**Expected Behaviour**

The generated script should be ordered so that dependencies are created before the objects that need them, as it was before the React port. The header already warns that circular dependencies may make the order less than optimal, which rather understates the current position: there is no dependency ordering at all.

**Context**

Found whilst making the Schema Diff regression test assert that its generated script actually applies (#10293). The test cannot be made strict about the script as a whole until this is fixed, so it will be made strict about each object's SQL being valid in the meantime.

dpage:
## What this is

`dependLevel` drives the order in which `generateFinalScript()` writes objects into the generated script, with higher levels emitted first so dependencies land before whatever needs them. Nothing has assigned it since the React port deleted `schema_diff_dependency.js` wholesale: every row got `Object.keys`' default `undefined` bucket, so the script came out in whatever order the rows happened to be in. In practice this means a generated script can fail outright when an object appears before something it depends on.

## The fix

`computeDependLevels()` rebuilds the levels for the full result set from each row's `dependencies`, walking the reverse graph so that a row which other rows depend on always ends up with the highest of their levels plus one, with a cycle guard for the circular dependencies the script header already warns about. It's called once, over the full row set, before the script is assembled.

## Testing

Added Jest unit tests for `computeDependLevels()` covering a dependency chain, unrelated rows, a circular dependency, a dependency pointing at an object outside the row set, and converging dependency chains of different depths. Existing `schema_diff` Jest specs and eslint both pass clean.

Fixes #10295.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema diff script generation by ordering statements according to object dependencies.
  * Added handling for circular dependencies, missing dependency targets, unrelated objects, and converging dependency chains.
* **Tests**
  * Added coverage for dependency ordering and edge cases to help ensure reliable script generation.


hiteshjambhale (approved):
Created a source DB with a real dependency chain (enum type → table using it → view + FK table), ran Schema Diff against an empty target, and the generated script came out in correct dependency order (type → table → view/FK) and ran clean with no "does not exist" errors. Confirms #10295 is fixed. LGTM.

PR #10325, opened 2026-08-19
Remove redundant passlib pin that conflicts with libpass

### LINKED ISSUE #10239: pgadmin4 is depending on both passlib and libpass (indirectly) which causes conflicts

**Describe the bug**

I got issue when trying to install pgadmin4 with uv and it eventually boils down to conflicting dependencies on both passlib and libpass. libpass is a fork of passlib and they have the same top level directory.

The two dependency trees are as follows:

- Direct dependency: `passlib==1.*`
- Transit dependency: `Flask-Security-Too==5.7.1` -> `libpass>=1.9.3`. See: https://github.com/pallets-eco/flask-security/blob/main/pyproject.toml

For me, this only happens in `uv` because `uv` seems to be downloading and installing packages in parallel but I think installing packages with the same top level directory may trigger undefined behavior. For my `uv` run, there are 3 possible cases:

1. The installed `passlib` directory contains files from both packages. I then get weird errors saying for example some packages are missing in `passlib`
2. If the installed `passlib` directory is indeed from `passlib`, I got `pkg_resources` not found
3. If the installed `passlib` directory is from `libpass`, then there is no error

**To Reproduce**

1. Install pgadmin4 with uv as a tool (uv tool install pgadmin4)
2. The above scenario 1 does not always happen but to trigger scenario 2, one can run `uv pip install --python /path/to/tools/pgadmin4 --reinstall passlib` which will produce the `pkg_resources` not found error
3. To trigger scenario 3, one can run `uv pip install --python /path/to/tools/pgadmin4 --reinstall libpass`

**Expected behavior**

Given that when `libpass` overrides `passlib`, there is no startup error, I think perhaps `passlib` is not needed any more. Anyway, `passlib` is 6 years old I think pgadmin4 should consider moving away from this

**Error message**

When files from both libraries got mixed, for me, it says: `cannot import name 'numeric_types' from 'passlib.utils.compat'

If `passlib` wins, it says: `pkg_resources` not found

**Screenshots**

 - OS: Linux RHEL9
 - pgAdmin version: 9.12
 - Mode: Server
 - Browser (if running in server mode): chrome
 - Package type: pip

**Additional context**

dpage:
## Summary
- `requirements.txt` pinned `passlib==1.*` directly, but `Flask-Security-Too` (>= 5.7) depends on `libpass`, a fork of passlib that installs its files under the same `passlib/` import path.
- Installing both packages in one environment causes a file-level collision under `site-packages/passlib/`; with parallel installers such as `uv` the result depends on install order and can leave a mixed or overwritten set of files (see issue for the exact failure modes).
- pgAdmin's own code never imports `passlib` directly. The older `Flask-Security-Too` 5.6.x branch (used on Python 3.9) still depends on real `passlib` itself, so removing our direct pin doesn't change what gets installed there; on newer Python versions only `libpass` is installed, which resolves the conflict.

## Test plan
- [x] Confirmed no direct `passlib`/`libpass` imports anywhere under `web/pgadmin` (excluding tests).
- [x] Confirmed via `pip show` that the installed `Flask-Security-Too==5.8.2` requires `libpass`, not `passlib`, and nothing in the current dependency graph requires `passlib`.
- [x] Confirmed via the `Flask-Security-Too` 5.6.1 wheel metadata that the Python 3.9 branch still requires `passlib>=1.7.4` transitively, so dropping the top-level pin doesn't remove it from that branch.
- [ ] No functional test applicable — this is a dependency pin removal with no code changes.

Closes #10239

## Summary by CodeRabbit

* **Chores**
  * Removed an unused runtime dependency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -48,7 +48,6 @@ keyring==25.*
 ldap3==2.*
 libgravatar==1.0.*
 paramiko==3.5.1
-passlib==1.*
 psutil==7.2.*
 psycopg[c]==3.2.13; python_version < '3.10'
 psycopg[c]==3.3.4; python_version >= '3.10'


hiteshjambhale (approved):
Verified: nothing in pgAdmin imports passlib directly, and Flask-Security-Too 5.8.x pulls in libpass. Sanity-tested a desktop build with the pin removed: app boots clean and password hashing/login work fine. LGTM.

PR #10326, opened 2026-08-19
fix: guard PasswordExec against None host/port/username

### LINKED ISSUE #10237: replace() argument 2 must be str, not None when using per-server Password Exec Command with Service field (pg_service.conf)
When a server is configured in pgAdmin using the Service field (pointing to a pg_service.conf entry) with host, port, and username left blank, and a per-server Password Exec Command is set, connecting to the server raises: replace() argument 2 must be str, not None

Steps to Reproduce

Create a pg_service.conf entry with host, port, dbname, and user defined.
In pgAdmin, register a new server. In the Connection tab, set only the Service field to the service name. Leave host, port, username, and password blank.
In the same Connection tab, set a Password Exec Command (e.g., a PowerShell script that retrieves an Azure AD access token via az account get-access-token).
Attempt to connect to the server.
Expected Result

The password exec command runs and its output is used as the password, establishing the connection successfully.

Actual Result

The connection fails immediately with: replace() argument 2 must be str, not None

Root Cause

The bug is in web/pgadmin/utils/passexec.py in the PasswordExec.get() method:

self.cmd = self.cmd.replace('%HOSTNAME%', self.host)
self.cmd = self.cmd.replace('%PORT%', str(self.port))
self.cmd = self.cmd.replace('%USERNAME%', driver.qtIdent(None, self.username))

When the server is configured with only the Service field set, self.host, self.port, and self.username are all None. Passing None as the second argument to str.replace() raises the error.

Suggested Fix

self.cmd = self.cmd.replace('%HOSTNAME%', self.host or '')
self.cmd = self.cmd.replace('%PORT%', str(self.port) if self.port is not None else '')
self.cmd = self.cmd.replace('%USERNAME%', driver.qtIdent(None, self.username) if self.username else '')

Environment

OS: Windows 11
pgAdmin version: 9.17
PostgreSQL: Azure Database for PostgreSQL (Flexible Server)
Authentication: Azure AD access token via az account get-access-token

dpage:
## Summary
- Service-only connections (`pg_service.conf` with host/port/username left blank) crashed a per-server Password Exec Command with `TypeError: replace() argument 2 must be str, not None`, because `PasswordExec.get()` passed `None` straight into `str.replace()`.
- `%HOSTNAME%`, `%PORT%` and `%USERNAME%` placeholders now fall back to an empty string when the corresponding attribute is unset.

## Test plan
- [x] Added `web/pgadmin/utils/tests/test_passexec.py`, verified it fails against the pre-fix code and passes with the fix.
- [x] `pycodestyle` clean on both changed files.

Closes #10237

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved password retrieval for Service-only connections when host, port, or username details are unavailable.
  * Ensured missing connection values are handled safely while preserving username formatting where applicable.

* **Tests**
  * Added regression coverage for password retrieval with incomplete connection details.

### DIFF
--- a/web/pgadmin/utils/passexec.py
+++ b/web/pgadmin/utils/passexec.py
@@ -39,10 +39,12 @@ class PasswordExec:
             raise NotImplementedError('Passexec not available in server mode')
         driver = get_driver(config.PG_DEFAULT_DRIVER)
         self.cmd = str(self.cmd)
-        self.cmd = self.cmd.replace('%HOSTNAME%', self.host)
-        self.cmd = self.cmd.replace('%PORT%', str(self.port))
-        self.cmd = self.cmd.replace('%USERNAME%',
-                                    driver.qtIdent(None,self.username))
+        self.cmd = self.cmd.replace('%HOSTNAME%', self.host or '')
+        self.cmd = self.cmd.replace(
+            '%PORT%', str(self.port) if self.port is not None else '')
+        self.cmd = self.cmd.replace(
+            '%USERNAME%',
+            driver.qtIdent(None, self.username) if self.username else '')
         with self.lock:
             if not self.password or self.is_expired():
                 if not self.cmd:
--- a/web/pgadmin/utils/tests/test_passexec.py
+++ b/web/pgadmin/utils/tests/test_passexec.py
@@ -1,0 +1,49 @@
+##########################################################################
+#
+# pgAdmin 4 - PostgreSQL Tools
+#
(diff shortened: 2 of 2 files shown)


hiteshjambhale (approved):
Verified manually in desktop mode: registered a service-only server (blank host/port/username, Service= set) with a Password exec command. Reproduced the TypeError: replace() argument 2 must be str, not None on master, and confirmed this branch connects cleanly. LGTM.

Issue #10251, opened 2026-08-12
Maintenance dialog REINDEX generates invalid SQL when "Concurrently" is enabled ("syntax error at or near CONCURRENTLY")


renato-machado-atman:
### Bug report

**pgAdmin version**: 9.17
**PostgreSQL version**: 13.2
**OS**: Ubuntu 24.04.4 LTS

### Description

Running REINDEX TABLE with the "Concurrently" option enabled from the
Maintenance dialog generates invalid SQL. `CONCURRENTLY` is placed inside
the parenthesized option list alongside `VERBOSE`, but PostgreSQL only
accepts `CONCURRENTLY` as a standalone keyword placed between the target
type and the target name — not as a parenthesized option.

### Steps to reproduce

1. Right-click a table (e.g. `public."Command"`) → Maintenance...
2. Select "REINDEX", enable "Verbose" and "Concurrently"
3. Run

### Generated SQL (invalid)

​```sql
REINDEX (VERBOSE, CONCURRENTLY) TABLE public."Command";
​```

### Error

​```
ERROR: syntax error at or near "CONCURRENTLY"
LINE 1: REINDEX (VERBOSE, CONCURRENTLY) TABLE public."Command";
                          ^
​```

### Expected SQL

​```sql
REINDEX (VERBOSE) TABLE CONCURRENTLY public."Command";
​```

### Root cause (from source)

In `web/pgadmin/tools/maintenance/templates/maintenance/sql/command.sql`,
`CONCURRENTLY` is appended to the same `maintenance_options` list as
`VERBOSE`/`TABLESPACE`/etc., and the whole list is rendered inside a single
`(...)` clause before the target type/name. Per the PostgreSQL grammar,
`CONCURRENTLY` is not a parenthesizable REINDEX option — it must appear
after the object type keyword (`TABLE`/`INDEX`/`SCHEMA`/...) and before the
object name. The template needs to emit `CONCURRENTLY` there instead of
inside the options parens.

### Impact

Any REINDEX (TABLE/INDEX/SCHEMA/DATABASE/SYSTEM) run via the Maintenance
dialog with "Concurrently" checked fails outright with a syntax error,
making the feature unusable from the UI. Workaround: run the corrected SQL
manually in the Query Tool.

### Screenshot

<img width="716" height="415" alt="Image" src="https://github.com/user-attachments/assets/8807355e-ee79-43b8-ba8a-ca11cf7a32b2" />

PR #10329, opened 2026-08-19
Exclude non-editable alias/expression columns from Query Tool UPDATE saves

### LINKED ISSUE #10103: Editing a row with expression/alias columns in Query Tool fails to save with "column does not exist" error

dpage:
## What this is

If a Query Tool result set includes a calculated or aliased column (e.g. `first_name || ' ' || last_name AS the_name`), editing any cell on an existing row and saving throws `column "the_name" does not exist`. The frontend already marks such columns non-editable (shown with a lock icon), but `save_changed_data()`'s update path still passed them through when rendering the `UPDATE` statement.

The insert path already had this guard, added for #9939; the update path just never got the equivalent filter.

## Fix

`save_changed_data()` now drops any key not present in `columns_info`, or explicitly marked `is_editable: False`, before rendering the `UPDATE`, in both the insert and update code paths.

Fixes #10103.

## Test plan

- [x] Added `TestSaveUpdatedRowSkipsNonEditableColumn`, mirroring the existing insert-path test for #9939.
- [x] `regression/runtests.py --pkg tools.sqleditor.utils.tests.test_save_changed_data` — 14/14 passed.
- [x] `regression/runtests.py --pkg tools.sqleditor` — 158 passed, 3 pre-existing/unrelated skips.
- [x] `pycodestyle` clean.

## Summary by CodeRabbit

- **New Features**
  - Users with PostgreSQL `ADMIN OPTION` can manage role memberships without broader role-editing privileges.
  - Shared server imports now support shared usernames while enforcing usernames for non-shared servers.

- **Bug Fixes**
  - Fixed concurrent `REINDEX` command generation.
  - Prevented non-editable query-result columns from being included in updates.
  - Improved transaction handling when server-side cursors are active.
  - Corrected authentication-secret handling in deployments.

- **Tests**
  - Added coverage for role permissions, username validation, concurrent maintenance commands, safe updates, and transaction handling.

### DIFF
--- a/web/pgadmin/tools/sqleditor/utils/save_changed_data.py
+++ b/web/pgadmin/tools/sqleditor/utils/save_changed_data.py
@@ -177,6 +177,21 @@ def save_changed_data(changed_data, columns_info, conn, command_obj,
             list_of_sql[of_type] = []
             for each_row in changed_data[of_type]:
                 data = changed_data[of_type][each_row]['data']
+
+                # Drop any column the client included that isn't a real
+                # editable column of the underlying table (e.g.
+                # `first_name || ' ' || last_name as the_name`). The
+                # frontend already marks such columns as non-editable
+                # (shown with a lock icon), but still includes them in
+                # the changed data. Without this guard the rendered
+                # UPDATE references a non-existent column and Postgres
+                # rejects it. Issue #10103.
+                data = {
+                    k: v for k, v in data.items()
+                    if k in columns_info and
+                    columns_info[k].get('is_editable', True)
+                }
+
                 pk_escaped = {
                     pk: pk_val.replace('%', '%%') if hasattr(
                         pk_val, 'replace') else pk_val
--- a/web/pgadmin/tools/sqleditor/utils/tests/test_save_changed_data.py
+++ b/web/pgadmin/tools/sqleditor/utils/tests/test_save_changed_data.py
@@ -1040,3 +1040,87 @@ class TestSaveAddedRowSkipsNonEditableColumn(TestSaveChangedData):
             "FROM {0};"
         ).format(self.test_table_name)
         utils.create_table_with_query(self.server, self.db_name, create_sql)
+
+
+class TestSaveUpdatedRowSkipsNonEditableColumn(TestSaveChangedData):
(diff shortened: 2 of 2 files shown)


dpage:
@coderabbitai review

dpage:
Rebased this branch onto current `upstream/master` and dropped 5 commits that had ridden along from an unrelated local development stack: #10214 (Helm existingSecret fix), #10251 (REINDEX CONCURRENTLY fix), #9450 (role admin-option membership feature), #10309 (username-import fix, already merged as part of #10320), and #8991 (server-cursor fix). None of those were part of this PR's actual change; the branch now carries only the one commit for the alias/expression column exclusion.

As a result, the diff no longer touches `roles/__init__.py`, `web/pgadmin/tools/maintenance/`, `utils/__init__.py`, or `psycopg3/connection.py` - all four of CodeRabbit's outstanding review threads were on code from those dropped commits, not on this PR's own change, so I've resolved them as moot. Force-pushed the cleaned-up branch.

Issue #9450, opened 2025-12-10
Group Role dialog doesn't allow users "with admin" option to add other users


TownCube:
**Describe the bug**

The "Group Role" dialog doesn't allow users who have been granted "with admin" to add other members to the role as the + icon is not displayed.

**To Reproduce**

1. Expand Servers > Server > Login/Group Roles
2. Right click on a role and choose "Properties"
3. Switch to the "Membership" tab
4. No plus icon is not shown to allow a user "with admin option" to add new members.

**Expected behavior**

Users granted "with admin" should see the + icon to allow adding other members to the role.

**Error message**

The user can't add role members because the plus icon isn't shown. However if the user "with admin" option attempts to grant "with admin" to another user then they incorrectly get the error "The current user does not have permission to update the role."

**Screenshots**

N/A

**Desktop (please complete the following information):**
 - OS: Fedora Linux 43
 - pgAdmin version: 9.10
 - Mode: Server
 - Browser (if running in server mode): Chromium
 - Package type: Python

**Additional context**

The [role UI only allows editing if the user is a super user or has the "create role" permission](https://github.com/pgadmin-org/pgadmin4/blob/dc801e362ebde7fe59ecf77eaaab740fe7452cc5/web/pgadmin/browser/server_groups/servers/roles/static/js/role.ui.js#L54), however if the user is a member of the role "with admin" option, they don't need either of these to be allowed to add new members to the role.

Issue #8991, opened 2025-07-25
Result grid does not move messages tab when commit/rollback button is clicked with server cursor on.


pravesh-sharma:
**Describe the bug**
If we click on commit or rollback button with server cursor, empty result grid is shown it should move to the messages tab same as on the case of normal execution

**Screenshots**

https://github.com/user-attachments/assets/0aa96a9f-e2fe-4251-8130-4dcc39c3739c

**Desktop (please complete the following information):**
 - OS: Macos Sequoia
 - Version: 9.6
 - Mode: Desktop

dpage:
Dug into this, and the actual bug goes a fair bit deeper than the UI symptom.

With "server cursor" mode on, the Commit/Rollback buttons run `BEGIN;`/`COMMIT;`/`ROLLBACK;` through `execute_void()`, which reuses whatever cursor is cached on the connection — and once any query has run, that's the named/server-side cursor psycopg created for it. A named cursor's `execute()` always wraps the statement as `DECLARE ... CURSOR FOR <query>`, which can't express a transaction-control statement (`DECLARE ... CURSOR FOR COMMIT` is a syntax error), and on top of that pgAdmin's own cursor wrapper always forwards a `prepare=` keyword the server-side cursor's `execute()` doesn't accept at all, so it actually failed one step earlier with `TypeError: keyword not supported: prepare`.

That exception was swallowed by a blanket `except Exception` in the background query thread, so the COMMIT/ROLLBACK never reached the database at all. The empty result grid this issue describes is the visible fallout: the next `/poll` call picked up the previous query's leftover column info rather than switching to the Messages tab.

In other words, with server cursor mode enabled, clicking Commit or Rollback didn't actually commit or roll back anything — the transaction was left open with no error shown to the user, which is a correctness issue, not just a cosmetic one.

Fixed by routing BEGIN/COMMIT/ROLLBACK through a throwaway plain cursor instead of the cached server-side one when server cursor mode is active, and clearing the stale column info so the app correctly reports no result set afterwards.

PR #10331, opened 2026-08-19
Skip virtualisation for small DataGridView grids to avoid a re-measure on tab show

### LINKED ISSUE #10143: Switching back to a dialog tab containing a large data grid is slow — DataGridView re-measures every row on show
**Describe the bug**

In SchemaView dialogs, each field group is rendered as a tab and inactive
tabs are kept mounted but hidden with `display: none`. A DataGridView
(collection grid) inside such a tab is always virtualized. The virtualizer's
`measureElement` runs a synchronous `getBoundingClientRect()` on every
rendered row through a fresh ref callback on each render.

While a tab is hidden, its scroll viewport measures 0, so the virtual window
collapses. When you switch back to that tab, the virtualizer's ResizeObserver
fires as the viewport goes from 0 to its real height, triggering a re-render
and a re-measure of every row — a forced-layout settle loop. On grids with
many rows of editable cell controls (react-select / switch / etc.), this
produces a visible lag and an apparent "reload" of the grid every time the
tab is shown again. At small row counts virtualization provides no benefit,
so this is pure overhead.

**To Reproduce**

1. Open a dialog that contains an editable data grid inside a tab — for
   example, right-click a table with many columns (30+) and open Properties,
   then go to the "Columns" tab.
2. Let the Columns grid render.
3. Switch to another tab (e.g. "Constraints"), then switch back to "Columns".
4. Observe a noticeable delay before the grid rows repaint each time you
   return to the tab (the grid appears to re-load its rows).

The effect grows with the number of rows and with the complexity of the
cell editors, and repeats on every switch back to the tab.

**Expected behavior**

Switching between tabs should be instant. A grid that has already rendered
should not re-measure and re-render all of its rows when its tab is shown
again.

**Error message**

No error message is produced — this is a UI performance/jank issue, not a
thrown error. Nothing is logged to the console.

**Screenshots**

(Attach a short screen recording showing the delay when switching back to
the tab.)

 - OS: macOS
 - pgAdmin version: [e.g. 8.x — fill in]
 - Mode: [Desktop or Server]
 - Browser (if running in server mode): [e.g. Chrome]
 - Package type: [e.g. Python, Container, etc.]

**Additional context**

Root cause: `DataGridView` (web/pgadmin/static/js/SchemaView/DataGridView/
grid.jsx) unconditionally virtualizes rows and uses `measureElement`
(getBoundingClientRect per row). Combined with tabs that hide content via
`display: none` (TabPanel), showing a tab again forces a full re-measure of
every rendered row.

Suggested fix: skip virtualization below a row-count threshold. Render every
row in normal document flow (no `measureElement`, no absolute positioning)
so hide/show becomes a pure CSS toggle; keep virtualization for large grids.
A minimal patch is available (adds a `virtualiseThreshold`, default 100, and
a `pgrt-row--static` CSS rule that opts small-grid rows out of the
virtualizer's absolute positioning).

dpage:
## What this is

SchemaView dialogs keep inactive tabs mounted with `display: none`. Every `DataGridView` collection grid is always virtualised via `@tanstack/react-virtual`, with `measureElement` doing a synchronous `getBoundingClientRect()` on every row through a fresh ref callback. Whilst a tab is hidden its scroll viewport measures 0, so the virtualizer's `ResizeObserver` sees a 0-to-real-height jump when the tab is shown again and treats it as a resize, re-measuring every row from scratch — slow for large grids, and pure overhead for small ones that had no offscreen window to skip in the first place.

Fixes #10143.

## Fix

Added a `virtualiseThreshold` (overridable via `viewHelperProps.virtualiseThreshold`, matching the existing `virtualiseOverscan`). Grids at or under the threshold skip virtualisation entirely: no `measureElement` ref, rows render via a plain `.map()` in normal document flow, and a new `pgrt-row--static` class overrides the row's usual `position: absolute`. Hide/show for these grids is now a pure CSS toggle with nothing for the virtualizer to remeasure. Grids above the threshold are unchanged.

The default threshold scales with visible column count (`min(400, max(25, round(700 / cols)))`) rather than being a flat row count, since render cost tracks total cells (rows × cols), not rows alone. That formula and its bounds are adapted from @VIBVEL47's independent fix for the same issue in #10146 — credit there, folded in here to keep one PR moving rather than two duplicates. Thank you for the fix, @VIBVEL47.

## Test plan

- [x] Added two tests in `SchemaDialogView.spec.js`: a small grid renders all rows statically (`pgrt-row--static`, no `transform`); a large grid (well above the threshold's clamp) still windows via the virtualizer.
- [x] `yarn run test:js-file SchemaDialogView` — 22/22 passed.
- [x] `yarn run test:js-file SchemaView` — 26/26 passed.
- [x] `eslint` clean on all changed files.

## Summary by CodeRabbit

* **Performance**
  * Improved data grid rendering for smaller datasets with a simpler, non-virtualized layout.
  * Large datasets continue to use virtualization for efficient scrolling.
  * Virtualization thresholds now adapt to the number of visible columns.

* **Bug Fixes**
  * Improved row positioning and rendering consistency for smaller data grids.

* **Tests**
  * Added coverage for static rendering of small grids and virtualization of large grids.

### DIFF
--- a/web/pgadmin/static/js/SchemaView/DataGridView/grid.jsx
+++ b/web/pgadmin/static/js/SchemaView/DataGridView/grid.jsx
@@ -122,12 +122,25 @@ export default function DataGridView({
     )
   ).includes(true);
 
+  // Virtualising a small grid buys nothing (there's no offscreen window to
+  // skip rendering) but still pays for measureElement's per-row
+  // getBoundingClientRect on every mount/remeasure. That remeasure is
+  // exactly what fires when a dialog tab holding the grid is hidden via
+  // `display: none` and then shown again, since the scroll viewport
+  // momentarily measures 0 and the virtualizer's ResizeObserver treats
+  // that as a real resize. Below the threshold we skip virtualisation
+  // entirely and render every row in normal document flow, so showing a
+  // hidden tab is a pure CSS toggle again.
+  const virtualiseThreshold = viewHelperProps.virtualiseThreshold ?? 100;
+  const shouldVirtualise = rows.length > virtualiseThreshold;
+
   const virtualizer = useVirtualizer({
     count: rows.length,
     getScrollElement: () => tableEleRef.current,
     estimateSize: () => 50,
     measureElement:
-      typeof window !== 'undefined' &&
+      shouldVirtualise &&
+        typeof window !== 'undefined' &&
         navigator.userAgent.indexOf('Firefox') === -1
         ? element => element?.getBoundingClientRect().height
         : undefined,
@@ -152,22 +165,29 @@ export default function DataGridView({
               ref={tableEleRef} table={table} data-test="data-grid-view"
               tableClassName='DataGridView-table'>
               <PgReactTableHeader table={table} />
-              <PgReactTableBody style={{
-                height: virtualizer.getTotalSize() + 'px'
-              }}>
+              <PgReactTableBody style={
+                shouldVirtualise ? {height: virtualizer.getTotalSize() + 'px'} : undefined
+              }>
                 {
-                  virtualizer.getVirtualItems().map((virtualRow) => {
+                  (
+                    shouldVirtualise
+                      ? virtualizer.getVirtualItems()
+                      : rows.map((_row, index) => ({index, start: 0}))
+                  ).map((virtualRow) => {
                     const row = rows[virtualRow.index];
                     return (
                       <PgReactTableRow
                         key={row.id}
                         data-index={virtualRow.index}
-                        ref={node => virtualizer.measureElement(node)}
-                        style={{
-                          // This should always be a `style` as it changes on
-                          // scroll.
-                          transform: `translateY(${virtualRow.start}px)`,
-                        }}
+                        ref={shouldVirtualise ? node => virtualizer.measureElement(node) : undefined}
+                        className={shouldVirtualise ? undefined : 'pgrt-row--static'}
+                        style={
+                          shouldVirtualise ? {
+                            // This should always be a `style` as it changes
+                            // on scroll.
+                            transform: `translateY(${virtualRow.start}px)`,
+                          } : undefined
+                        }
                       >
                         <GridRow
                           rowId={virtualRow.index} isResizing={isResizing}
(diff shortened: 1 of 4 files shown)


dpage:
@coderabbitai review

dpage:
Rebased this branch onto current `upstream/master` and dropped 5 commits that had ridden along from other in-progress work but were never part of this PR's actual change: #10214 (Helm `existingSecret` fix), #10251 (REINDEX CONCURRENTLY fix), #9450 (role-membership ADMIN OPTION feature), #10309 (username-import fix, already merged upstream), and #8991 (server-cursor fix). The branch now carries exactly one commit: the DataGridView re-measure fix (#10143).

As a result, the diff no longer touches `roles/__init__.py` or `psycopg3/connection.py`, so I've resolved the two open CodeRabbit threads on those files since neither concerns code this PR still touches.

dpage:
Fixed the `run-javascript-tests (windows-latest)` failure.

The 'virtualises a large grid' test mounted 450 rows so that it would clear the default threshold whatever the visible column count turned out to be, which took about sixteen seconds on Linux and tipped over Jest's eighteen second budget on the Windows runners. The aborted render then left `act()` warnings behind that failed the next test in the file too, which is why two tests went red rather than one.

The grid only needs enough rows to clear whatever threshold is in force, so the test now pins the threshold low through `viewHelperProps` and mounts sixty rows, exercising the same windowing in about five seconds. Rather than lose the coverage of the scaling itself, the formula is now `getVirtualiseThreshold()` in its own right, with tests covering the scaling, both bounds and the no-columns-yet case; those run in a fraction of a second. Full spec file and eslint both pass locally.

dpage:
Rebased onto current `upstream/master` and picked up the remaining nitpicks from CodeRabbit's review of 20 August:

- **Threshold boundary and override** (`SchemaDialogView.spec.js`): the override was already covered by the windowing test, which pins `viewHelperProps.virtualiseThreshold` low; 05583e2d2 adds the boundary, a grid of exactly the threshold that must stay static (checked that it fails if the comparison is changed to `>=`).
- **Concurrent schema reindexing** (`test_maintenance_create_job_unit_test.py`): no longer applies, since that file belonged to one of the unrelated commits dropped from this branch in the earlier rebase.

PR #10332, opened 2026-08-19
Fix inherited-column read-only state and Definition-tab type restriction in the Table dialog

### LINKED ISSUE #10179: Table dialog: inherited columns are editable/deletable and aren't stripped when the parent is removed
## Describe the bug

In the Create/Properties dialog for a table that inherits from one or more parent tables, inherited columns are not handled correctly in the **Columns** grid:

1. **On opening an inheriting table**, the inherited columns are shown as fully editable — their Name / Data type cells and the row edit/delete buttons are enabled — instead of read-only.
2. **When a parent is added interactively** via "Inherited from table(s)", the fetched columns are added to the grid but are also fully editable (not marked inherited).
3. **When a parent is removed**, its columns are (in the fixed SQL case) dropped, but the remaining local columns' edit/delete buttons incorrectly become disabled.

Root causes (all in the shared SchemaView / table dialog code, so this is upstream pgAdmin, not distro-specific):

- The column properties query hardcodes the child table OID in the `pg_inherits` join, so `inheritedfrom`/`inheritedid` are never populated:
  `web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/columns/sql/default/properties.sql:8` → `... AND ph.inhrelid = 33896::oid` (should use the `{{tid}}` parameter already used on the same template's `WHERE att.attrelid = {{tid}}::oid`). This is the `default` template (no version-specific override for columns), so it affects **all supported PostgreSQL versions**, not just PG 14+.
- Interactively adding a parent fetches its columns via the columns endpoint (which returns them as local, since the parent doesn't itself inherit) and never stamps `inheritedfrom`/`inheritedid` on them.
- The column's cell-editability check (`inSchemaWithColumnCheck`) returns "editable" for any new row *before* it checks `inheritedfrom`, so interactively-added inherited columns stay editable.
- The row edit/delete cells read their options without subscribing to option changes, so after the columns array is modified and rows shift index, they keep a stale enabled/disabled state.

## To Reproduce

1. In the Query Tool, create an inheritance hierarchy:
   ```sql
   CREATE TABLE parent_a (a_id int, a_name text);
   CREATE TABLE child_t (own_col text) INHERITS (parent_a);
   ```
2. Right-click **child_t** → **Properties** → **Columns** tab.
3. Observe the inherited columns `a_id`, `a_name` — they are editable (Name/Data type not greyed, edit/delete enabled).
4. Add another table (e.g. `pg_amproc`) in **Inherited from table(s)** → its columns appear but are also editable.
5. Remove `parent_a` from **Inherited from table(s)** → the local column `own_col`'s edit/delete buttons become disabled.

## Expected behavior

- Inherited columns (from any parent, whether present on open or added interactively) are shown **read-only**: Name / Data type cells greyed and the row edit/delete buttons disabled.
- Local columns (e.g. `own_col`) remain fully editable/deletable, including after a parent is de-selected.
- De-selecting a parent removes only that parent's inherited columns; local columns remain.
- `inheritedfrom` and `inheritedid` are populated for inherited columns in the column properties response.

## Error message

No error is thrown — this is a silent UI/logic issue. The underlying cause is visible in the column properties SQL: the `INH_TABLES` CTE joins `pg_inherits` on a hardcoded OID (`33896`) instead of `{{tid}}`, so every column returns `inheritedfrom = NULL` / `inheritedid = NULL`.

## Screenshots

Compare the Columns tab of an inheriting table: inherited columns' Name/Data type should be greyed and edit/delete disabled (like other tools), but currently render editable. (Attach a before/after of the Columns grid showing inherited columns editable vs. correctly disabled.)

## Desktop

- OS: macOS
- pgAdmin version: 9.16
- Mode: Server
- Browser (if running in server mode): Chrome
- Package type: Python (running from source)

## Additional context

- The hardcoded OID was introduced in commit `7b9b8cdd0` ("Rebase the version-specific SQL templates to target PostgreSQL 14 (#10054)"), so the read-only regression is recent; the interactive add/remove gaps are longer-standing.
- Affected areas: `tables/templates/columns/sql/default/properties.sql`, `tables/columns/static/js/column.ui.js` (`inSchemaWithColumnCheck`), `tables/static/js/table.ui.js` (inherit add/remove), and the shared `SchemaView/DataGridView` edit/delete cells (`deletable.js`, `expandabledFormView.jsx`) which read row options without subscribing.

### LINKED ISSUE #10180: Table dialog: "Data type" dropdown in the expanded Definition tab shows all types, ignoring the allowed-type restriction applied inline
## Describe the bug

In the Table Create/Properties dialog, the **Columns** grid exposes the "Data type" field through two editors that present **inconsistent option lists** for the same column:

- The **inline grid-cell** "Data type" dropdown correctly applies the allowed-type restriction. For an existing column it offers only the types the column can be altered to (the `edit_types` subset).
- The **expanded row's Definition tab** "Data type" dropdown offers the **full superset** of types, ignoring the restriction — so it lets you pick types the column cannot be changed to.

Both editors are backed by the same `cltype` field, so they should offer the same list. In practice they disagree: the inline cell is filtered (correct), the Definition-tab control is unfiltered (wrong).

Root cause (shared SchemaView code, so this is upstream pgAdmin, not distro-specific):

- The `cltype` field defines the same `edit_types` filter for both its `cell` (inline) and `type` (form) editors. The filter only applies when the column is not new: `if (!isNew(state) && !inErd) { filter to edit_types }`.
- For the inline cell, the filter receives the **row's data** (with `cid`, `attnum`, `edit_types`), so `isNew()` correctly returns `false` and the list is filtered.
- For the expanded Definition-tab control, `MappedFormControl` evaluates the field's `type(state)` callback with `state = schemaState.data` — the **root (table) data**, not the field's containing row. That root object has no `cid`/`attnum`/`edit_types`, so `isNew()` returns `true`, the filter is skipped, and the full list is shown.
  - `web/pgadmin/static/js/SchemaView/MappedControl.jsx` (`MappedFormControl`): `const state = schemaState.data;` then `evalFunc(null, field.type, state)`.
  - It should evaluate against the object that *contains* the field (the field's parent accessPath) — which is the root for a top-level field, and the row object for a nested collection form.

## To Reproduce

1. Connect to a PostgreSQL server and create a table with at least one column, e.g.:
   ```sql
   CREATE TABLE public.t_types (c1 integer);
   ```
2. Right-click **t_types** → **Properties** → **Columns** tab.
3. On the existing column `c1`, open the **Data type** dropdown directly in the grid cell → note the limited (restricted) list of types.
4. Click the edit (pencil) icon to expand the same row, go to the **Definition** tab, and open the **Data type** dropdown there → note the much larger, unrestricted list.
5. The two lists disagree for the same column.

## Expected behavior

Both "Data type" dropdowns (inline grid cell and the expanded Definition tab) present the **same** list for a given column — i.e. the expanded Definition-tab control applies the same `edit_types` restriction as the inline cell for an existing column, and both show the full list for a new column.

## Error message

No error is thrown — this is a silent inconsistency. It is visible via the field callback receiving the wrong state: in the expanded form the `cltype` filter sees `isNew = true` and `edit_types = []` (because it is handed the root/table data), so it returns all types; the inline cell sees `isNew = false` with the real `edit_types` and returns the restricted subset.

## Screenshots

Compare, for the same existing column, the inline grid-cell "Data type" dropdown (restricted list) against the expanded Definition-tab "Data type" dropdown (full list). (Attach both dropdowns open side by side.)

## Desktop

- OS: macOS
- pgAdmin version: 9.16
- Mode: Server
- Browser (if running in server mode): Chrome
- Package type: Python (running from source)

## Additional context

- Affected area: shared SchemaView form control — `web/pgadmin/static/js/SchemaView/MappedControl.jsx` (`MappedFormControl`), which evaluates every field's `type(state)` callback against the root data instead of the field's containing object. This affects any nested collection form whose field `type`/other callbacks depend on sibling values in the same row (the Table → Columns "Data type" editor is the visible case).
- Fix direction: evaluate `field.type` against `schemaState.value(accessPath.slice(0, -1))` (the containing object). For a top-level field this equals the root data (unchanged behavior); for a nested row it correctly resolves to the row's data.

dpage:
## What this is

Two related bugs in the Table Create/Properties dialog's Columns grid.

**Inherited columns editable/deletable (#10179).** `column.ui.js`'s `inSchemaWithColumnCheck` and `table.ui.js`'s `canEditDeleteRowColumns` had two mismatches:
- When a table is opened already inheriting, the properties fetch sets `inheritedfromtable` on inherited columns, but the frontend only ever checked `inheritedfrom` — so those rows were never recognised as inherited.
- When a parent is added interactively via "Inherited from table(s)", the fetched rows do carry `inheritedfrom`, but carry no `attnum` yet, and the inherited check ran *after* an `isNew()` short-circuit — so these attnum-less rows were still treated as fully editable/deletable.

Removal-on-detach already worked via the existing `deferredDepChange` pruning logic, confirmed by pre-existing tests, so no change was needed there.

**Definition-tab type dropdown ignoring the restriction (#10180).** The inline grid-cell editor and the expanded row's Definition tab each had their own copy of the `edit_types` filter for `cltype`, but `MappedControl` resolves a field's `type()` callback against the whole table's data rather than the field's own row (unlike `cell()`, which already gets the row) — so the tab's `isNew()`/`edit_types` checks always read the wrong object and no-opped, showing every type.

Fixes #10179, fixes #10180.

## Fix

- Check `inheritedfrom`/`inheritedfromtable` before the `isNew()` short-circuit in both `inSchemaWithColumnCheck` and `canEditDeleteRowColumns`.
- `MappedControl` now resolves a field's declared `deps` against its own row (as `listenDepChanges` already does for `cell`) and forwards them as a second argument to `type()`. `column.ui.js` declares `edit_types`/`attnum` as `deps` on `cltype` and factors the filter into one shared `editTypesFilter()` used by both `cell` and `type`, so the two can't drift apart again.

## Test plan

- [x] Extended `column.ui.spec.js` and `table.ui.spec.js` to cover both cases.
- [x] `yarn run jest schema_ui_files/table.ui.spec.js schema_ui_files/column.ui.spec.js schema_ui_files/catalog_object_column.ui.spec.js` — 36/36 passed.
- [x] Full `test:js-once` (152 suites / 948 tests) passed with the `MappedControl.jsx` change in place.
- [x] `eslint` clean on all changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Inherited table columns are now correctly protected from editing and deletion, whether loaded from a parent table or added interactively.
  - Inherited column length and precision settings remain read-only when appropriate.
  - Column data-type options now consistently respect editing rules across inline and expanded editors.
  - New-row and diagram-editing scenarios now display the appropriate data-type choices.

- **Tests**
  - Added regression coverage for inherited-column protection and consistent data-type editing behavior.


dpage:
@coderabbitai review

dpage:
Cleaned up this branch and worked through the outstanding review threads:

- Rebased onto current `upstream/master` and dropped 5 commits that had ridden along from my personal dev stack but never belonged to this PR (#10214 Helm fix, #10251 REINDEX fix, #9450 role-membership feature, #10309 username-import fix — already merged upstream, and #8991 server-cursor fix). The branch now carries only this PR's own two commits (#10179, #10180) plus a small follow-up.
- Fixed the `attlen`/`attprecision` editable-check finding in `column.ui.js`: those callbacks only checked `inheritedfrom`, so a loaded row carrying just `inheritedfromtable` could be left incorrectly editable. Added a shared `isInheritedColumn()` helper used by both callbacks (and by `inSchemaWithColumnCheck`), plus a regression test covering the `inheritedfromtable`-only case.
- Resolved the other three open threads (on `roles/__init__.py`, `utils/__init__.py`, `psycopg3/connection.py`) — they were about code from the dropped commits, which no longer appears in this branch's diff.

kundansable (approved):

PR #10342, opened 2026-08-25
Fix invalid SQL from index TABLESPACE clause on partitioned tables

### LINKED ISSUE #10341: Generated SQL for indexes invalid for partitioned tables
When pgAdmin 4 generates the SQL definition for an index on a partitioned table, it includes a TABLESPACE clause.

For example, pgAdmin may generate a definition similar to:

CREATE INDEX my_index
    ON my_partitioned_table (my_column)
    TABLESPACE my_tablespace;

This is incorrect for a partitioned index. PostgreSQL does not allow specifying a tablespace when creating an index on a partitioned table. The tablespace must instead be defined on the indexes of the individual partitions, if required.

As a result, the SQL definition generated by pgAdmin cannot be executed directly against PostgreSQL and produces an error.

Image

Image

dpage:
## Summary

- PostgreSQL rejects an explicit `TABLESPACE` clause naming the database's own default tablespace when the index belongs to a partitioned table (`cannot specify default tablespace for partitioned relations`), so the SQL pgAdmin generated for a new index on a partitioned table could not be executed whenever the Tablespace field was left at its default (`pg_default`).
- The `index_constraint` and `exclusion_constraint` create.sql templates already guard against emitting this redundant clause (`data.spcname != "pg_default"`); this extends the same guard to the plain index `create.sql` templates (`default` and `15_plus` buckets).
- Updated all affected fixture files under `indexes/tests/{default,15_plus}` to match the corrected output.

## Test plan

- [x] Added `test_create_sql_partitioned_tablespace.py`, a DB-free unit test rendering the templates directly and asserting the redundant clause is omitted while an explicit non-default tablespace is still emitted.
- [x] `regression/runtests.py --pkg browser.server_groups.servers.databases.schemas.tables.indexes.tests` — 49/49 passing.
- [x] `regression/runtests.py --pkg browser.server_groups.servers.databases.schemas.tables.tests` — 113/113 passing (no regression in the parent Tables SQL, which embeds indexes).
- [x] Verified against a live PostgreSQL 18 server that `CREATE INDEX ... TABLESPACE pg_default` on a partitioned table errors, while a real non-default tablespace works fine.

Closes #10341

## Summary by CodeRabbit

- **Bug Fixes**
  - Index creation now omits the explicit default tablespace setting when using `pg_default`, allowing partitioned indexes to be created successfully.
  - Explicit non-default tablespace selections continue to be preserved.

- **Tests**
  - Added regression coverage for default and custom tablespaces across supported PostgreSQL versions.
  - Updated index creation scenarios to reflect the corrected default-tablespace behavior.


asheshv:
The bug is real, but I think this fix makes things worse in one common configuration, so please don't merge it yet.

**The core problem: `pg_default` isn't the same thing as "the database's default tablespace".**

`indexes/sql/default/properties.sql:13-24` derives `spcname` from `pg_database.dattablespace` — that is, **the name of whatever tablespace the database defaults to**, which is `pg_default` only in the common case. PostgreSQL's own check in `DefineIndex`/`DefineRelation` is `tablespaceId == MyDatabaseTableSpace` — also the database default, not the literal `pg_default`. This PR hardcodes the string `pg_default`, so the two disagree the moment a database was created with `CREATE DATABASE x TABLESPACE ts_foo`.

Two consequences, both bad:

**1. Silent wrong tablespace placement (`indexes/sql/default/create.sql:23`, `indexes/sql/15_plus/create.sql:29`) — this is the serious one.**
With the database default at `ts_foo`, an index the user *deliberately* put in `pg_default` stores `reltablespace = 1663`, so `properties.sql` reports `spcname = 'pg_default'` and the guard suppresses the clause. The reverse-engineered SQL (and Schema Diff) then recreates that index in `ts_foo` instead. Data silently lands in the wrong tablespace — strictly worse than the visible error this PR set out to fix.

**2. In that same setup, the original bug isn't fixed.**
An index inheriting the `ts_foo` default reports `spcname = 'ts_foo'`, the clause is still emitted, and `CREATE INDEX ... TABLESPACE ts_foo` on a partitioned table still fails with `cannot specify default tablespace for partitioned relations`.

Both fall out of the same fix: compare against the database's *actual* default tablespace rather than the literal string. `properties.sql` already computes the value — surface it into the render context (e.g. `db_default_tablespace`) and guard on `data.spcname != db_default_tablespace`.

**Also still broken: partitioned tables (`tables/sql/default/create.sql:131-132` and the `16_plus` sibling).**
Identical unguarded `{% if data.spcname %}TABLESPACE ...`, and `tests/pg/default/create_table_with_range_partition.sql:10,19` bakes in `PARTITION BY RANGE (arr); ... TABLESPACE pg_default;` — which PostgreSQL rejects for exactly the same reason. Creating a partitioned *table* is the more common path than a partitioned index, so either extend the fix or scope the PR explicitly and file a follow-up.

**The suppression isn't limited to partitioned relations.** It strips the clause from ordinary indexes' reverse-engineered SQL too. There's no partitioned indicator in the index template context today (only `data.indisonly`), so a targeted fix needs a bit of backend plumbing. If that's not worth it, please at least state the tradeoff in the commit message rather than leaving it implicit.

**Cosmetic regression baked into the fixtures (32 files).** Removing the clause leaves a stray blank line between `WITH (...)` and `WHERE`, and in the no-`WHERE` case a bare `;` on its own line (e.g. `indexes/tests/default/create_btree_asc_null_last.sql:10`, `create_hash_index.sql:8`). Rather than updating fixtures to match, fix the whitespace with `{%- ... -%}` trims around the block.

**The new test doesn't test what it looks like it tests (`test_create_sql_partitioned_tablespace.py:47-51`).** It installs its own `FileSystemLoader`, bypassing `VersionedTemplateLoader`, so it proves nothing about version-bucket selection — which matters here, since there's no `indexes/sql/16_plus/create.sql` and PG 16-17 fall through to `15_plus` (correct, but untested). It also only renders `mode='create'`, never the reverse-engineered path, and has no case for a non-`pg_default` database default — the very scenario that breaks. (FWIW I checked the `_FakeApp` / `get_driver` trick for test pollution: `DriverRegistry.get` is memoized and neither driver defines `init_app`, so it's a harmless no-op.)

No injection concern — `data.spcname` still goes through `conn|qtIdent` and this is a comparison change only.

hiteshjambhale (approved):
Tested manually on master and PR branch.
Master: Creating an index on a partitioned table with pg_default tablespace generates TABLESPACE pg_default in the SQL — Postgres rejects it with "cannot specify default tablespace for partitioned relations".
PR branch: TABLESPACE pg_default is correctly omitted — index creates successfully. Custom tablespaces (e.g. eav) still appear in the SQL as expected.

PR #10343, opened 2026-08-25
Fix server-cursor queries silently failing with 'keyword not supported: prepare'

dpage:
## Problem

Every query run under "Use server cursor?" currently fails silently.

`AsyncServerCursor.execute()` (psycopg) has never accepted a `prepare` keyword — a server-side `DECLARE CURSOR` can't be a prepared statement — and raises `TypeError` on any unexpected keyword, even one whose value is `None`.

#10030 widened `AsyncDictCursor.execute()`/`_execute()` to forward `prepare`/`binary` to whatever cursor it holds, correctly fixing a `psycopg_pool.ConnectionPool.check_connection` breakage. `AsyncDictServerCursor` inherits that same `execute()`/`_execute()` without overriding it, so every server-cursor query now unconditionally forwards `prepare=None` into `AsyncServerCursor.execute()` and fails with:

```
TypeError: keyword not supported: prepare
```

That `TypeError` isn't a `psycopg.Error`, so `execute_async()`'s `except psycopg.Error` doesn't catch it. It propagates into the background `QueryThread`'s generic exception handler, which logs it and builds an `internal_server_error` response that's discarded (the thread's return value goes nowhere) — so the query silently never runs. The user sees no error; the async cursor is left with no query ever actually executed on it, which is also the precondition behind the crash in #10321 (`get_explain_query_length` being handed a cursor whose `_query` is `None`).

## Fix

Give `AsyncDictServerCursor` its own `_execute()` that drops `prepare` before delegating, leaving `DictCursor`/`AsyncDictCursor`'s forwarding (the actual #10030 fix) untouched for the pool-checkout case it was written for.

## Tests

- Strengthened `test_server_cursor.py`'s existing scenario to assert the poll response's actual `status`/`result`, rather than only the HTTP status code and the echoed `server_cursor` flag (which is just config, not proof the query ran) — this is what let the regression through un-caught.
- Added a fast, DB-less unit test asserting `AsyncDictServerCursor._execute` never forwards `prepare` to the underlying cursor.

Both fail without the fix and pass with it; ran locally against PostgreSQL 18.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed server-cursor query execution with the asynchronous PostgreSQL driver.
  * Queries now execute successfully while preserving binary-result handling and parameters.
  * Improved validation to confirm server-cursor queries return the expected results.

* **Tests**
  * Added regression coverage for server-cursor execution and parameter handling.


hiteshjambhale:
Tested manually with server cursor enabled on both master and the PR branch.

Master: SELECT 1 runs silently — Messages says "Query returned successfully" but Data Output is empty. No error shown to the user.

PR branch: Same query returns 1 as expected. Fix confirmed.

hiteshjambhale (approved):

Issue #10334, opened 2026-08-19
Unable to run pgadmin4 web version with reverse proxy using Caddy


ranjanport:
When I try to run the web version of pgadmin4 behind the reverse proxy using caddy, the loading spinner comes and revolve continuously. The login page not come and hence unable to use pgadmin4 behind reverse proxy. While the login page is fully functional if I use my `IP:PORT/pgadmin4`

**To Reproduce**

1. Install Caddy
2. Add the config in /etc/caddy/Caddyfile (I have replaced the actual domain)
```
pgadmin.example.com {
        reverse_proxy 127.0.0.1:5433
}
```
where I have initialized my pgadmin4 server on 5433

3. Try to open the domain on web ideally the login page should pop up and the pgadmin4 is accessible
4. But below error message is shown on console of developer tool. And originally the spinner rotate for too long time and nothing really comes up.

**Error message**

```
[Error] Failed to load resource: bad URL (v4513226cdae34746b4dedf0b4dfa099e1781791509496, line 0)
[Error] Refused to execute https://pgadmin.example.com/pgadmin4/static/vendor/require/require.min.js?ver=91700 as script because "X-Content-Type-Options: nosniff" was given and its Content-Type is not a script MIME type.
[Error] ReferenceError: Can't find variable: require
	Global Code (login:54)
[Log] ReferenceError: Can't find variable: require — login:113 (login, line 123)
[Warning] Source Map "https://pgadmin.example.com/pgadmin4/static/js/generated/pgadmin_commons.js.map" has SyntaxError: JSON Parse error: Unrecognized token '<'
```

** Environment**
 - OS: Linux
 - pgAdmin version: latest
 - Mode: [Server]
 - Browser (if running in server mode): [chrome, safari]
 - Package type: [DEB, Python, Container] attempted tests using these might impact other package type too.

**Additional context**
It is better if there should be a document section under deployment using reverse proxy using Caddy this will help most of the user running Caddy for there deployment.

dev-hari-prasad:
I'm looking into this and will try to reproduce the issue on my VPS. I'll share an update once I have a better understanding of what's causing it.

dev-hari-prasad:
Thanks for reporting this, @ranjanport. Before digging further, could you check two things:

1. **Network tab:**
[...]

PR #10344, opened 2026-08-25
docs: add Caddy reverse proxy configuration examples

dpage:
## Summary

- Adds Caddy examples alongside the existing NGINX ones for hosting pgAdmin with Gunicorn or uWSGI, both at the root of a server and under a sub-directory (mirroring the structure of the existing NGINX sections).
- Documents the two things that must line up for a sub-directory deployment behind any reverse proxy, not just Caddy: the proxy forwarding the request path unmodified with an `X-Script-Name` header, and a `SCRIPT_NAME` environment variable set on the pgAdmin process itself (a separate, static mechanism that sets `APPLICATION_ROOT`/`SESSION_COOKIE_PATH`).

Related: #10334, where the reported "login succeeds then bounces back to the login page" turned out to be caused by the missing `SCRIPT_NAME` process environment variable, not by cookie security flags.

## Test plan

- [x] `make docs` equivalent (targeted `sphinx-build`) with no new warnings for `server_deployment.rst`
- [x] Verified the documented Caddy config end-to-end locally: pgAdmin running under `/pgadmin4` behind a local Caddy instance, full browser login → dashboard → reload cycle with no console errors and no redirect loop

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for hosting pgAdmin in a sub-directory using the `SCRIPT_NAME` environment variable.
  * Updated NGINX and Caddy examples for root and `/pgadmin4` deployments.
  * Documented Unix-socket reverse proxying, path matching, URL prefix configuration, and header handling.
  * Clarified Caddy’s systemd `PrivateTmp=true` socket-access issue.
  * Clarified that uWSGI deployments with Caddy should use HTTP mode.
  * Updated deployment examples for Python 3.13 paths.

### DIFF
--- a/docs/en_US/server_deployment.rst
+++ b/docs/en_US/server_deployment.rst
@@ -331,6 +331,15 @@ And configure Caddy:
         reverse_proxy unix//tmp/pgadmin4.sock
     }
 
+.. note:: The systemd unit shipped with Caddy sets ``PrivateTmp=true``, which
+    gives the Caddy process a private ``/tmp`` directory of its own. A socket
+    created by Gunicorn or uWSGI in the system ``/tmp`` is therefore invisible
+    to Caddy, which will report a 502 error for every request. Place the
+    socket in a directory outside ``/tmp`` that both processes can access, for
+    example ``/run/pgadmin4/``, ensuring that the user Caddy runs as has
+    permission to read and write it; alternatively, have Gunicorn or uWSGI
+    listen on a TCP port such as ``127.0.0.1:5050`` and proxy to that instead.
+
 Alternatively, pgAdmin can be hosted in a sub-directory (/pgadmin4 in this
 case) on the server. Start Gunicorn as when using the root directory, but
 also set the ``SCRIPT_NAME`` environment variable for the Gunicorn process,
--- a/docs/en_US/server_deployment.rst
+++ b/docs/en_US/server_deployment.rst
@@ -305,6 +305,126 @@ Then, configure NGINX:
       uwsgi_pass unix:/tmp/pgadmin4.sock;
     }
 
+Caddy Configuration with Gunicorn
+----------------------------------
+
+pgAdmin can be hosted by Gunicorn, with Caddy in front of it as a reverse
+proxy. Note that these examples assume pgAdmin was installed using the Python
+Wheel (you may need to adjust the path to suit your installation).
+
+To run with pgAdmin in the root directory of the server, start Gunicorn using
+a command similar to:
+
+.. code-block:: bash
+
+    gunicorn --bind unix:/tmp/pgadmin4.sock \
+             --workers=1 \
+             --threads=25 \
+             --chdir /usr/lib/python3.7/dist-packages/pgadmin4 \
+             pgAdmin4:app
+
+And configure Caddy:
+
+.. code-block:: text
+
+    pgadmin.example.com {
+        reverse_proxy unix//tmp/pgadmin4.sock
+    }
+
+Alternatively, pgAdmin can be hosted in a sub-directory (/pgadmin4 in this
+case) on the server. Start Gunicorn as when using the root directory, but
+also set the ``SCRIPT_NAME`` environment variable for the Gunicorn process,
+so that pgAdmin knows to generate URLs, redirects (including the one issued
+after a successful login) and session cookies relative to the sub-directory,
+rather than to the root of the server:
+
+.. code-block:: bash
+
+    SCRIPT_NAME=/pgadmin4 gunicorn --bind unix:/tmp/pgadmin4.sock \
+             --workers=1 \
+             --threads=25 \
+             --chdir /usr/lib/python3.7/dist-packages/pgadmin4 \
+             pgAdmin4:app
+
+Then configure Caddy. The request path must be passed through to Gunicorn
+unchanged, so do not strip the /pgadmin4 prefix; a custom ``X-Script-Name``
+header must be set so that pgAdmin can identify requests made under the
+sub-directory; and the path matcher must include both ``/pgadmin4`` and
+``/pgadmin4/*``, so that pgAdmin's own redirect to the bare sub-directory
+path (with no trailing slash) is also proxied, rather than falling through
+to Caddy's default (and rather unhelpful) empty response for an unmatched
+path:
+
+.. code-block:: text
+
+    pgadmin.example.com {
+        @pgadmin path /pgadmin4 /pgadmin4/*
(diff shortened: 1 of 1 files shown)


asheshv:
The core of this is genuinely valuable — thanks for writing it up. I verified the central claim: `evaluate_config.py:161-164` reads `SCRIPT_NAME` from the **process environment** and uses it to set `APPLICATION_ROOT` and `SESSION_COOKIE_PATH`. That really is a separate mechanism from the per-request `X-Script-Name` handled in `pgAdmin4.py:67-72`, it really was undocumented, and it does explain the #10334 redirect loop.

Also good: the Caddy examples get HTTPS automatically, which is better than our existing nginx examples, and `X-Forwarded-For` is handled safely (`PROXY_X_FOR_COUNT = 1` reads the rightmost entry, so a client can't spoof it by prepending fake addresses).

The `run-feature-tests-pg (16)` failure is unrelated flake — a `StaleElementReferenceException` in the backup/restore Selenium test, on a docs-only change. Ignore it.

A few things to change before this lands:

**1. The examples are documented not to work (`server_deployment.rst:329-337`).**
Every example uses `/tmp/pgadmin4.sock`, and then the note right below explains that Caddy's shipped systemd unit sets `PrivateTmp=true`, so Caddy can't see that socket and "will report a 502 error for every request." That's backwards — we're shipping four configs that fail out of the box and then telling the reader to edit them. Please use `/run/pgadmin4/pgadmin4.sock` (or plain `127.0.0.1:5050`) in the examples themselves, and turn the note into an explanation of *why* `/tmp` is avoided.

**2. That `PrivateTmp` warning only appears in the Gunicorn section (`:329`)** but applies equally to the uWSGI examples that also use `/tmp/pgadmin4.sock`. Hoist it or cross-reference it.

**3. The nginx examples still have the bug this PR documents (`:246-252`, `:292-310`).**
The `SCRIPT_NAME` requirement applies to *any* reverse proxy — your PR description says so — but the existing nginx sub-directory examples were left untouched, so an nginx user walks straight into #10334. Either add `SCRIPT_NAME=/pgadmin4` there too, or lift the explanation into one shared note that all four sections point at.

**4. Please strip `X-Script-Name` in the root-directory examples (`:311-317`).**
`pgAdmin4.py:67` trusts the `X-Script-Name` header unconditionally — no trust count, no allowlist, unlike our `PROXY_X_*` settings. On a root-mounted deployment a client can send `X-Script-Name: /evil` and redirect every URL pgAdmin generates. To be clear: this hole is **pre-existing** and not something this PR introduces (nginx's `proxy_params` doesn't clear it either). But a brand-new documented config is the right place to close it — adding `header_up -X-Script-Name` to both root-directory `reverse_proxy` blocks costs nothing. Same argument applies to `X-Scheme` (`pgAdmin4.py:73`), which has no trust gate at all.

Minor:

- **`:~425-437`** — the uWSGI sub-directory example uses `--manage-script-name --mount` *and* sets `X-Script-Name`. I traced it: it works, because the second strip is a no-op. But it's redundant, and the nginx uWSGI example deliberately omits the header. Drop it, or add a sentence saying why it's harmless.
- **`:317, 355, 390`** — the new examples copy `/usr/lib/python3.7/dist-packages/pgadmin4` and `--workers=1` from the older nginx sections. Consistent with what's there, but worth refreshing while the file is open.

dpage:
Thanks @asheshv, all four are addressed in 9071ba86e, along with both minors.

**1 and 2.** The Caddy examples now use `/run/pgadmin4/pgadmin4.sock` themselves, so they work as shipped, and the note has become an explanation of why `/tmp` is avoided rather than an instruction to edit the example first. It is labelled and the uWSGI section points at it, as does a second note covering the header stripping below.

**3.** Rather than repeat the explanation in four places I have lifted it into a new "Hosting in a Sub-Directory" section ahead of the reverse proxy sections, which all four sub-directory examples now reference, and both NGINX sub-directory examples set `SCRIPT_NAME=/pgadmin4` accordingly. The NGINX uWSGI one needs it just as much as the others, since `--manage-script-name` only shapes the per-request WSGI environment.

**4.** Both root-directory Caddy examples now carry `header_up -X-Script-Name` and `header_up -X-Scheme`. I went a little further than you asked and stripped `X-Scheme` in the sub-directory examples too, because nothing sets it there either, so a client could otherwise supply its own; `X-Script-Name` in the Gunicorn sub-directory example is set rather than removed, which replaces whatever the client sent and so closes the same hole.

On the minors: the redundant `X-Script-Name` is gone from the uWSGI sub-directory example, matching the NGINX uWSGI one, and the illustrative path is now `python3.13`, since 3.7 has not been supported for a long time. I left `--workers=1` alone.

One thing worth flagging: I have not stripped `X-Script-Name` in the NGINX examples, on the grounds that this PR is documenting Caddy and the pre-existing NGINX configurations are a separate question, but the underlying hole is in `pgAdmin4.py:67` and `:73` rather than in any one proxy configuration. A trust gate on those two headers, along the lines of the existing `PROXY_X_*_COUNT` settings, would be the real fix and I am happy to raise it separately if you agree.

Docs build clean with `-W`.

dpage:
Merged to master as 2ba68de777319231e979c7d2965c0ff153380dee, squashed to a single commit, so GitHub will not close this automatically.

Both CodeRabbit findings were valid and are addressed: the "NGINX examples set `X-Script-Name`" sentence is now singular, since only the Gunicorn example does, and the socket note now says the directory must exist and be writable before the socket can be bound, with a pointer to `RuntimeDirectory=pgadmin4` for the systemd case.

One thing came out of review that was not in the original scope. pgAdmin honours `X-Script-Name` and `X-Scheme` unconditionally in `ReverseProxied.__call__`, unlike the `X-Forwarded-*` headers, which go through `ProxyFix` and the `PROXY_X_*_COUNT` settings; the Caddy examples defended against a client supplying its own, but the NGINX ones did not, which rather implied NGINX was safe. There is now a short `Reverse Proxy Headers` section covering both, and every example overwrites or clears both headers. For the `uwsgi_pass` cases that is `uwsgi_param HTTP_X_SCRIPT_NAME "";`, which works because a configured param suppresses the header nginx would otherwise generate from the client request, whilst an empty value is still emitted and so fails pgAdmin's truthiness check.

I also softened the claim that `SCRIPT_NAME` is *required* alongside `--manage-script-name`. I could not find a mechanism that makes it strictly necessary: uWSGI sets the per-request `SCRIPT_NAME` itself, so `url_for()` and the post-login redirect allow-list both behave. What it does buy is a session cookie confined to the sub-directory, which is what the text now says.

PR #10354, opened 2026-08-26
Fix MAINTAIN privilege silently dropped for views/mviews on PG17+

### LINKED ISSUE #10350: Views/Materialized Views: MAINTAIN privilege silently dropped on PG17+ (allowed_privs.json looked up at wrong path)
**Describe the bug**

`views/__init__.py:360` sets a fixed relative suffix for locating the allowed-privileges list:

```python
_ALLOWED_PRIVS_JSON = 'sql/allowed_privs.json'
```

used at `views/__init__.py:289`:

```python
self.allowed_acls = render_template(
    "/".join([self.template_path, self._ALLOWED_PRIVS_JSON])
)
```

For the `default` version bucket, the file genuinely lives under a `sql/` subdirectory (e.g. `templates/views/pg/default/sql/allowed_privs.json`), so this resolves correctly. But for the PG17+ bucket, the file was added directly under the bucket directory, with no `sql/` subdirectory:

```
web/pgadmin/browser/server_groups/servers/databases/schemas/views/templates/views/pg/17_plus/allowed_privs.json
web/pgadmin/browser/server_groups/servers/databases/schemas/views/templates/mviews/pg/17_plus/allowed_privs.json
web/pgadmin/browser/server_groups/servers/databases/schemas/views/templates/views/ppas/17_plus/allowed_privs.json
web/pgadmin/browser/server_groups/servers/databases/schemas/views/templates/mviews/ppas/17_plus/allowed_privs.json
```

So on PG17+/EPAS17+, `render_template` is asked for `views/pg/17_plus/sql/allowed_privs.json`, which doesn't exist. The failure is swallowed silently:

```python
try:
    self.allowed_acls = render_template(...)
    self.allowed_acls = json.loads(self.allowed_acls)
except Exception as e:
    current_app.logger.exception(e)
```

with no user-visible error — `self.allowed_acls` simply doesn't pick up the 17_plus list, which is the one that adds `MAINTAIN` (`'m'`) for views and materialized views. The result: the MAINTAIN checkbox still appears in the Grant Wizard / privileges UI (added via `table.ui.js`, shared across table-like objects), the user can tick it, but because `allowed_acls` never validated/loaded it for views/mviews, `_parse_privileges` drops it and the generated GRANT statement never includes MAINTAIN.

**To Reproduce**

1. Connect to a PostgreSQL 17+ (or EPAS 17+) server.
2. Open a view or materialized view's Properties/Security tab.
3. Grant a role the MAINTAIN privilege and save.
4. Check the generated SQL / re-open the properties: MAINTAIN was not actually granted.

**Expected behavior**

Either move the 17_plus `allowed_privs.json` files under a `sql/` subdirectory to match the `default` bucket's layout, or make `_ALLOWED_PRIVS_JSON` version-bucket-aware, so the 17+ allowed-privileges list (including MAINTAIN) is actually loaded.

Found while re-verifying #5597 (MAINTAIN privilege support) — the table-side implementation is correct and complete; this is a views/mviews-specific packaging bug in the same feature.

dpage:
## Summary
- `_ALLOWED_PRIVS_JSON = 'sql/allowed_privs.json'` in `views/__init__.py` expects the allowed-privileges list under a `sql/` subdirectory of each version bucket, matching the `default` bucket's layout. The PG17+ `allowed_privs.json` files for views and materialized views (both `pg` and `ppas`) were added directly under the `17_plus` bucket directory instead, so `render_template('views/pg/17_plus/sql/allowed_privs.json')` never found them and silently fell through to the `default` bucket's list via the versioned template loader's fallback - which predates the MAINTAIN privilege and lacks `'m'`.
- Net effect: the MAINTAIN checkbox appears in the Grant Wizard for views/mviews on PG17+/EPAS17+, but ticking it and saving never actually grants MAINTAIN, because `_parse_privileges` drops any privilege not in `allowed_acls`.
- Fix: moved the four affected `allowed_privs.json` files (views/mviews × pg/ppas) under `sql/` to match the `default` bucket's layout, matching one of the two fixes the issue suggested and requiring no code changes.

## Test plan
- [x] Added `web/pgadmin/browser/server_groups/servers/databases/schemas/views/tests/test_allowed_privs_json_path.py`, rendering the PG17 template path directly for all four combinations and asserting `'m'` is present.
- [x] Confirmed all four scenarios fail against the pre-fix file layout (falling back to `default`, missing `'m'`) and pass with the fix.
- [x] Ran the full `views.tests` package - no regressions.
- [x] `pycodestyle` clean on the new test file.

Closes #10350

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL 17+ support for managing privileges on views and materialized views.
  * Added support for the `MAINTAIN` privilege across PostgreSQL and PPAS view objects.
  * Added support for standard materialized-view privileges, including grants and revocations.
  * Added PostgreSQL 17+ view and materialized-view management scenarios, including comments, ownership, definitions, dependencies, and access methods.

* **Tests**
  * Added regression coverage for PostgreSQL and PPAS privilege definitions, including quoted object names and special characters.


kundansable:
### Review

**The core fix is correct and working as expected.** `ViewNode.BASE_TEMPLATE_PATH = 'views/{0}/#{1}#'` plus `_ALLOWED_PRIVS_JSON = 'sql/allowed_privs.json'` resolves `views/pg/#170000#/sql/allowed_privs.json`, and `VersionedTemplateLoader` falls back per-file, so a `17_plus/allowed_privs.json` without the `sql/` level was genuinely unreachable. Moving the four files is the right minimal fix, nothing references the old paths, and the `schemas`/`catalog` templates that also keep `allowed_privs.json` at the bucket root are consistent with *their* own `_ALLOWED_PRIVS_JSON` value, so no wider cleanup is needed.

Two problems with the test-fixture half, though.

#### 1. The new `17_plus` test folders will silently disable the entire views/mviews RE-SQL suite on PG17+

`get_test_folder()` (`web/regression/re_sql/tests/test_resql.py:380-410`) picks the **first existing** version directory and then `os.listdir`s it for `.json` scenario files — folder-level resolution, unlike `get_test_file()`, which falls back per file. That is why every other bucket (`pg/15_plus`, `pg/16_plus`, …) duplicates `tests.json` and `test_mview.json`.

The new `views/tests/pg/17_plus` and `views/tests/ppas/17_plus` contain only the two `.sql` files, so on a PG17+ server the runner resolves to those directories, finds zero JSON files, and runs **no** view/mview RE-SQL scenarios at all — including the ones this PR is trying to pin down. Both new directories need copies of `tests.json` and `test_mview.json`.

#### 2. The RE-SQL (non-msql) fixtures for the same two scenarios are now stale as well

The `sql()` path re-filters through `parse_priv_to_db(..., allowed_acls)` (`views/__init__.py:1487-1492`), so the same ALL-collapse rule in `parse_priv_to_db` applies to reverse-engineered SQL too. Both affected scenarios grant the 7-privilege set to `PUBLIC`, and their expected RE-SQL still reads `GRANT ALL ... TO PUBLIC`:

- `views/tests/pg/default/alter_mview.sql:17` and `views/tests/ppas/default/alter_mview.sql:17`
- `views/tests/pg/default/alter_mview_change_grantee_priv.sql:17` and `views/tests/ppas/default/alter_mview_change_grantee_priv.sql:17`

These need `17_plus` overrides with the explicit privilege list, exactly like the msql fixtures added here.

The `TO <OWNER>` lines and the `alter_mview_drop_all_priv*` fixtures are fine — the owner's PG17 ACL includes `m`, so 8-of-8 still collapses to `ALL`.

#### Minor

`test_allowed_privs_json_path.py` overrides `setUp` to `pass`, which skips the server connection but still relies on the regression harness for `self.app`. That works given how `BaseTestGenerator` is wired, it is just unusual for this directory.

dpage:
@kundansable thanks, both of those were real, and both are fixed in d827f4c08.

Your first point turned out to be even more clear cut than described, since I could reproduce it locally against PostgreSQL 18. With the branch as it was, `regression/runtests.py --pkg resql --modules views` completed in a tenth of a second having run **no** scenarios at all, exactly as you predicted from `get_test_folder()` resolving at folder level: it picked `views/tests/pg/17_plus`, found no JSON in it, and quietly ran nothing. Both 17_plus directories now carry copies of `tests.json` and `test_mview.json`, as every other bucket does, and 43 scenarios run again.

Your second point then showed up as two failures the moment the scenarios were running, in `alter_mview.sql` and `alter_mview_change_grantee_priv.sql`, both still expecting `GRANT ALL ... TO PUBLIC`. Both now have 17_plus overrides for pg and ppas, with the explicit privilege list. Worth noting that the reverse-engineered SQL renders the privileges in ACL character order (`GRANT TRUNCATE, INSERT, DELETE, SELECT, TRIGGER, UPDATE, REFERENCES`), which is not the order the msql fixtures carry, since those follow the order the client sent them in; I checked it is stable across runs rather than something set iteration order was deciding for us. The `TO <OWNER>` lines still collapse to ALL, as you said they would.

On the minor: `setUp` is overridden because the base class's version connects to the test server, which that test has no use for, given that it only renders a template. I have added a comment saying so rather than changing the structure.

Verified with PostgreSQL 18: 43 RE-SQL scenarios pass, plus the six scenarios of `test_allowed_privs_json_path.py`.

dpage:
@coderabbitai On the `ClassVar` nitpick for `scenarios` in `test_allowed_privs_json_path.py`: I'm leaving it as it is. `scenarios` is the attribute `testscenarios` reads from every `BaseTestGenerator` subclass, and all 247 test classes in `web/pgadmin` declare it as a plain `scenarios = [...]`, with none using `ClassVar`; the project doesn't run Ruff either, so annotating just this one would make it the odd one out without satisfying any check we actually use.

dpage:
@coderabbitai On the nitpick asking for a scenario that grants MAINTAIN through the full GRANT path: agreed, and added in 5fd9721ab for views and materialised views, pg and ppas. Each grants MAINTAIN to PUBLIC and checks both the generated and the reverse engineered SQL. The materialised view has a NULL ACL at that point, so the grant also materialises the owner's default ACL, which is why its expected SQL carries the owner's `GRANT ALL` line too. All 47 views RE-SQL checks pass against PostgreSQL 18.

Issue #10363, opened 2026-08-30
Hardening supply chain security against recent npm attacks


dev-hari-prasad:
Needless to mention there's been a wave of npm supply chain attacks recently (Shai-Hulud, the axios hijack, TanStack compromise, etc.), often through install scripts. Given how many packages we depend on, it would be good to harden the setup of pgAdmin and vist the idea once. 

**A few suggestions:**

<details>
<summary><strong>TL;DR</strong></summary>

A few simple hardening changes for pgAdmin:

- **Disable install scripts** → prevents malicious `preinstall`/`postinstall` payloads.
- **Immutable lockfiles + checksum checks** → catches dependency tampering in CI.
- **Pin Font Awesome** → avoids silently pulling unexpected `latest` versions.
- **3-day package age gate** → gives compromised packages more time to be detected.

</details>

## 1. Block install scripts from running automatically

By default, any package can run arbitrary code during install via `preinstall`/`postinstall` scripts, almost always without the knowledge of maintainers or ddevloper. This is actually how most of these recent attacks execute their payload. **Setting `enableScripts: false` in `.yarnrc.yml` blocks this by default, with an explicit allowlist for packages that genuinely need it** (like `sharp`, which downloads a native binary on install).

## 2. Enforce immutable lockfile checks in CI

`yarn.lock` can be tampered with or silently modified during CI builds to pull in mismatched hashes or unreviewed dependencies. Not to say the maintainers here aren't careful, but this kind of change is easy to miss in a normal PR review, since nobody reads a thousand-line lockfile diff line by line. **Switching to `yarn install --immutable` in our GitHub Actions workflows and setting `checksumBehavior: throw` in `.yarnrc.yml` would catch any lockfile drift or hash mismatch automatically before a build even runs.** This benefits the project by guaranteeing 100% reproducible builds and failing early if a lockfile is out of sync or compromised, ensuring unverified code is never bundled into public releases.

## 3. Pin `@fortawesome/fontawesome-free` to an explicit version

[...]

PR #10372, opened 2026-09-02
chore(deps): pin and bump @fortawesome/fontawesome-free to 7.3.1 in /web

dev-hari-prasad:
### Description
Replaces `@fortawesome/fontawesome-free: latest` with exact `7.3.1` in `web/package.json` and updates `yarn.lock`.

### Motivation
- Replaces floating `latest` spec with exact version `7.3.1` to prevent silent minor/patch/major shifts during lockfile refreshes.
- Aligns with upstream maintainer feedback on #10372.

This PR is related to #10363.

## Summary by CodeRabbit

* **Chores**
  * Improved release consistency by locking the icon library to a specific version.
  * No user-facing features or behavior were changed.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -79,7 +79,7 @@
     "@date-io/date-fns": "^3.2.1",
     "@emotion/sheet": "^1.4.0",
     "@fluentui/keyboard-key": "^0.4.23",
-    "@fortawesome/fontawesome-free": "latest",
+    "@fortawesome/fontawesome-free": "7.3.1",
     "@mui/icons-material": "^7.3.11",
     "@mui/material": "^7.3.11",
     "@mui/x-date-pickers": "^8.29.3",
--- a/web/yarn.lock
+++ b/web/yarn.lock
@@ -3195,10 +3195,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@fortawesome/fontawesome-free@npm:latest":
-  version: 7.2.0
-  resolution: "@fortawesome/fontawesome-free@npm:7.2.0"
-  checksum: 10c0/fc95989739857a2622c993929c6748c8c7eac3b490b0de6beb843b0b6a5de99f1e99c92901706c73e4da6be9413853a7382dfe82d2313c217489210a5b9965ed
+"@fortawesome/fontawesome-free@npm:7.3.1":
+  version: 7.3.1
+  resolution: "@fortawesome/fontawesome-free@npm:7.3.1"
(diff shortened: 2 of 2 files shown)


asheshv:
Right problem to go after — this is currently the **only** floating spec in `web/package.json`, so fixing it properly closes the whole category. Two issues with the fix as written, though.

**1. `^7.2.0` isn't a pin (`web/package.json:82`).**
The caret still re-resolves to the newest 7.x every time the lockfile is refreshed. So the PR title says "pin" and the description says it "prevents unintended major version bumps", but what actually remains is the same reproducibility hazard one notch down, at minor/patch level. That's not hypothetical for this repo — we already got burned when `@simonwep/pickr` `^1.9.1` floated to a broken 1.10.0 that only failed in the production bundle. Please use `"7.2.0"` with no range operator, or drop the word "pin".

**2. This quietly performs a version bump (`web/yarn.lock`).**
On master, `@fortawesome/fontawesome-free@npm:latest` resolves to **7.2.0**. With this PR, `^7.2.0` resolves to **7.3.1** (npm `latest` today). So a change framed as *removing* a floating spec actually does the bump it claims to prevent, and hides it in lockfile churn. Either pin `7.2.0` exactly — lockfile bytes stay identical to what's shipping, zero risk, trivially reviewable — or retitle to "pin and bump to 7.3.1" and say so in the body.

No regression risk from the version itself: the only consumer is `web/pgadmin/static/css/style.css:1` (a plain `@import` of `all.css`), there's no `@fortawesome/react-fontawesome` and no per-icon JS imports, so 7.2.0 → 7.3.1 is just a CSS/webfont refresh within the same major. The 5 `run-feature-tests-pg` jobs build the production bundle and passed, which does exercise that import.

Separate note, out of scope: if supply-chain hardening is the motivation here, `web/.yarnrc.yml`'s `checksumBehavior: update` undermines it much more than one floating spec does — it silently rewrites integrity hashes instead of failing on a mismatch. Worth its own issue.

dev-hari-prasad:
Agree, this went through. I will lock it to 7.3.1 instead.

dev-hari-prasad:
@asheshv I've updated the branch to lock `@fortawesome/fontawesome-free` to exact version **`7.3.1`** (with no range operator), rebased onto the latest `master`, and updated the PR title and description accordingly. 
 
Regarding the `checksumBehavior: update` note in `.yarnrc.yml`, agreed; that was discussed previously and is being tracked separately out of scope for this PR in #10363.  Sorry for the earlier mess and commit churn; it was caused by a bit of misunderstanding on my end, but everything is now clean, aligned with your feedback, and passing all checks!

dpage:
Merged to master as d84d3ce9e, thanks. Closing manually since the squash changed the SHA and GitHub won't spot it.

PR #10389, opened 2026-09-06
ci: make yarn install --immutable explicit in CI workflows

dev-hari-prasad:
Part of #10363.

Yarn's `enableImmutableInstalls` setting already defaults to `true` when `CI` is present in the environment (such as in GitHub Actions). Passing `--immutable` explicitly in CI workflows keeps this enforcement self-documenting and transparent.

Additionally, this adds `--immutable` to build and packaging scripts outside CI (`Makefile`, `Dockerfile`, and packaging build scripts for pip, Linux, and macOS) so release and local build processes fail loudly if `yarn.lock` would be modified instead of silently rewriting it.

#### Changes
- Added `--immutable` to `yarn install` in GitHub Actions workflows:
  - `.github/workflows/check-javascript-style.yml`
  - `.github/workflows/run-javascript-tests.yml`
  - `.github/workflows/run-feature-tests-epas.yml`
  - `.github/workflows/run-feature-tests-pg.yml`
- Added `--immutable` to `yarn install` in build and packaging scripts:
  - `Makefile`
  - `Dockerfile`
  - `pkg/pip/build.sh`
  - `pkg/linux/build-functions.sh`
  - `pkg/mac/build-functions.sh`

## Summary by CodeRabbit

* **Chores**
  * Updated dependency installation across automated checks, application builds, and packaging processes to consistently honor the committed lockfile.
  * Builds, style checks, and JavaScript tests now fail when dependency definitions and the lockfile are out of sync, helping prevent unplanned dependency changes.
  * Improved consistency and reproducibility across supported build environments.

### DIFF
--- a/.github/workflows/check-javascript-style.yml
+++ b/.github/workflows/check-javascript-style.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install Node modules
         run: |
           cd web
-          yarn install
+          yarn install --immutable
 
       - name: Run the linter
         run: |
--- a/.github/workflows/run-feature-tests-epas.yml
+++ b/.github/workflows/run-feature-tests-epas.yml
@@ -150,7 +150,7 @@ jobs:
       - name: Build the JS bundle
         run: |
           cd web
-          yarn install
+          yarn install --immutable
           yarn run bundle
 
       - name: Run the tests
--- a/.github/workflows/run-feature-tests-pg.yml
+++ b/.github/workflows/run-feature-tests-pg.yml
@@ -160,7 +160,7 @@ jobs:
       - name: Build the JS bundle
         run: |
           cd web
-          yarn install
+          yarn install --immutable
           yarn run bundle
 
       - name: Run the tests
--- a/.github/workflows/run-javascript-tests.yml
+++ b/.github/workflows/run-javascript-tests.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Install Node modules
         run: |
           cd web
-          yarn install
+          yarn install --immutable
 
       - name: Run the tests
         run: |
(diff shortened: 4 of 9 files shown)


dpage:
I would like to take this, but widened a little first.

On the four workflows the flag is documentation only, since `enableImmutableInstalls` is already true whenever `CI` is set, and the CI logs show Yarn additionally enabling hardened mode on public pull requests. That is fine and worth keeping for exactly the reason you gave, but the places where `--immutable` actually changes behaviour are the five installs that can run outside CI:

- `Makefile:28`
- `Dockerfile:45`
- `pkg/pip/build.sh:65`
- `pkg/linux/build-functions.sh:307`
- `pkg/mac/build-functions.sh:311`

All five run with `yarn.lock` beside them, including the Docker build, which copies `web/` wholesale, and the `yarn set version` call that precedes three of them only rewrites `.yarnrc.yml` and the `packageManager` field, so it cannot trip the flag. A release build that quietly rewrites the lockfile is shipping something other than what we reviewed, so failing loudly there is the point of the exercise.

Add those five to what you already have and I will merge it. The other four PRs in the series are merged, and I have closed #10392 with an explanation.

dev-hari-prasad:
The requested changes are in place exactly as you asked for them. We can merge this and close #10363.

dpage:
Merged as 351376af7, thank you.

I folded in one extra line whilst merging: `Make.bat` had the same unguarded `yarn install` in the Windows packaging path, in the same `yarn set version` + install + bundle shape as the three shell scripts, so leaving it out would have given us a build matrix where five of the six release builds refuse lockfile drift and the sixth silently accepts it. That was an omission in the list I gave you rather than anything missing on your side.

Squashed to a single commit with you as author, and I widened the subject a little since the packaging changes ended up being the substantial part of it. Closing this manually, as the squash changes the SHA and GitHub will not spot it. #10363 is done with this.

PR #10391, opened 2026-09-07
chore(security): disable install scripts by default in /web

dev-hari-prasad:
Part of #10363

Following up on Dave's feedback, this drops the global `enableScripts: true` from `web/.yarnrc.yml` so we're back to Yarn's safe default of blocking install scripts. 

To keep things building properly, `ttf2woff2` and `unrs-resolver` are explicitly allowlisted via `dependenciesMeta` in `web/package.json`. The only other script in the tree (`core-js`) is just a funding banner, so that stays blocked.

I also narrowed down `approvedGitRepositories` to just `pgadmin-org/react-data-grid` while touching the config.

Tested locally with `yarn install --immutable`, the linter, and Jest tests all ran and passed without issues.

> Pretty sure that covers all the packages that actually need build scripts, but if I missed any or if something else should be excluded, please point it out, happy to fix it.

## Summary by CodeRabbit

* **Chores**
  * Updated build and package configuration to allow required dependencies to compile correctly.
  * Restricted approved Git repository access during package installation.
  * Disabled package installation scripts.

### DIFF
--- a/web/.yarnrc.yml
+++ b/web/.yarnrc.yml
@@ -1,9 +1,7 @@
 approvedGitRepositories:
-  - "**"
+  - "https://github.com/pgadmin-org/react-data-grid.git"
 
 checksumBehavior: update
-
-enableScripts: true
 
 logFilters:
   - code: YN0013
--- a/web/package.json
+++ b/web/package.json
@@ -158,6 +158,14 @@
     "wkx": "^0.5.0",
     "zustand": "^5.0.15"
   },
+  "dependenciesMeta": {
+    "ttf2woff2": {
+      "built": true
+    },
+    "unrs-resolver": {
+      "built": true
+    }
+  },
   "resolutions": {
     "rc-resize-observer": "1.4.0",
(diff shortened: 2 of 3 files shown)


dpage:
Merged to master as 809e812a1, thanks; this was the strongest of the set. I checked the allowlist independently before merging: scanning the installed tree for install scripts gives exactly `ttf2woff2`, `unrs-resolver` and `core-js`, so nothing is missing, and I confirmed with a cold cache that the narrowed `approvedGitRepositories` really does match `react-data-grid`, since Yarn strips the `#commit=` treeish before matching. Worth knowing that a local install with a warm cache would not have exercised that at all.

One observation for the record: a fresh install also reports `@parcel/watcher` as having its build script blocked. That is harmless, as it is an optional dependency of `jest-haste-map` and the prebuilt platform binary is installed instead, which is why the Jest runs stayed green on all three platforms.

Closing manually as the squash changed the SHA.

PR #10392, opened 2026-09-07
ci: add --check-cache to yarn install in CI workflows

dev-hari-prasad:
Part of #10363 (Item 6 from the tracking table).

As Dave noted in the issue discussion, `yarn install --check-cache` is a stronger form of cache and lockfile verification in CI.

Instead of just checking that the lockfile is self-consistent and trusting whatever is already in the runner's disk cache, `--check-cache` tells Yarn to re-verify the cached archives against the registry checksums to catch any cache tampering or poisoned archives.

#### Changes
- Added `--check-cache` to `yarn install` across the 4 GitHub Actions workflows:
  - `check-javascript-style.yml`
  - `run-javascript-tests.yml`
  - `run-feature-tests-epas.yml`
  - `run-feature-tests-pg.yml`

## Summary by CodeRabbit

- **Chores**
  - Updated automated JavaScript checks and feature-test workflows to validate the Yarn cache during dependency installation.
  - Workflows now fail when cached dependencies are incomplete, stale, or corrupted.

### DIFF
--- a/.github/workflows/check-javascript-style.yml
+++ b/.github/workflows/check-javascript-style.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install Node modules
         run: |
           cd web
-          yarn install
+          yarn install --check-cache
 
       - name: Run the linter
         run: |
--- a/.github/workflows/run-feature-tests-epas.yml
+++ b/.github/workflows/run-feature-tests-epas.yml
@@ -150,7 +150,7 @@ jobs:
       - name: Build the JS bundle
         run: |
           cd web
-          yarn install
+          yarn install --check-cache
           yarn run bundle
 
       - name: Run the tests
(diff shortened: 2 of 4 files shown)


dpage:
I am going to close this one, and I want to be clear that it is about timing rather than reasoning: your description of what `--check-cache` does is correct, it just rests on an assumption about our CI that does not hold.

`--check-cache` re-verifies archives that came from a cache restored from an earlier run, but none of these four workflows restore one. There is no `cache: yarn` on `setup-node` and no `actions/cache` step, so every job starts with an empty cache, populates it from the registry, and would then be asked to re-verify what it downloaded seconds earlier in the same job. That protects nothing today.

The protection you were actually after arrived with #10395: with `checksumBehavior` back at `throw`, every archive fetched on a cold cache is checksummed against the lockfile, and because CI is always cold that covers the whole tree on Linux, macOS and Windows on every run. If we ever add Yarn caching to these workflows then this flag becomes worth having and should land in the same change, so it is worth remembering rather than discarding.

Thanks for the thought, and see #10389 for where I would like the effort to go instead.

PR #10394, opened 2026-09-07
fix: render Validate binary path modal as HTML instead of raw markup

### LINKED ISSUE #10393: Validate binary path dialog displays raw HTML tags instead of rendered content
**Describe the bug**

Preferences → Paths → Binary Paths → **Validate binary path** shows the raw HTML markup literally instead of rendering it:

```
<b>pg_dumpall:</b> not found on the specified binary path.<br/><b>pg_dump:</b> not found on the specified binary path.<br/><b>pg_restore:</b> not found on the specified binary path.<br/><b>psql:</b> not found on the specified binary path.<br/>
```

**To Reproduce**

1. Open Preferences → Paths → Binary Paths.
2. Enter any path (valid or invalid) in a server's Binary Path field.
3. Click "Validate binary path".

**Expected behavior**

Rendered HTML — bold utility names, one entry per line via `<br/>` — as it did before the regression.

**Cause in the code**

`web/pgadmin/misc/__init__.py` (`validate_binary_path`, around line 322) builds an HTML string by hand:

```python
version_str += "<b>" + utility + ":</b> " + "not found on the specified binary path.<br/>"
```

`web/pgadmin/preferences/static/js/components/binary_path.ui.js` used to render this with `notifier.alert()`, which passes the string through `HTMLReactParser(DOMPurify.sanitize(text))`. The XSS hardening commit `9e370d3cb` ("fix(xss): comprehensive XSS hardening across notification and Explain flows") switched this call site to `notifier.alertText()`, which routes through `SafeMessage` (`web/pgadmin/static/js/components/SafeMessage.jsx`) and renders the string as literal, escaped text — correct behavior for untrusted content, but the backend still emits HTML instead of plain data, so the tags now show up literally.

**Suggested fix**

Stop transporting HTML from the backend. Have `validate_binary_path` return structured data instead (e.g. `[{utility, version}]`, with `version: null` meaning "not found"), and have the frontend compose the modal content as real React elements (bold label + `SafeMessage` for the untrusted version text) passed to `notifier.alert()`. This keeps binary `--version` stdout — untrusted OS-process output — out of any HTML parser entirely, while restoring the intended bold/line-broken formatting.

 - pgAdmin version: 9.x (regression introduced by 9e370d3cb, XSS hardening pass)
 - Mode: both Desktop and Server mode (affects the shared modal helper)

**Additional context**

This is a straightforward regression from a security fix that only patched one side (frontend rendering) without updating the other (backend response format). A fix is already prepared and verified (pep8, eslint, Jest) on a local branch.

kundansable:
## Summary
Fixes #10393.

- `Preferences → Paths → Binary Paths → Validate binary path` was showing literal `<b>...</b>`/`<br/>` tags instead of formatted text. The endpoint built a pre-rendered HTML string, and the frontend call site was switched to the escaping `notifier.alertText()` during the XSS hardening pass (`9e370d3cb`) — correct for untrusted binary output, but the backend never stopped speaking HTML.
- `misc.validate_binary_path` now returns structured data (`[{utility, version}]`) instead of markup.
- The frontend composes the modal content as real React elements (`<b>` label + `SafeMessage` for the untrusted version text) and calls `notifier.alert()`, so nothing is HTML-parsed and no plaintext escaping is needed.
- Renamed `binary_path.ui.js` → `.jsx` since it now contains JSX; widened `AlertContent`'s `text` propType to `node` to match its existing behavior (it already accepted non-string nodes from other callers).

## Test plan
- [x] `pycodestyle` on the changed Python file — clean
- [x] `yarn run eslint` on the changed JS/JSX files — clean
- [x] `yarn run jest --testPathPatterns="binary_path.ui"` — 3 passed, including a new case asserting bold labels/line breaks render and no literal `<b>`/`<br/>` ever appears in the DOM
- [x] `yarn run jest --testPathPatterns="ModalProvider"` — 4 passed (propType widening doesn't regress existing behavior)
- [ ] Manual: Preferences → Paths → Binary Paths → enter an invalid path → Validate binary path → confirm formatted, line-broken output (no raw tags)

## Summary by CodeRabbit

- **Bug Fixes**
  - Binary path validation results now display clear, per-utility version information.
  - Utilities without detected versions now show an appropriate fallback message.
  - Validation alerts render formatted labels and line breaks correctly instead of displaying raw markup.

- **UI Improvements**
  - Binary path validation feedback is presented in a clearer alert dialog.
  - Alert content now supports richer formatted messages for improved readability.

### DIFF
--- a/web/pgadmin/misc/__init__.py
+++ b/web/pgadmin/misc/__init__.py
@@ -335,23 +335,19 @@ def validate_binary_path():
     if data != '':
         data = json.loads(data)
 
-    version_str = ''
-
     # Do not allow storage dir as utility path
     if 'utility_path' in data and data['utility_path'] is not None and \
         Path(config.STORAGE_DIR) != Path(data['utility_path']) and \
             Path(config.STORAGE_DIR) not in Path(data['utility_path']).parents:
         binary_versions = get_binary_path_versions(data['utility_path'])
-        for utility, version in binary_versions.items():
-            if version is None:
-                version_str += "<b>" + utility + ":</b> " + \
-                               "not found on the specified binary path.<br/>"
-            else:
-                version_str += "<b>" + utility + ":</b> " + version + "<br/>"
+        utilities = [
+            {'utility': utility, 'version': version}
+            for utility, version in binary_versions.items()
+        ]
     else:
         return precondition_required(gettext('Invalid binary path.'))
 
-    return make_json_response(data=gettext(version_str), status=200)
+    return make_json_response(data=utilities, status=200)
 
 
 @blueprint.route("/upgrade_check", endpoint="upgrade_check",
--- a/web/pgadmin/preferences/static/js/components/binary_path.ui.jsx
+++ b/web/pgadmin/preferences/static/js/components/binary_path.ui.jsx
@@ -13,6 +13,7 @@ import url_for from 'sources/url_for';
 import BaseUISchema from 'sources/SchemaView/base_schema.ui';
 import getApiInstance from '../../../../static/js/api_instance';
 import pgAdmin from 'sources/pgadmin';
+import { SafeMessage } from '../../../../static/js/components/SafeMessage';
 
 export function getBinaryPathSchema() {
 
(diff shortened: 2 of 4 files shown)


hiteshjambhale:
Backend (misc/__init__.py) — Clean change. Returning structured [{utility, version}] instead of an HTML string is the right API design. Removing gettext() around the whole blob was also a latent bug fix.
Frontend (binary_path.ui.jsx) — SafeMessage on the version string is correct since it comes from an external binary. One nit: use || instead of ?? — if a binary ever returns empty string stdout, ?? won't fall back to "not found" but || will.
ModalProvider.jsx — PropTypes.string → PropTypes.node widening is correct and matches how AlertContent already works internally. Non-string nodes bypass DOMPurify, so callers own sanitization — which this PR handles correctly.
Tests — Solid. The key assertion (no literal <b>/<br/> in DOM) is exactly the right regression guard. The await Promise.resolve() double-flush is fragile; waitFor would be safer, but it works for now.

@kundansable

hiteshjambhale (approved):
Tested manually — validated with /opt/homebrew/bin, versions render correctly with bold labels and line breaks. No raw <b>/<br/> tags.
LGTM ✅

PR #10396, opened 2026-09-07
fix(pkg): stop the Windows build overwriting the venv's pip

asheshv:
The Windows x64 build has been failing since the buildfarm's virtualenv started seeding pip 26.2.x:

```
File "...\win-temp\venv\Lib\site-packages\pip\_internal\build_env\installer.py", line 22, in <module>
  from pip._internal.utils.misc import get_runnable_pip
ImportError: cannot import name 'get_runnable_pip' from 'pip._internal.utils.misc'
```

`CREATE_VIRTUAL_ENV` creates a virtualenv and then copies the whole of the system Python's `Lib` over it, so the relocated interpreter has a standard library. That copy includes `Lib\site-packages`, and `XCOPY /Y` overwrites the files present on both sides without removing the ones only the venv has, so the venv's freshly seeded pip ends up spliced together with whatever pip `C:\Python313` happens to have.

That was harmless while the two layouts agreed. pip 26.2 turned `pip/_internal/build_env` from a module into a package, so on a builder whose system pip predates 26.2 the new `build_env/` directory survives the copy whilst `utils/misc.py` is overwritten by the older one — and the resulting install is internally inconsistent. I checked the pristine 26.2.1 wheel: `build_env/` is a package and `utils/misc.py` does define `get_runnable_pip`, so the released artifact is fine; only the spliced copy is broken.

It cannot recover on its own, because the next line of the script is the pip invocation that upgrades pip.

Copy the standard library with `ROBOCOPY` and exclude `site-packages`, which the venv is entitled to keep to itself. `ROBOCOPY` rather than `XCOPY /EXCLUDE` because `/XD` takes a directory to skip, whereas `/EXCLUDE` takes a file of path substrings and breaks on quoted paths containing spaces. `/R:3 /W:5` so a locked file fails the build instead of retrying a million times.

Note that `CALL :CHECK_ROBOCOPY_ERROR` is followed by `|| EXIT /B 1` here: `EXIT /B` inside the label returns from the label rather than from the caller, so without it a failed copy would be ignored.

Upgrading the system pip on the builder (`C:\Python313\python.exe -m pip install --upgrade pip`) also unblocks it immediately, but the splice will bite again the next time the two pip layouts diverge.

Not verified on a Windows builder — I have no Windows host to run `Make.bat` on. A buildfarm run is the real test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reliability by stopping setup when virtual environment creation or required file-copy operations fail.
  * Updated environment setup to avoid overwriting bundled package tooling.
  * Simplified file-copy progress output for a cleaner build experience.

### DIFF
--- a/Make.bat
+++ b/Make.bat
@@ -161,13 +161,13 @@ REM Main build sequence Ends
     CD "%TMPDIR%"
 
     REM Note that we must use virtualenv.exe here, as the venv module doesn't allow python.exe to relocate.
-    "%PGADMIN_PYTHON_DIR%\Scripts\virtualenv.exe" venv
+    "%PGADMIN_PYTHON_DIR%\Scripts\virtualenv.exe" venv || EXIT /B 1
 
     XCOPY /S /I /E /H /Y "%PGADMIN_PYTHON_DIR%\DLLs" "%TMPDIR%\venv\DLLs" > nul || EXIT /B 1
     REM Copy the standard library, but NOT site-packages: the venv already has its
     REM own seeded pip there, and overwriting only the files the system Python also
     REM has leaves a mix of two pip versions behind.
-    ROBOCOPY /E /R:3 /W:5 "%PGADMIN_PYTHON_DIR%\Lib" "%TMPDIR%\venv\Lib" /XD "%PGADMIN_PYTHON_DIR%\Lib\site-packages" > nul
+    ROBOCOPY /E /R:3 /W:5 /NFL /NDL /NP "%PGADMIN_PYTHON_DIR%\Lib" "%TMPDIR%\venv\Lib" /XD site-packages
     CALL :CHECK_ROBOCOPY_ERROR || EXIT /B 1
 
     ECHO Activating virtual environment -  %TMPDIR%\venv...
@@ -220,8 +220,8 @@ REM Main build sequence Ends
     RD /Q /S "%WD%\web\pgadmin\static\js\generated\.cache" 1> nul 2>&1
 
     ECHO Copying web directory...
-    ROBOCOPY /S "%WD%\web" "%BUILDROOT%\web" > nul
-    CALL :CHECK_ROBOCOPY_ERROR
+    ROBOCOPY /S /NFL /NDL /NP "%WD%\web" "%BUILDROOT%\web"
+    CALL :CHECK_ROBOCOPY_ERROR || EXIT /B 1
 
     ECHO Installing javascript dependencies...
     CD "%BUILDROOT%\web"
--- a/Make.bat
+++ b/Make.bat
@@ -164,7 +164,11 @@ REM Main build sequence Ends
     "%PGADMIN_PYTHON_DIR%\Scripts\virtualenv.exe" venv
 
     XCOPY /S /I /E /H /Y "%PGADMIN_PYTHON_DIR%\DLLs" "%TMPDIR%\venv\DLLs" > nul || EXIT /B 1
-    XCOPY /S /I /E /H /Y "%PGADMIN_PYTHON_DIR%\Lib" "%TMPDIR%\venv\Lib" > nul || EXIT /B 1
+    REM Copy the standard library, but NOT site-packages: the venv already has its
+    REM own seeded pip there, and overwriting only the files the system Python also
+    REM has leaves a mix of two pip versions behind.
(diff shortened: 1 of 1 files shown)


asheshv:
One consequence worth a reviewer's eye before this is merged.

The venv is seeded with pip alone — the build log shows `added seed packages: pip==26.2.1`, no setuptools, no wheel — and `virtualenv.exe venv` runs without `--system-site-packages`. So every third-party package that was importable inside that venv got there *only* through the `XCOPY` this PR narrows. Excluding `site-packages` means the system Python's globally installed packages are no longer copied in.

What the build actually needs is unaffected, as far as I can tell from the script:

- `setuptools` is in `requirements.txt` (`setuptools==84.*` for `python_version > '3.9'`), installed into the venv at line 178
- `pip` is seeded by virtualenv, then upgraded at line 177
- `sphinx` and `sphinxcontrib-youtube` are installed explicitly in `BUILD_DOCS`
- `virtualenv` itself is only needed in the system Python, never inside the venv

There is a second effect, which I think is an improvement but is a behaviour change either way. `CREATE_PYTHON_ENV` copies the venv's `site-packages` into the shipped installer:

```
XCOPY /S /I /E /H /Y "%TMPDIR%\venv\Lib\site-packages" "%BUILDROOT%\python\Lib\site-packages"
```

Under the old behaviour that meant whatever the builder happened to have pip-installed globally shipped inside pgAdmin too — `virtualenv` and its dependencies at minimum — so installer contents varied with the state of the build machine. After this change they follow from `requirements.txt` alone.

The risk is the mirror image of that: if some import is currently satisfied only by a globally installed package that never made it into `requirements.txt`, this will surface it, either as a build failure or as a module missing from the installer. I have no Windows host, so I could not check. Could someone with a builder compare `win-build\python\Lib\site-packages` before and after, and confirm nothing needed has gone missing? A failure on the buildfarm is the outcome I would prefer over continuing to depend on whatever `C:\Python313` happens to contain, but it should be a deliberate choice rather than a surprise.

dpage:
Reviewed, and the diagnosis holds up against the Jenkins log: the traceback is `build_env/installer.py` (the 26.2 package layout) importing `get_runnable_pip` from a `utils/misc.py` that does not define it, which is exactly what a file-by-file overwrite of a newer pip by an older one produces. Excluding `site-packages` is the right fix, and the `CALL :CHECK_ROBOCOPY_ERROR || EXIT /B 1` form is correct, including the point about `EXIT /B` returning from the label.

I have pushed four follow-up fixes to the branch rather than leaving them as comments, since they are all one-liners:

- **Exclude by name, not by full path.** `/XD` now takes `site-packages`. The full path was built from `PGADMIN_PYTHON_DIR`, so if that variable is ever set with a trailing backslash, both the source and the `/XD` argument gain a doubled separator; if the match then failed, `site-packages` would be copied again and the build would fail with the same baffling `ImportError`. Nothing in the standard library is legitimately called `site-packages`, so the name form is unconditional.
- **`> nul` dropped from both `ROBOCOPY` calls** in favour of `/NFL /NDL /NP`. ROBOCOPY writes errors to stdout, so with the redirect a return of 8 or more aborted the build with nothing in the log to say what failed, which `/R:3 /W:5` makes rather more likely to be hit than the old million retries did.
- **The `web` directory copy is now guarded too.** `CALL :CHECK_ROBOCOPY_ERROR` on line 220 had no `|| EXIT /B 1`, so a failed copy was silently ignored and the build carried on packaging a partial tree.
- **`virtualenv.exe` is guarded.** With the standard library copied by ROBOCOPY, which creates its own destination, a virtualenv failure would otherwise have surfaced at the activate step rather than where it happened.

On your question about the shipped `site-packages`: I agree with the analysis, and `requirements.txt` does carry `setuptools==84.*`, so nothing the script itself invokes needs the builder's global packages. Worth being explicit that the residual risk is not build-time but runtime, though: an import satisfied only by a globally installed package would now be absent from the installer and would fail when a user reached that code path, which a green buildfarm run will not catch. A file-list diff of `win-build\python\Lib\site-packages` before and after is what would settle it, and that needs someone with a builder.

Neither your change nor mine has been run on Windows, so the buildfarm remains the only test either of us can give it; nothing under `.github/` touches `Make.bat`. I am happy for this to go in and be proven there, since the alternative is continuing to depend on whatever `C:\Python313` happens to contain.

There is also a pre-existing typo at `Make.bat:349`, where the `pg_dumpall.exe` copy ends `|| EXIT /B 1L%`. I have left it out of this branch to keep the change focused, but it wants fixing at some point.

dpage (approved):

PR #10398, opened 2026-09-07
Fix dynamic type function not receiving correct state for collection

robe2:
…row fields

When a schema field uses a dynamic type function (type as a function) and is rendered inside a collection (e.g., chain tasks), the function always received the top-level schema data instead of the row-level data. This caused the field to fall through to its default type, ignoring the actual field values in the row.

Pass the parent-path state to the dynamic type function so it receives the correct context for collection row fields.

I ran into this issue when building pg_timetable UI - #10152  that when building a new Chain and Tasks in one step, the Kind toggle didn't change options based on Kind.  This patch fixes it, but didn't seem appropriate to put in as part of pg_timetable since it's a global issue.  I don't think any existing elements are impacted by it yet.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dynamic field type evaluation for nested form fields.
  * Ensures field types are determined using the appropriate parent field state when applicable.
  * Improved handling of dependent field values when dynamically determining form behavior.

### DIFF
--- a/web/pgadmin/static/js/SchemaView/MappedControl.jsx
+++ b/web/pgadmin/static/js/SchemaView/MappedControl.jsx
@@ -401,7 +401,12 @@ export const MappedFormControl = ({
   }
 
   if (typeof (field.type) === 'function') {
-    const typeProps = evalFunc(null, field.type, state);
+    let typeState = state;
+    if (accessPath && accessPath.length > 1) {
+      const parentPath = accessPath.slice(0, -1);
+      typeState = schemaState.value(parentPath);
+    }
+    const typeProps = evalFunc(null, field.type, typeState);
     newProps = {
       ...newProps,
       ...typeProps,


dpage:
Thanks Regina, this is a good catch, and it fixes rather more than the pg_timetable case that prompted it. I traced the access paths before merging: nested tabs and fieldsets carry no `id`, so they don't extend the path, which means a path longer than one segment only ever comes from a collection row, and `schemaState.value([])` already returns the whole data. That let me collapse the branch to a single call and drop the now-unused `state` local; I also rewrote the comment above it, which was left saying the opposite of what the code now does, and added a Jest spec in `web/regression/javascript/SchemaView/DynamicFieldType.spec.js` covering both the top-level and the collection-row case (it fails against the old behaviour).

Worth noting for the archaeology: as well as your case, this repairs the expanded-row forms that were already relying on row state and silently getting the table's, namely the foreign key 'Referencing' column list, the partition name attach/create switch, and the column dialog's identity/generated options.

Merged to master as 6312baa34. Since the squash changes the SHA, GitHub won't close this automatically, so I'm closing it by hand.

Issue #10409, opened 2026-09-11
[LOG_AUTHENTICATED_USER] missing name when logout


albundy83:
**_Please note that security bugs or issues should be reported to security@pgadmin.org._**

**Describe the bug**

Hello,

sine the new feature to log authenticated user has been done [here](https://github.com/pgadmin-org/pgadmin4/issues/9990), it  seems that when you logout, username is not displayed in log.

Here an example with logout from oauth2:
```bash
1.2.3.4 - - [11/Sep/2026:06:37:07 +0000] "GET /oauth2/logout?next=/browser/ HTTP/1.1" 302 6269 "https://pgadmin4.youpi.fr/browser/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/152.0.0.0 Safari/537.36"
```

Here an example with "classical" logout:
```bash
1.2.3.4 - - [11/Sep/2026:06:34:21 +0000] "GET /logout?next=/browser/ HTTP/1.1" 302 205 "https://pgadmin4.youpi.fr/browser/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/152.0.0.0 Safari/537.36"
```

**To Reproduce**
Deploy pgadmin4 and enable LOG_AUTHENTICATED_USER, click on Logout button.

**Expected behavior**

Add missing username in log:
```bash
1.2.3.4 - me@youpi.fr - [11/Sep/2026:06:34:21 +0000] "GET /logout?next=/browser/ HTTP/1.1" 302 205 "https://pgadmin4.youpi.fr/browser/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/152.0.0.0 Safari/537.36"
```

**Desktop (please complete the following information):**
 - OS: Kubernetes 1.34.8
 - pgAdmin version: 9.17
 - Mode: Server
 - Browser (if running in server mode): Chrome/152.0.0.0
 - Package type: Container

Thanks a lot

Garvit-png:
@albundy83 , can you assign me this issue

dev-hari-prasad:
> [@albundy83](https://github.com/albundy83) , can you assign me this issue

@albundy83 Seems to be already working on this in PR #10410.

Garvit-png:
[...]

PR #10410, opened 2026-09-11
Try to remember username and inject it on logout

albundy83:
Hello,

here a fix to add missing username when clicking on Logout button (GET on /logout)
Fix issue https://github.com/pgadmin-org/pgadmin4/issues/10409

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved authenticated usernames in remote-user headers during logout.
  * Continued sanitizing usernames for encoding compatibility and removal of non-printable characters.
  * Prevented spoofed remote-user headers from persisting for anonymous requests.
  * Ensured authenticated-user logging can be disabled through configuration.

* **Tests**
  * Added coverage for authentication, logout, anonymous requests, configuration, and username sanitization.

### DIFF
--- a/web/pgadmin/__init__.py
+++ b/web/pgadmin/__init__.py
@@ -21,7 +21,7 @@ from types import MethodType
 from collections import defaultdict
 from importlib import import_module
 
-from flask import Flask, abort, request, current_app, session, url_for
+from flask import Flask, abort, request, current_app, g, session, url_for
 from flask_socketio import SocketIO
 from werkzeug.exceptions import HTTPException
 from flask_babel import Babel, gettext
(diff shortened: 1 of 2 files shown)


dpage:
Thanks Grégoire, merged as 53729372a73c515283c4826c05a2f3856f0cc484.

I folded in two small test changes whilst merging: the stub response now uses a real `werkzeug.datastructures.Headers` rather than a plain dict, so the sanitisation cases genuinely prove Werkzeug will accept the value, and there's an extra case that drives a real `GET /logout` through the test client and asserts the header survives it, which guards the `before_request` wiring itself.

GitHub won't close this automatically since the commit was squashed, so I'll close it manually.

Issue #10407, opened 2026-09-10
Desktop package builds resolve runtime dependencies without a lockfile


dpage:
Whilst reviewing the supply chain hardening PRs from #10363 I noticed that none of that hardening reaches the desktop package builds, which is arguably where it matters most, since whatever npm serves on the day of the build ends up inside a signed installer.

### 1. The runtime install runs without a lockfile

All three bundle builds copy `runtime/package.json` and `runtime/.yarnrc.yml` into the staged application directory and then run `yarn workspaces focus --production` there, but none of them copies `runtime/yarn.lock`:

- `pkg/linux/build-functions.sh:244-255`
- `pkg/mac/build-functions.sh:64-79`
- `Make.bat:293-305`

The lockfile does exist in the tree, and it is what CI installs and tests against, so the effect is that the runtime dependencies inside the shipped packages are resolved fresh from the semver ranges on every build, rather than being the versions we reviewed and tested. It also means that the `checksumBehavior: throw` which #10395 has just restored in `runtime/.yarnrc.yml`, and which these builds dutifully copy into the bundle, has nothing to compare anything against and sits there inert.

Copying the lockfile alongside the other two files should be most of the fix, though it wants testing on each platform, since the resolutions it pins may well differ from whatever the last few builds happened to pick up.

### 2. The Windows build still takes whatever Electron is newest

`Make.bat:314` resolves the Electron version with `npm info electron version`, which returns whatever currently carries the `latest` dist tag on the registry. The Linux and macOS builds were changed to read the pinned version out of `runtime/package.json` instead, and carry a comment at `pkg/linux/build-functions.sh:205-210` spelling out why: the registry lookup lets a newly published Electron release land in shipped binaries without anyone reviewing it. Windows should do the same thing.

@dev-hari-prasad this follows on directly from your work in #10363, so it seemed only fair to offer it to you first if you fancy it.

dev-hari-prasad:
Thanks for analysis Dave. I'll look into this and do the needful.

dev-hari-prasad:
I have opened #10413, which addresses the second issue: the Windows build using the latest Electron version.

While tracking this down, I also found something that should probably be addressed immediately:

[...]

PR #10413, opened 2026-09-12
build(win32): pin electron version from runtime/package.json instead of resolving latest in Make.bat

dev-hari-prasad:
`Make.bat` currently resolves Electron via `npm info electron version`. This pulls whatever version is tagged `latest` on npm at build time, allowing unreviewed upstream releases into Windows builds.

This PR ports the fix **already applied to Linux and macOS build scripts:**
* Reads the pinned Electron version directly from `"%WD%\runtime\package.json"`.
* Strips any leading caret and fails with an explicit error if the version is missing.
* Keeps the Windows build deterministic and avoids live registry queries.

As discussed in #10407, this hardens the security profile of the runtime and applies some of the supply chain hardening benefits from #10363.

### Testing
* Resolves to `43.4.0` (pinned in `runtime/package.json`) instead of `44.3.0` (npm latest).
* Verified the explicit `%WD%` path works regardless of the current working directory.
* Confirmed it errors out and exits with code 1 if the version cannot be found.

## Summary by CodeRabbit

- **Bug Fixes**
  - Windows, Linux, and macOS builds now use the Electron version pinned in the lockfile for consistent downloads.
  - Builds now stop with a clear error when the lockfile is missing or the pinned Electron version cannot be resolved or does not match the expected version format.


dev-hari-prasad:
Please do not merge the PR untill a pending issue in the linked issue is fixed. I shall deleted the comment once it's addressed.

Thanks

dpage:
Thanks for tracking this down, Hari, and for the write-up in Discord: the mismatch is real and worth fixing before the next build tagging.

I have pushed a commit onto your branch rather than leaving it as review comments, since it was easier to show than to describe. Please review it, and push back if you disagree with any of it.

Two changes to what you had:

**Read the resolved version from `runtime/yarn.lock`, not the range in `runtime/package.json`.** The range is only a lower bound, so the version yarn actually resolves can drift ahead of it, and we would then ship a different Electron from the one anyone gets out of `yarn install`. `yarn info electron --json` reports what the lockfile pins, and I confirmed it needs neither the network nor `node_modules`, so it is fine in a build script. It also retires the caret strip, which only handled `^`: given `~43.4.0` the old code passed the string through intact and the build would have sat in the `wget` retry loop forever chasing a tag that does not exist.

**Applied the same change to `pkg/linux/build-functions.sh` and `pkg/mac/build-functions.sh`.** They read the range too, so changing Windows alone would have reintroduced the cross-platform divergence this PR exists to close, just in a subtler form.

One trap worth knowing about: `yarn info` has to run inside `runtime/`, because outside a yarn project it quietly falls back to querying the registry and still exits 0, which is exactly the behaviour we are removing. The commit checks the lockfile is present first and validates that what comes back looks like a version, so both failure modes stop the build rather than downloading something unintended.

I have verified the Linux path here: it resolves `43.4.0` in line with the lockfile, and with the range edited down to `^43.0.0` the old code gives `43.0.0` whilst the new code still gives `43.4.0`, which is the drift case in miniature. **Could you run a Windows build against it?** The `node -e` call in `Make.bat` sits inside a `FOR /F` with an escaped pipe, and cmd's parsing of that construct is precisely where this sort of thing tends to come apart. If it does misbehave, the fallback is to move the parser into a small helper script that all three platforms call, which removes the quoting problem and the triplication along with it; say the word and I will do that instead.

Also worth checking that the `yarn` on the Windows builder's path is v4. Yarn Classic would produce a different JSON shape, which the validation rejects, so it would fail loudly rather than silently, but better to know now.

No release note needed; we are batching those up shortly before release now.

PR #10416, opened 2026-09-13
fix: Prefer email as Gravatar identifer when available

### LINKED ISSUE #10415: Gravatar is based on username even if email is available

**Describe the bug**

When a pgadmin user has an email address set (e.g. from an OAuth2 provider), the Gravatar shown in the user menu is still based on the local username, which may be different from the Gravatar username of the user (or worse, the local username matches a Gravatar username of a different person).

**To Reproduce**

1. Ensure that `SHOW_GRAVATAR_IMAGE = True`
2. Create a pgadmin user with a username for which no Gravatar profile exists, and a valid email address for which a Gravatar profile exists
3. Log in as that user
4. Observe a default Gravatar next to the user’s name

**Expected behavior**

When a user’s email address is known, it should be used as the Gravatar identifier.

**Error message**

**Screenshots**

Image

 - OS: Alpine Linux v3.24, Linux-7.1.8-200.fc44.aarch64-aarch64-with-musl1
 - pgAdmin version: 9.17
 - Mode: Server
 - Browser (if running in server mode): Safari Version 26.6.2 (21624.5.1.11.3)
 - Package type: Container

**Additional context**

marvinruder:
* Fixes #10415

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved profile avatar rendering by using the signed-in user’s email address when available, with the username used as a fallback.
  - Ensured avatar identifiers containing apostrophes and other special characters are handled correctly, preventing malformed avatar links.
  - Improved consistency between profile information and the avatar displayed in the user menu.


dpage:
Merged as 2bb59c0b1, thanks! I folded in a `:param identifier:` line on the filter docstring whilst I was there, since the rename had left it undocumented; no functional change beyond what you had.

The squash rewrites the SHA, so GitHub won't close this automatically and I'm doing it by hand.

PR #10423, opened 2026-09-16
Update Japanese translation for v9.18

rlbxku1r:
Update Japanese translation for the upcoming pgAdmin v9.18 and PostgreSQL 19 release.


dpage:
Merged as 5c7f0bdb0 on `master`, with you preserved as the author; the squash changed the SHA, so GitHub won't close this one automatically.

Thank you, this was a pleasure to review. Full coverage against the current template, no whitespace or placeholder regressions anywhere (including the fiddly concatenated strings such as `Failed to generate report: `), and the compiled `.mo` you included is byte-identical to what the source compiles to, which is more than I can usually say.

One question, purely out of curiosity rather than anything needing action now: the two entries at `pga_schedule.ui.js:134-135`, `Schedules are specified using a ` and `cron-style`, are still marked fuzzy and so are excluded from the `.mo`, meaning they display in English. Both already carry what look like good translations, so I wondered whether the marker is deliberate: they are two fragments of one sentence, and I can well imagine the word order not working when they are stitched together in Japanese. If it is just a leftover marker then dropping it in a follow-up PR would ship them, but if the concatenation is the problem then the right fix is in the source rather than the catalogue, and I am happy to look at that separately.

rlbxku1r:
> One question, purely out of curiosity rather than anything needing action now: the two entries at `pga_schedule.ui.js:134-135`, `Schedules are specified using a ` and `cron-style`, are still marked fuzzy and so are excluded from the `.mo`, meaning they display in English. Both already carry what look like good translations, so I wondered whether the marker is deliberate: they are two fragments of one sentence, and I can well imagine the word order not working when they are stitched together in Japanese. If it is just a leftover marker then dropping it in a follow-up PR would ship them, but if the concatenation is the problem then the right fix is in the source rather than the catalogue, and I am happy to look at that separately.

They are deliberately marked fuzzy (for several years) simply because they cannot be translated:
https://github.com/pgadmin-org/pgadmin4/blob/7c6fb89b82cc26cd6dacef0486d60b7de919b8c8/web/pgadmin/browser/server_groups/servers/pgagent/schedules/static/js/pga_schedule.ui.js#L134-L136
The full sentence is `Schedules are specified using a <strong>cron-style</strong> format.`, but the `format.` part is not marked as translatable.

PR #10427, opened 2026-09-17
Only run each CI workflow when something it tests has changed

dpage:
Every workflow fires on every push and every pull request, so a commit touching nothing but documentation or a workflow file starts the whole matrix: the feature tests against both PostgreSQL and EPAS, the container build, the translations check and the rest. That is a slow wait for a green tick on a one-line docs fix, and it burns hosted runner capacity we sometimes need elsewhere.

Each workflow now declares the paths it actually depends on, plus its own file so that editing a workflow still runs it.

The scoping is per workflow rather than one shared list:

- **pycodestyle** on any `.py` anywhere, plus `.pycodestyle`.
- **eslint** and **Jest** on the JavaScript and the files that configure it, rather than on all of `web/`, so a Python-only change no longer runs them.
- **Documentation** on `docs/`, plus `web/version.py` and `web/branding.py`, which it renders into the output.
- **Packaging checks** on their own `pkg/` subdirectory alongside the application and `requirements.txt`.
- **Python and feature tests** on `web/`, `requirements.txt` and the `Makefile`.

`sonarqube-scan` is deliberately untouched, being a whole-repository quality scan rather than a test of any one thing.

No status checks are required for merging, so a workflow that correctly skips does not block anything.

## Summary by CodeRabbit

* **Chores**
  * Updated automated build, style, test, translation, packaging, and feature-test workflows to run when relevant source, documentation, configuration, licensing, or shared setup files change.
  * Added consistent path-based filtering for pushes and pull requests, reducing unnecessary CI runs while ensuring affected changes receive the appropriate validation.
  * Refined several workflows to limit execution to relevant changes on the primary branch.


dpage:
All eight applied in b098e4e5c. I checked each against the tree rather than taking them on trust, and they all held up; my filters were too tight in exactly the way that matters, since a test that wrongly skips is worse than a runner that is wrongly used.

For the record of what each one actually depends on:

- The container build does `COPY docs /pgadmin4/docs`, and `.dockerignore` determines what is in the build context at all.
- `docs/en_US/conf.py` imports `web/config.py`.
- `make install-python` runs `tools/setup-python-env.sh`, which affects all eight workflows that call it.
- `pkg/pip/build.sh` packages `LICENSE` and `README.md` and builds the docs; `pkg/src/build.sh` does much the same for the tarball.
- `web/.yarnrc.yml` and `web/babel.config.json` change what eslint and Jest do, and the JavaScript workflows filter on JavaScript rather than on all of `web/`.

One deviation. `tools/get_sb_package.py` is added to `run-python-tests-pg.yml` and `run-python-tests-epas.yml`, which fetch it, rather than to `run-python-tests-server-mode.yml`, which does not reference it.

